PR #6, opened 2025-02-04
SQL-2577: Refine Schemas in matches that require Any or AnyOf Schemas…

pmeredit:
… to actually be Documents

Fresh PR to the new repo. I'll be deleting the old one.

### DIFF
--- a/agg-ast/schema_derivation/src/lib.rs
+++ b/agg-ast/schema_derivation/src/lib.rs
@@ -121,6 +121,83 @@ pub(crate) fn get_schema_for_path_mut(
     schema
 }
 
+/// Gets or creates a mutable reference to a specific field or document path in the schema. This
+/// should only be used in a $match context or in some other context where the MQL operator can
+/// actually create fields. Consider a $match: we could think a field has type Any, or
+/// AnyOf([String, Document]) before the $match, and the $match stage can only evaluate to true, if
+/// that field is specifically a Document. In this case, we can refine that schema to Document, and
+/// this can recurse to any depth in the Schema. Note that this still returns an Option because, if
+/// the field is known to have a Schema that cannot be a Document, we cannot create a path! This
+/// would mean that the aggregation pipeline in question will return no results, in fact, because
+/// the $match stage will never evaluate to true.
+pub(crate) fn get_or_create_schema_for_path_mut(
+    schema: &mut Schema,
+    path: Vec<String>,
+) -> Option<&mut Schema> {
+    let mut schema = Some(schema);
+    for field in path {
+        schema = match schema {
+            Some(Schema::Document(d)) => {
+                if !d.keys.contains_key(&field) {
+                    if !d.additional_properties {
+                        return None;
+                    }
+                    d.keys.insert(field.clone(), Schema::Any);
+                }
+                d.keys.get_mut(&field)
+            }
+            Some(Schema::Any) => {
+                let mut d = schema::Document::any();
+                d.keys.insert(field.clone(), Schema::Any);
+                // this is a wonky way to do this, putting it in the map and then getting it back
+                // out with this match, but it's what the borrow checker forces (we can't keep the
+                // reference across the move of ownership into the Schema::Document constructor).
+                **(schema.as_mut()?) = Schema::Document(d);
+                schema?.get_key_mut(&field)
+            }
+            Some(Schema::AnyOf(schemas)) => {
+                // By first checking to see if there is a Document in the AnyOf, we can avoid
+                // cloning the Document. In general, I expect that the AnyOf will be smaller than
+                // the size of the Document schema, meaning this is more efficient than cloning
+                // even ignoring "constant factors". This is especially true given that cloning
+                // means memory allocation, which is quite a large "constant factor".
+                if !schemas.iter().any(|s| matches!(s, &Schema::Document(_))) {
+                    return None;
+                }
+                // This is how we avoid the clone. By doing a std::mem::take here, we can take
+                // ownership of the schemas, and thus the Document schema. Sadly, we still have to
+                // allocate a BTreeSet::default(), semantically. There is a price to pay for safety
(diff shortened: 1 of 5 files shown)


agg-ast/schema_derivation/src/lib.rs
```diff
@@ -121,6 +121,82 @@ pub(crate) fn get_schema_for_path_mut(
     schema
 }
 
+/// Gets or creates a mutable reference to a specific field or document path in the schema. This
+/// should only be used in $match context where can can refine the schema. For example, we could
```
mattChiaravalloti, line 125:
```suggestion
/// should only be used in $match context where we can refine the schema. For example, we could
```

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -121,6 +121,82 @@ pub(crate) fn get_schema_for_path_mut(
...
+                // even ignoring "constant factors". This is especially true given that cloning
+                // means memory allocation, which is quite a large "constant factor".
+                if !schemas.iter().any(|s| matches!(s, &Schema::Document(_))) {
+                    return None;
+                }
+                // This is how we avoid the clone. By doing a std::mem::take here, we can take ownership
+                // of the schemas, and thus the Document schema. Sadly, we still have to allocate
+                // an empty BTreeSet. There is a price to pay for safety sometimes, but, there is a
+                // good chance the compiler will be smart enough to know that
+                // BTreeSet::default() is never used and optimize it out.
```
mattChiaravalloti, line 170: Is this comment outdated? I don't see a `BTreeSet::default()` call anywhere. Is it implicitly included somewhere?

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1288,19 +1282,7 @@ mod date_ops {
...
-                "day".to_string() => Schema::Any,
-                "hour".to_string() => Schema::Any,
-                "minute".to_string() => Schema::Any,
-                "second".to_string() => Schema::Any,
-                "millisecond".to_string() => Schema::Any,
-            },
-            required: set!("year".to_string(), "month".to_string(), "day".to_string()),
-            ..Default::default()
-        })
+        ref_schema = Schema::Any
```
mattChiaravalloti, line 1285: Ha! Nice, that's a clever use of this in the context of testing.

mattChiaravalloti (approved):
I had been looking at this one when I was tagged on it on the other repo, so I decided to finish up looking at it here even though I wasn't tagged here. Ultimately, I have nothing much to add beyond a typo and a compliment, but I figured I'd finish what I started 🤷‍♀️

agg-ast/schema_derivation/src/lib.rs
terakilobyte: I believe it's from the `take` operation.

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -121,6 +121,82 @@ pub(crate) fn get_schema_for_path_mut(
...
+                let mut d = schemas.into_iter().find_map(|s| {
+                    if let Schema::Document(doc) = s {
+                        // we would have to clone here without the std::mem::take above.
+                        Some(doc)
+                    } else {
+                        None
+                    }
+                })?;
+                if !d.keys.contains_key(&field)
+                    // We can only add keys, if additionalProperties is true.
```
terakilobyte, line 181:
```suggestion
                    // We can only add keys if additionalProperties is true.
```
nittiest of nits

terakilobyte (approved):
LGTM.

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -121,6 +121,82 @@ pub(crate) fn get_schema_for_path_mut(
...
 
+/// Gets or creates a mutable reference to a specific field or document path in the schema. This
+/// should only be used in $match context where can can refine the schema. For example, we could
+/// think a field has type Any, or AnyOf([String, Document]), and a $match stage can only evaluate
+/// to true if that field is specifically a Document. In this case, we can refine that schema to
+/// Document, and this can recurse to any depth in the Schema. Note that this still returns an
+/// Option because if the field is known to have a Schema that cannot be a Document, we cannot
+/// create a path! This would mean that the aggregation pipeline in question will return no
+/// results, in fact, because the $match stage will never evaluate to true.
+pub(crate) fn get_or_create_schema_for_path_mut(
```
jchemburkar, line 132: my only q would be is this limited to $match? Couldn't we also use this for additive operations?

eg lets say a stage adds a field foo.bar with schema A, could we say, give me whatever the schema has in the path [foo, bar]. Then if it is Some(B), I can union A U B to get the new schema, otherwise B. Or am I spacing on a more ergonomic way to do this that is existing?

jchemburkar (approved):

agg-ast/schema_derivation/src/lib.rs
pmeredit: Yes, `take` uses `BTreeSet::default()` for the swap value.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
pmeredit: This is actually how I found out we should have this get_or_create function. This was the original version of this test and it didn't work :)

agg-ast/schema_derivation/src/lib.rs
pmeredit: Comma is required before if!

agg-ast/schema_derivation/src/lib.rs
pmeredit: As mentioned in standup, in most cases we would not want to use this, but there could be additive places where this is needed, so I'm updating the comment.

PR #7, opened 2025-02-04
SQL-2574: Add CLI for translating and displaying results from queries

pmeredit:


cli/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+        current_db.as_str(),
+        query.as_str(),
+        &catalog,
+        mongosql::options::SqlOptions::default(),
+    )?;
+    let print_translation = |pipeline| {
+        println!(
+            "target_db: {},\ntarget_collection: {:?},\nschema:{:?},\npipeline:",
+            translation.target_db, translation.target_collection, translation.result_set_schema
+        );
```
bucaojit, line 76: Try this change, it prints out the schema in a nice readable way.  The pipeline could use a pretty print option if you want, although it is also good the way it is since it's printing a newline for each entry in the pipeline. 
```suggestion
    let print_translation = |pipeline: bson::Bson| -> Result<(), CliError> {
        let schema = serde_json::to_string_pretty(&translation.result_set_schema)
            .map_err(|e| CliError(e.to_string()))?;
        println!(
            "target_db: {},\ntarget_collection: {:?},\nschema:\n{}\npipeline:",
            translation.target_db, translation.target_collection, schema
        );
```

cli/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "cli"
```
bucaojit, line 2: `cli` does match the naming of our other binaries.  I suspect we want customers to use this as well, I'm thinking we'll then want to make this more descriptive like `mongosql-cli`.

cli/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+        long,
+        help = "translation, automatically true if result not set"
+    )]
+    translation: bool,
+    #[arg(
+        short,
+        long,
+        help = "Display the result of this query when applied against the mongodb uri"
+    )]
+    result: bool,
```
bucaojit, line 49: I wasn't too sure what this would do until I ran the executable.  I'm wondering if this would be clearer if named `execute` or `run` with the following `help`:
```
        help = "Execute the query against the MongoDB URI and display the results"
```

bucaojit (changes requested):
Looks great, and will be super helpful to use.  That'll be much easier to use than what I'm currently doing, running the `pipeline_generator`. 

I had a few minor suggestions.

cli/src/main.rs
```diff
@@ -0,0 +1,277 @@
...
+    db: Option<String>,
+    #[arg(index = 1, help = "The query to translate")]
+    query: String,
+    #[arg(
+        short,
+        long,
+        help = "translation, automatically true if result not set"
+    )]
+    translation: bool,
+    // clap visible_aliases not showing up in --help, so we just make manual aliases, sad.
```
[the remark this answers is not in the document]
bucaojit: I like `execute` too.

cli/src/main.rs
```diff
@@ -0,0 +1,277 @@
...
+    T: std::error::Error,
+{
+    fn from(e: T) -> Self {
+        CliError(e.to_string())
+    }
+}
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about=None)]
+struct Cli {
```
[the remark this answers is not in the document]
pmeredit: I'd rather do this in a separate ticket since it's a non-trivial amount of work.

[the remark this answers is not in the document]
pmeredit: Filed https://jira.mongodb.org/browse/SQL-2635

mattChiaravalloti (approved):
lgtm 🖥️

cli/src/main.rs
```diff
@@ -0,0 +1,260 @@
...
+        long,
+        help = "The current database where collections in the query are assumed to live (cross database queries are not supported), default = test"
+    )]
+    db: Option<String>,
+    #[arg(index = 1, help = "The query to translate")]
+    query: String,
+    #[arg(
+        short,
+        long,
+        help = "translation, automatically true if result not set"
```
bucaojit, line 41:
```suggestion
        help = "translation, automatically true if execute not set"
```

bucaojit (approved):
One update to the `help` string. 

Also, I still think it would be helpful to change the binary name to `mongosql-cli` but I'll leave that up to you.

PR #11, opened 2025-02-05
ensure go paths are updated properly

terakilobyte:
I didn't even consider this initially, my bad. I forgot about how unforgiving go paths are. I'll re-release the same tags once this is fixed.


nbagnard (approved):
LGTM!

We only thought about it for the ADF update 🤦‍♀️

PR #12, opened 2025-02-06
SQL-2357: Refine schema derivation for Addition, Subtraction, and Multiplication

mattChiaravalloti:
Migrated from the old private repo.

This PR updates schema derivation for addition, subtraction, and multiplication. It is still technically incomplete but I think at this point I'll file a follow-up ticket for remaining work.

My main goals were:
1. Fix multiplication to properly account for nullability. It previously did not acknowledge when it was possibly nullable and now it does.
2. Fix addition and subtraction with respect to Date arguments. The first implementation I merged under this ticket number too strictly assumed the result type would be Date in certain circumstances. Now, it returns the full set of possible types when an operand MAY or MUST be Date.

The things left to do are more precise numeric handling for Addition and Multiplication. Specifically, if you have `AnyOf(Int, Double)` and `AnyOf(Int, Double)` as operands to addition, the result is `AnyOf(Int, Long, Double)` since there could be `Int-Int`, `Int-Double`, and `Double-Double` pairings. Currently, we'd just say this is `Double`. That's not the end of the world but it is imprecise.

I _could_ address that in this ticket using the new cartesian product stuff I added for subtraction, _but_ it's not as easy for n-ary operators like addition and multiplication as it is for a binary operator like subtraction. I think a follow-up ticket that just sits in the backlog until someone has a week to properly hammer this out is the best course of action.

### DIFF
--- a/agg-ast/schema_derivation/src/schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/schema_derivation.rs
@@ -826,10 +826,18 @@ impl DeriveSchema for UntaggedOperator {
             }
             UntaggedOperatorName::Add => {
                 // Separate any possible Date arguments from non-Date arguments.
-                let arg_schemas = args.iter().map(|arg| arg.derive_schema(state).map(|schema| schema.upconvert_missing_to_null())).collect::<Result<BTreeSet<_>>>()?;
-                let (dates, mut non_dates): (BTreeSet<_>, BTreeSet<_>) = arg_schemas.into_iter().partition(|arg_schema| {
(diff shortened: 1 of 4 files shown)


mattChiaravalloti:
@pmeredit  This is exactly the same as the one you already approved, except I used your suggestion to get `itertools::cartesian_product` working (thanks) and cut down on some of the more frivolous unit testing of that method. Not sure if you want to look again -- it probably isn't necessary. Just tagging @jchemburkar  on this one assuming your approval stands.

jchemburkar (approved):
lgtm! nice work

mattChiaravalloti:
I'll file the follow-up ticket for addition and multiplication precision improvements soon!

PR #14, opened 2025-02-07
SQL-2623: Update spec testing for GROUP BY accumulators to demonstrate document and array self-comparability

mattChiaravalloti:
This PR updates the `group_by.yml` spec query and type constraint tests. For the query tests, I removed the accumulator type-checking tests in favor of the more robust type constraint test suite; I updated the group key test to use an `AnyOf` schema instead of `Document` (since `Document` is now valid). For the type constraint tests, I added tests for distinct versions of all accumulators, including those that already had non-distinct tests. This covers everything described in the design doc for this ticket.

Note that `COUNT(DISTINCT *)` and `COUNT(DISTINCT col1, col2, ...)` will be spec tested as part of their respective implementation tickets.

### DIFF
--- a/tests/spec_tests/type_constraint_tests/group_by.yml
+++ b/tests/spec_tests/type_constraint_tests/group_by.yml
@@ -22,6 +22,7 @@ variables:
 tests:
   # Note: Non-distinct ADD_TO_ARRAY has no type constraints
   - description: Distinct ADD_TO_ARRAY argument must be self-comparable
+    skip_reason: "SQL-2618: enable document and array self-comparison"
     query: "SELECT * FROM foo GROUP BY NULL AS _groupKey1 AGGREGATE ADD_TO_ARRAY(DISTINCT arg1) AS agg"
     valid_types: *selfComparisonValidTypes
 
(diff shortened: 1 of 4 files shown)


tests/spec_tests/type_constraint_tests/group_by.yml
```diff
@@ -2,35 +2,112 @@ variables:
...
+    - { "arg1": [ "REGEX", "NULL", "MISSING" ] }
+    - { "arg1": [ "SYMBOL", "NULL", "MISSING" ] }
+    - { "arg1": [ "TIMESTAMP", "NULL", "MISSING" ] }
+    - { "arg1": [ "MINKEY", "NULL", "MISSING" ] }
+    - { "arg1": [ "MAXKEY", "NULL", "MISSING" ] }
 
 tests:
+  # Note: Non-distinct ADD_TO_ARRAY has no type constraints
+  - description: Distinct ADD_TO_ARRAY argument must be self-comparable
+    skip_reason: "SQL-2618: enable document and array self-comparison"
```
mattChiaravalloti, line 25: I'll unskip all of these as part of this PR as soon as #13 is merged, which should be soon!

terakilobyte (approved):
LGTM. Having a look at the generated tests, we should probably generate the type constraint tests too at some point in the future!

bucaojit (approved):
LGTM Nice work!

PR #16, opened 2025-02-10
SQL-2591: Generate and scan SBOM lite for libmongosqltranslate

EthanHardyMongo:
This PR handles [SQL-2591](https://jira.mongodb.org/browse/SQL-2591) and [SQL-2592](https://jira.mongodb.org/browse/SQL-2592). I decided to do both because Jon already created the SBOM generating and scanning functions, so I just had to create a `task` and `buildvariant` using them. 

Successful run: https://spruce.mongodb.com/version/67abbfcc52bbba0007bea6e9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


Cargo.lock
```diff
@@ -1010,14 +1010,14 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "4.5.24"
+version = "4.5.28"
```
EthanHardyMongo, line 1013: automatic updates

evergreen.yml
```diff
@@ -696,7 +696,7 @@ functions:
           echo "------------------------------------"
 
           # Install CycloneDX CLI and JQ
-          JQ_URL="https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-arm64"
+          JQ_URL="https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-amd64"
           echo "-- Downloading JQ $JQ_URL --"
```
EthanHardyMongo, line 700: I'm not sure why this was set to `arm64` when we were running it on `ubuntu2204-small`. In the JDBC driver and BIC, we're using `jq-linux-amd64`, so I changed it to that.

EthanHardyMongo:
```
WARN merging packages have with different pURLs: "03f0972699b079f8"="pkg:cargo/mongosqltranslate@0.0.0?download_url=file://." vs "03f0972699b079f8"="pkg:cargo/mongosqltranslate@0.0.0?download_url=file://.#src/lib.rs"
```

There's this warning in the SBOM, but it seems like its a recent `grype` bug as shown [here](https://github.com/anchore/grype/issues/2422).

evergreen.yml
pmeredit: I assume this was a typo. This is what irks me about how some institutions use amd64 for x86-64. Yes, I get, AMD created it as an extension to x86, but amd and arm are too similar and easy to mess up :joy:

pmeredit (approved):
looks great! Thanks!

PR #17, opened 2025-02-10
SQL-2370: implement schema derivation for misc stages

jchemburkar:
This pr is much shorter than it looks -- I split the macro for testing derive_schema into two so that it can easily deserialize stages and expressions (could perhaps make it one, if we wanted to get fancy, but it was easier this way).

Two stages of the original seven in this ticket were skipped: $redact (that will be hard and annoying -- certainly worth its own ticket) and $geoNear (it was not going to be possible to construct a meaningfully accurate result_set_schema because the information needed to get the type of the coordinates being used would be imbedded in the index, not in the schema).

### DIFF
--- a/agg-ast/schema_derivation/src/lib.rs
+++ b/agg-ast/schema_derivation/src/lib.rs
@@ -200,6 +200,9 @@ pub(crate) fn get_or_create_schema_for_path_mut(
     schema
 }
 
+// this helper simply checks for document schemas and inserts a field with a given schema into
+// that document, ensuring it is required. It follows the same structure as get_or_create_schema_for_path_mut,
+// except that we ignore Schema::Any's.
 fn insert_required_key_into_document_helper(
     mut schema: Option<&mut Schema>,
     field_schema: Schema,
@@ -210,27 +213,19 @@ fn insert_required_key_into_document_helper(
             if !d.keys.contains_key(&field) {
                 d.keys.insert(field.clone(), field_schema);
             }
+            // even if the document already included the key, we want to mark it as required.
+            // this enables nested field paths to be marked as required down to the required
+            // key being inserted.
             d.required.insert(field.clone());
             d.keys.get_mut(&field)
         }
         Some(Schema::AnyOf(schemas)) => {
-            // By first checking to see if there is a Document in the AnyOf, we can avoid
-            // cloning the Document. In general, I expect that the AnyOf will be smaller than
-            // the size of the Document schema, meaning this is more efficient than cloning
-            // even ignoring "constant factors". This is especially true given that cloning
-            // means memory allocation, which is quite a large "constant factor".
             if !schemas.iter().any(|s| matches!(s, &Schema::Document(_))) {
                 return None;
             }
-            // This is how we avoid the clone. By doing a std::mem::take here, we can take
-            // ownership of the schemas, and thus the Document schema. Sadly, we still have to
-            // allocate a BTreeSet::default(), semantically. There is a price to pay for safety
-            // sometimes, but, there is a good chance the compiler will be smart enough to know
-            // that BTreeSet::default() is never used and optimize it out.
             let schemas = std::mem::take(schemas);
             let mut d = schemas.into_iter().find_map(|s| {
                 if let Schema::Document(doc) = s {
-                    // we would have to clone here without the std::mem::take above.
                     Some(doc)
                 } else {
                     None
@@ -239,10 +234,8 @@ fn insert_required_key_into_document_helper(
             if !d.keys.contains_key(&field) {
                 d.keys.insert(field.clone(), Schema::Any);
             }
+            // see comment above for why we require the field even if it existed in the document
             d.required.insert(field.clone());
-            // this is a wonky way to do this, putting it in the map and then getting it back
-            // out with this match, but it's what the borrow checker forces (we can't keep the
-            // reference across the move of ownership into the Schema::Document constructor).
             **(schema.as_mut()?) = Schema::Document(d);
             schema?.get_key_mut(&field)
         }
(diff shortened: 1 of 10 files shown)


agg-ast/schema_derivation/src/schema_derivation_tests/expression.rs
```diff
@@ -13,110 +13,110 @@ use std::collections::BTreeMap;
 mod literal {
     use super::*;
 
-    test_derive_schema!(
```
jchemburkar, line 16: ignore this file -- just a naming change on the test macro

agg-ast/schema_derivation/src/schema_derivation_tests/tagged_ops.rs
```diff
@@ -10,7 +10,7 @@ use std::collections::BTreeMap;
 mod misc_ops {
     use super::*;
 
-    test_derive_schema!(
```
jchemburkar, line 13: ignore this file -- just a naming change on the test macro

agg-ast/schema_derivation/src/schema_derivation_tests/untagged_ops.rs
```diff
@@ -46,35 +46,35 @@ macro_rules! test_type_conversion_op {
 
 mod array_ops {
     use super::*;
-    test_derive_schema!(
```
jchemburkar, line 49: ignore this file -- just a naming change on the test macro

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -198,6 +200,72 @@ pub(crate) fn get_or_create_schema_for_path_mut(
     schema
 }
 
+// this helper simply checks for document schemas and inserts a field with a given schema into
+// that document, ensuring it is required. It follows the same structure as get_or_create_schema_for_path_mut,
+// except that we ignore Schema::Any's.
+fn insert_required_key_into_document_helper(
```
jchemburkar, line 206: After trying to retrofit the existing `get_or_create_schema_for_path_mut` function, I found creating a separate helper to be cleaner in the end. This will create the whole path as required, and only work with documents (it ignores Schema::Any). This prevents the nested documents being created from being `additional_properties: true`, which `get_or_create_schema_for_path_mut` does because it inserts new keys with `Schema::Any`

If a use case arises in the future, we can make the behavior about required optional in the future.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -35,21 +37,203 @@ pub(crate) struct ResultSetState<'a> {
     pub null_behavior: Satisfaction,
 }
 
+fn derive_schema_for_pipeline(pipeline: Vec<Stage>, state: &mut ResultSetState) -> Result<Schema> {
```
jchemburkar, line 40: this is a generic helper that I believe we will also be able to use for the overall api in milestone 4!

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1 +1,347 @@
...
+        })),
+        input = r#"{"$documents": [
+                     {"a": 1, "b": 2},
+                     {"a": "yes", "b": null},
+                     {"a": {"b": {"c": true}}}
+        ]}"#
+    );
+}
+
+mod facet {
```
jchemburkar, line 133: To make these tests sensible / follow existing examples, I thought it might be best to just wait for more stages to be implemented. I left them here though so that the future implementer doesn't need to copy them in.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -35,21 +37,210 @@ pub(crate) struct ResultSetState<'a> {
...
+            Ok(schema?.unwrap_or(Schema::Document(Document::default())))
+        }
+
+        fn facet_derive_schema(
+            facet: &LinkedHashMap<String, Vec<Stage>>,
+            state: &mut ResultSetState,
+        ) -> Result<Schema> {
+            // facet contains key - value pairs where the key is the name of a field in the output document,
+            // and the value is an aggregation pipeline. We can use the same generic helper for aggregating over a whole pipeline
+            // to get the sch ema for each field, cloning the incoming state.
```
pmeredit, line 135:
```suggestion
            // to get the schema for each field, cloning the incoming state.
```

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -35,21 +37,210 @@ pub(crate) struct ResultSetState<'a> {
...
+                    }
+                }
+                _ => None,
+            };
+            if let Some(path) = path {
+                if let Some(s) = get_schema_for_path_mut(&mut state.result_set_schema, path) {
+                    // the schema of the field being unwound goes from type Array[X] to type X
+                    match s {
+                        Schema::Array(a) => {
+                            *s = *a.clone();
```
pmeredit, line 185: We can do an `std::mem::take` for sure here.

pmeredit: Oh, we may not have default defined for Schema. I forget. If not, just derive Default for Schema with Unsat at the Default.

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
pmeredit: I'm a fan of this

pmeredit (approved):
This looks great! Just a few small things to fix up that shouldn't need retagging me.

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -198,6 +200,72 @@ pub(crate) fn get_or_create_schema_for_path_mut(
...
+    mut schema: Option<&mut Schema>,
+    field_schema: Schema,
+    field: String,
+) -> Option<&mut Schema> {
+    match schema {
+        Some(Schema::Document(d)) => {
+            if !d.keys.contains_key(&field) {
+                d.keys.insert(field.clone(), field_schema);
```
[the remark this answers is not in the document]
jchemburkar: I'll move it, yes. It has only come up in the context of schema derivation, and I am hopeful these helpers will be the end of document manipulation work for the sake of this epic. But having this just be impl'd on the Document itself and be well commented surely can't hurt.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -35,21 +37,210 @@ pub(crate) struct ResultSetState<'a> {
     pub null_behavior: Satisfaction,
 }
 
+fn derive_schema_for_pipeline(pipeline: Vec<Stage>, state: &mut ResultSetState) -> Result<Schema> {
+    pipeline.iter().try_for_each(|stage| {
+        state.result_set_schema = stage.derive_schema(state)?;
+        Ok(())
+    })?;
+    Ok(state.result_set_schema.clone())
```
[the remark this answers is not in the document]
jchemburkar: Yes, I think you're right. if for some reason we need to in the future we can always change it back.

PR #19, opened 2025-02-11
SQL-2361: Add match schema derivation for string ops

pmeredit:
Had to add expression derivation for replace also, to make the eq null tests work

### DIFF
--- a/agg-ast/ast/src/definitions.rs
+++ b/agg-ast/ast/src/definitions.rs
@@ -1115,6 +1115,12 @@ pub enum UntaggedOperatorName {
     Type,
 }
 
+impl std::fmt::Display for UntaggedOperatorName {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", serde_json::to_string(self).unwrap())
+    }
+}
+
 impl TryFrom<&str> for UntaggedOperatorName {
     type Error = String;
 
@@ -1150,6 +1156,10 @@ pub enum TaggedOperator {
     Like(Like),
     #[serde(rename = "$regexMatch")]
     Regex(RegexAggExpression),
+    #[serde(rename = "$regexFind")]
+    RegexFind(RegexAggExpression),
+    #[serde(rename = "$regexFindAll")]
+    RegexFindAll(RegexAggExpression),
     #[serde(rename = "$sqlDivide")]
     SQLDivide(SQLDivide),
     #[serde(rename = "$trim")]
@@ -1158,16 +1168,12 @@ pub enum TaggedOperator {
     LTrim(Trim),
     #[serde(rename = "$rtrim")]
     RTrim(Trim),
+    #[serde(rename = "$replaceAll")]
+    ReplaceAll(Replace),
+    #[serde(rename = "$replaceOne")]
+    ReplaceOne(Replace),
 
     // Subquery Operators (extended from MQL)
-    #[serde(rename = "$regexFind")]
-    RegexFind(RegexFind),
-    #[serde(rename = "$regexFindAll")]
-    RegexFindAll(RegexFindAll),
-    #[serde(rename = "$replaceAll")]
-    ReplaceAll(ReplaceAll),
-    #[serde(rename = "$replaceOne")]
-    ReplaceOne(ReplaceOne),
     #[serde(rename = "$subquery")]
     Subquery(Subquery),
     #[serde(rename = "$subqueryComparison")]
@@ -1415,35 +1421,7 @@ pub struct MinNArrayElement {
 }
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct RegexMatch {
-    pub input: Box<Expression>,
-    pub regex: Box<Expression>,
-    pub options: Option<Box<Expression>>,
-}
-
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct RegexFind {
-    pub input: Box<Expression>,
-    pub regex: Box<Expression>,
-    pub options: Option<Box<Expression>>,
-}
-
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct RegexFindAll {
-    pub input: Box<Expression>,
-    pub regex: Box<Expression>,
-    pub options: Option<Box<Expression>>,
-}
-
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct ReplaceAll {
-    pub input: Box<Expression>,
-    pub find: Box<Expression>,
-    pub replacement: Box<Expression>,
-}
-
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct ReplaceOne {
+pub struct Replace {
     pub input: Box<Expression>,
     pub find: Box<Expression>,
     pub replacement: Box<Expression>,
--- a/agg-ast/schema_derivation/src/match_schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/match_schema_derivation.rs
@@ -1,13 +1,13 @@
 use crate::{
     get_or_create_schema_for_path_mut, maybe_any_of,
     negative_normalize::{NegativeNormalize, DECIMAL_ZERO},
-    promote_missing, schema_difference, schema_for_bson, schema_for_type_str, DeriveSchema, Result,
-    ResultSetState,
+    promote_missing, schema_difference, schema_for_bson, schema_for_type_str, DeriveSchema, Error,
+    Result, ResultSetState,
 };
 use agg_ast::definitions::{
     DateExpression, DateFromParts, DateFromString, DateToParts, DateToString, Expression, Let,
     MatchBinaryOp, MatchExpr, MatchExpression, MatchField, MatchLogical, MatchNotExpression,
-    MatchStage, Ref, Switch, TaggedOperator, UntaggedOperator,
+    MatchStage, Ref, RegexAggExpression, Replace, Switch, TaggedOperator, Trim, UntaggedOperator,
 };
 use bson::Bson;
 use mongosql::{
@@ -148,7 +148,7 @@ impl DeriveSchema for MatchStage {
         state.result_set_schema = promote_missing(&state.result_set_schema);
         for expr in self.expr.iter() {
             let expr = expr.get_negative_normal_form();
-            expr.match_derive_schema(state);
+            expr.match_derive_schema(state)?;
         }
         Ok(state.result_set_schema.clone())
     }
@@ -158,11 +158,11 @@ trait MatchConstrainSchema {
     // match_derive_schema does not need to return Schema because it only applies constraints
     // to already existing Schema. It also does not need to return a Result because it is infallible
     // (modulo a panic that can only result due to programmer error).
-    fn match_derive_schema(&self, state: &mut ResultSetState);
+    fn match_derive_schema(&self, state: &mut ResultSetState) -> Result<()>;
 }
 
 impl MatchConstrainSchema for MatchExpression {
-    fn match_derive_schema(&self, state: &mut ResultSetState) {
+    fn match_derive_schema(&self, state: &mut ResultSetState) -> Result<()> {
         match self {
             MatchExpression::Expr(e) => e.match_derive_schema(state),
             MatchExpression::Misc(_) => todo!(),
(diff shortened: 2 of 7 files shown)


agg-ast/ast/src/definitions.rs
```diff
@@ -1158,16 +1162,12 @@ pub enum TaggedOperator {
     LTrim(Trim),
     #[serde(rename = "$rtrim")]
     RTrim(Trim),
-
-    // Subquery Operators (extended from MQL)
-    #[serde(rename = "$regexFind")]
-    RegexFind(RegexFind),
-    #[serde(rename = "$regexFindAll")]
-    RegexFindAll(RegexFindAll),
     #[serde(rename = "$replaceAll")]
```
pmeredit, line 1165: There was no reason to use different structs here since they have the same fields. Makes the code cleaner

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1085 @@ mod misc_ops {
...
+    test_derive_schema_for_match_stage!(
+        strcasecmp,
+        expected = Ok(Schema::Document(Document {
+            keys: map! {
+                "foo".to_string() => Schema::Atomic(Atomic::String),
+            },
+            required: set!("foo".to_string()),
+            ..Default::default()
+        })),
+        input = r#"{"$match": {"$expr": {"$strcasecmp": ["$foo", "$bar"]}}}"#,
```
jchemburkar, line 1762: what happens to $bar here -- should it not be in the result set? maybe we can change this (and the `test_derive_schema_for_binary_string_expression_match`) to use the `starting_schema` argument instead and have foo be something like any, and have bar be ... something else, maybe an anyof of something?

jchemburkar (changes requested):
Changes look great, and the macros you created for testing are awesome. Just a few minor things to clarify (I also still need to scan through the rest of the test cases, but dont expect anything else notable).

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -1135,24 +1230,125 @@ impl MatchConstrainSchema for Expression {
...
+                u.args[0].match_derive_schema(state);
+            }
+        }
+
+        fn match_derive_replace(r: &Replace, state: &mut ResultSetState) {
+            let arg_schema = match state.null_behavior {
+                Satisfaction::Not => Schema::Atomic(Atomic::String),
+                Satisfaction::May | Satisfaction::Must => STRING_OR_NULLISH.clone(),
+            };
+            macro_rules! handle_arg {
```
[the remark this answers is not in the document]
pmeredit: Definitely, but these macros are actually capturing local state, so they are just being used as lazy closures.

agg-ast/schema_derivation/src/schema_derivation_tests/tagged_ops.rs
```diff
@@ -165,6 +165,18 @@ mod array_ops {
         )),
         input = r#"{"$reduce": {"input": "hello", "initialValue": "stuff", "in": "foo"}}"#
     );
+    test_derive_schema!(
```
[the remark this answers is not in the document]
pmeredit: I added some that test the null output. We don't actually constrain the args to String since this is the inductive schema where the pipeline would just fail if the args aren't strings (or null).

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
mattChiaravalloti: I was just writing the same comment in the macro above but stopped since I saw this comment. My understanding is that since `bar` is not in the initial schema at all, it is assumed to not be in the output schema. Although, if it is guaranteed to be missing I feel like that should have some impact on the schema derivation -- the expression should definitely resolve to `null` if one of the arguments is definitely `missing`, right?

In terms of the test, agreed that we should make a change. Either update to use `starting_schema` or use a constant value instead of a field ref in the expression.

mattChiaravalloti (changes requested):

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -846,6 +846,99 @@ impl MatchConstrainSchema for Expression {
...
+                        }
+                    }
+                } else {
+                    arg.match_derive_schema(state);
+                }
+            });
+        }
+
+        fn match_derive_index_of(u: &UntaggedOperator, state: &mut ResultSetState) {
+            assert!(u.args.len() > 1);
```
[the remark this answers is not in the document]
pmeredit: I think maybe we should change these to errors. The goal was really just to avoid bounds checking. I'm not sure if mongo even lets you store a pipeline with missing args, but since it may, it's probably worth making these errors.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1085 @@ mod misc_ops {
...
+                    Schema::Atomic(Atomic::Null),
+                )),
+                "bar".to_string() => Schema::Atomic(Atomic::String),
+                "car".to_string() => Schema::Atomic(Atomic::Integer),
+            },
+            required: set!("bar".to_string(), "car".to_string()),
+            ..Default::default()
+        })),
+        input = r#"{"$match": {"$expr": {"$eq": [null, {"$indexOfBytes": ["$foo", "$bar", "$car"]}]}}}"#,
+        starting_schema = Schema::Document(Document {
```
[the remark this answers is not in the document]
pmeredit: I switched styles half way through since I saw the same issue.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1085 @@ mod misc_ops {
...
+        input = r#"{"$match": {"$expr": {"$eq": [null, {"$regexMatch": {"input": "$foo", "regex": "$bar"}}]}}}"#,
+        starting_schema = Schema::Document(Document {
+            keys: map! {
+                "foo".to_string() => Schema::Any,
+                "bar".to_string() => Schema::Any,
+            },
+            required: set!("foo".to_string(), "bar".to_string()),
+            ..Default::default()
+        })
+    );
```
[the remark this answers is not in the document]
pmeredit: We worried about saying definitively that a pipeline will return 0 documents, since we can never be completely sure that our input catalog is correct.

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -1135,24 +1230,125 @@ impl MatchConstrainSchema for Expression {
...
+                    }
+                };
+            }
+            handle_arg!(*r.input, arg_schema.clone());
+            handle_arg!(*r.find, arg_schema.clone());
+            handle_arg!(*r.replacement, arg_schema);
+        }
+
+        fn match_derive_regex(r: &RegexAggExpression, state: &mut ResultSetState) {
+            // regex cannot ever be null, so the null_behavior is irrelevant
```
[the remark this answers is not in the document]
pmeredit: I wish we didn't have such bizarre random semantics in MQL.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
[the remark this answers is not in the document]
mattChiaravalloti: ooh ok makes sense.

agg-ast/schema_derivation/src/match_schema_derivation.rs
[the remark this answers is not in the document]
mattChiaravalloti: Agreed. I'm trying to wrap my head around why this one would evaluate to null while the others don't and I can't come up with anything reasonable (yet). Maybe it's for not good reason at all 🤷‍♀️ could just be an accident that they behave differently.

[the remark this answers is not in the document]
pmeredit: Well it evaluates to empty array. It's fine :joy:

agg-ast/schema_derivation/src/match_schema_derivation.rs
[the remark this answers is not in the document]
pmeredit: Ok, this was a big lift, but I think it was worth it

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
jchemburkar: yeah, I guess I wasn't thinking or communicating super clearly in this comment either. Because this stage is of course valid, and the type of `$bar` is simply missing. Which is totally fine, but even setting it as `null` (not required) should function the same for most of our cases and be easier to read

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -1135,30 +1285,181 @@ impl MatchConstrainSchema for Expression {
...
+                ($input:expr, $sch:expr) => {
+                    if let Expression::Ref(ref reference) = $input {
+                        intersect_if_exists(reference, state, $sch);
+                    } else {
+                        $input.match_derive_schema(state)?;
+                    }
+                };
+            }
+            handle_arg!(*t.input, arg_schema.clone());
+            if let Some(chars) = &t.chars {
```
jchemburkar, line 1451: we've had this discussion before and I think its pretty outside the scope of this epic. But we could technically be more precise in these situations. What if one of the args must not be null (eg it is a literal string). Then, when the null behavior is "must be null", we might be able to infer a field reference _must_ be null. 

Nothing to see here, just noting it. Perhaps something that we should make sure to include in the closeout doc for posterity.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1099 @@ mod misc_ops {
...
+        expected = Ok(Schema::Document(Document {
+            keys: map! {
+                "foo".to_string() => Schema::Atomic(Atomic::String),
+                "bar".to_string() => Schema::AnyOf(set!(
+                    Schema::Atomic(Atomic::String),
+                    Schema::Atomic(Atomic::Regex),
+                )),
+                "car".to_string() => Schema::Atomic(Atomic::String),
+            },
+            required: set!("foo".to_string(), "bar".to_string(), "car".to_string()),
```
mattChiaravalloti, line 2296:
```suggestion
                "car".to_string() => STRING_OR_NULL.clone(),
            },
            required: set!("foo".to_string(), "bar".to_string()),
```
It's possible for the `options` to be `null` or `missing` and this to still be valid, so we can't rule those types out for `car`.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1099 @@ mod misc_ops {
...
```
mattChiaravalloti, line 2321:
```suggestion
                "car".to_string() => STRING_OR_NULL.clone(),
            },
            required: set!("foo".to_string(), "bar".to_string()),
```

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1099 @@ mod misc_ops {
...
+    test_derive_schema_for_match_stage!(
+        regexfindall_opts_eq_null,
+        expected = Ok(Schema::Document(Document {
+            keys: map! {
+                "foo".to_string() => Schema::Atomic(Atomic::String),
+                "bar".to_string() => Schema::AnyOf(set!(
+                    Schema::Atomic(Atomic::String),
+                    Schema::Atomic(Atomic::Regex),
+                )),
+                "car".to_string() => Schema::Atomic(Atomic::String),
```
jchemburkar, line 2526: nit: I think car can actually be nullish while still being valid -- ie its `STRING_OR_NULLISH` in all nullability contexts.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1099 @@ mod misc_ops {
...
+            },
+            required: set!("foo".to_string(), "bar".to_string(), "car".to_string()),
```
mattChiaravalloti, line 2503:
```suggestion
                "car".to_string() => STRING_OR_NULL.clone(),
            },
            required: set!("foo".to_string(), "bar".to_string()),
```

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -1675,3 +1675,1099 @@ mod misc_ops {
...
```
mattChiaravalloti, line 2528:
```suggestion
                "car".to_string() => STRING_OR_NULL.clone(),
            },
            required: set!("foo".to_string(), "bar".to_string()),
```

mattChiaravalloti (approved):
The error change is huge 💯 Other than the possibly nullish `options` fields, this looks good! No need to retag me after updating the tests and the implementation to match.

jchemburkar (approved):
The results in match_derive_schema are awesome. I went through the tests and only noticed one correctness thing, but otherwise these look great. Awesome stuff 🤠

agg-ast/schema_derivation/src/match_schema_derivation.rs
pmeredit: I think so. We could be more precise, but a lot of the precision will result in cartesian blowups. But maybe that's fine for operators that have 1-4 args.

PR #21, opened 2025-02-12
SQL-2621: Implement COUNT(DISTINCT *)

mattChiaravalloti:
This PR implements support for `COUNT(DISTINCT *)`. It also updates the implementation of `COUNT(*)`, `COUNT(<expr>)`, and `COUNT(DISTINCT <expr>)` to be consistent in terms of implementation and behavior.

There's a lot going on in this one:
- I refactored the `desugarer/accumulators.rs` file to better support this accumulator, and I set it up for SQL-2622 (support multi-column `COUNT` aggregations).
- As noted in the ticket, I removed the `CountDistinctStarNotSupported` error from the `mir/schema` errors library, and updated the translation pipeline to support this expression.
- I added spec tests for `COUNT(*)` and `COUNT(DISTINCT *)`. We don't have an explicit `COUNT(DISTINCT <expr>)` test in the spec query tests, but we do have a test that implicitly covers this. **Do you think I should add a spec test for that?**
- We'll need to do some follow-up work to make sure `COUNT(<doc column>)` and `COUNT(DISTINCT <doc column>)` behave consistently with `COUNT(col1, col2, ...)` and `COUNT(DISTINCT col1, col2, ...)`. Specifically, we'll likely want those to follow the same rules (i.e. omit empty docs and docs where all values are null). **Should I add that as a comment to SQL-2622, or should I file a separate ticket to deal with it?**

### DIFF
--- a/mongosql/src/air/desugarer/accumulators.rs
+++ b/mongosql/src/air/desugarer/accumulators.rs
@@ -81,7 +81,7 @@ impl AccumulatorsDesugarerVisitor {
     /// each case are applied to the $addToSet or $sum, depending on the value of distinct.
     fn rewrite_count(count_expr: &AccumulatorExpr) -> (AccumulatorExpr, ProjectItem) {
         let new_acc_expr = match count_expr.arg.as_ref() {
-            Variable(v) if v.parent.is_none() && v.name == ROOT_NAME.to_string() => {
+            Variable(v) if v.parent.is_none() && v.name == *ROOT_NAME => {
                 Self::rewrite_count_star(count_expr.distinct, count_expr.alias.clone())
             }
             Document(_doc) => {
--- a/mongosql/src/air/desugarer/accumulators.rs
+++ b/mongosql/src/air/desugarer/accumulators.rs
@@ -79,65 +79,23 @@ impl AccumulatorsDesugarerVisitor {
     /// and follow it with a projection assignment that does the counting (via $size). If it is
     /// non-distinct, we rewrite the accumulator to use $sum. The conditions described above for
     /// each case are applied to the $addToSet or $sum, depending on the value of distinct.
-    fn rewrite_count(count_expr: AccumulatorExpr) -> (AccumulatorExpr, ProjectItem) {
+    fn rewrite_count(count_expr: &AccumulatorExpr) -> (AccumulatorExpr, ProjectItem) {
         let new_acc_expr = match count_expr.arg.as_ref() {
-            // *
             Variable(v) if v.parent.is_none() && v.name == ROOT_NAME.to_string() => {
-                let (function, arg) = if count_expr.distinct {
-                    (AggregationFunction::AddToSet, Box::new(ROOT.clone()))
-                } else {
-                    (
-                        AggregationFunction::Sum,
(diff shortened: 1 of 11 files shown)


mongosql/src/air/desugarer/accumulators.rs
mattChiaravalloti, line 1: I did a fairly large refactor here since the previous implementation was not very amenable to expanding for `COUNT(DISTINCT? *)` support (and would not be good for the follow-up `COUNT(DISTINCT? col1, col2, ...)` support).

One major thing I also changed is that instead of writing the count as a `$reduce` in the `$project` stage for `COUNT(DISTINCT col)`, we conditionally add the column value to the set the same way we conditionally summed 1 or 0 for the non-distinct version. This leaves the `$project` expression to be a simple `$size`. I personally found this consistency in end-result useful for reducing the amount of code needed to support this (and, eventually, `COUNT(DISTINCT? <multi-column>)`.

mongosql/src/air/desugarer/testdata/desugar_accumulators.yml
```diff
@@ -69,43 +69,176 @@ tests:
...
+      }
       - {
         "$project":
           {
             "_id": 1,
+            "acc": { "$size": "$acc" },
+          },
+      }
+
+  - name: "desugar non-distinct $sqlCount multi-column arg"
```
mattChiaravalloti, line 98: Adding these 2 skipped tests really helped me understand the subtleties of this problem. It also helped to reveal the issue I discussed offline with Oliver re: `*` should be unconditional, and non-`*` should be conditional. Since I wrote the tests for understanding, I figured it was best just to leave them in with a `skip_reason`.

tests/spec_tests/query_tests/group_by.yml
mattChiaravalloti, line 1: When I started this ticket, as I was writing the spec tests, I found it difficult to keep scrolling through all the data to see what already existed. I think we wrote this a very long time ago when we were unsure how the rust yaml and bson parsers worked. We do not need to specify `$numberInt` for very small numbers, so I removed that from the majority of data, which allowed this section to become more concise and, in my opinion, readable.

tests/spec_tests/query_tests/group_by.yml
```diff
@@ -581,6 +538,22 @@ tests:
       - { "": { "a": { "$numberInt": "2" }, "gcount": { "$numberInt": "2" } } }
       - { "": { "a": null, "gcount": { "$numberInt": "0" } } }
 
+  - description: COUNT(*) correctness test unconditionally count rows
```
bucaojit, line 541: Nice, thanks for adding these tests to show the COUNT(*) and COUNT(DISTINCT *) match behavior in terms of unconditionally counting.

mongosql/src/air/desugarer/accumulators.rs
bucaojit: Great, it does look like an improvement and will help with the multi-column count distinct.

bucaojit (approved):
LGTM.  Well done. 🚀

EthanHardyMongo (approved):
LGTM! Amazing commenting through out!

PR #22, opened 2025-02-13
SQL-2619: Implement SELECT DISTINCT

bucaojit:


errors.md
```diff
@@ -46,7 +46,6 @@ The following errors occur when something goes wrong while converting the SQL qu
 |---------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | [Error 3002](#error-3002) | A SELECT list with multiple values cannot contain a non-namespaced `*` (i.e., `SELECT a, *, b FROM myTable` is not supported). A non-namespaced `*` must be used by itself.                                  |
 | [Error 3004](#error-3004) | The array data source contains an identifier. Array data sources must be constant.                                                                                                                           |
-| [Error 3005](#error-3005) | SELECT DISTINCT is not allowed.                                                                                                                                                                              |
```
mattChiaravalloti, line 49: Good catch updating this! It reminded me to do the same on #21. There is also a longer description of the error down on line 222 we should remove.

tests/spec_tests/query_tests/distinct.yml
mattChiaravalloti, line 1: Nice. I think it's a good idea making this a separate file. Do you envision doing the distinct `UNION` tests in this file, too?

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,132 @@
...
+        "required": ["_id", "a", "b"],
+        "additionalProperties": false,
+        "properties": {
+          "_id":   { "bsonType": "int" },
+          "a":     { "bsonType": "int" },
+          "b":     { "bsonType": "int" },
+          "c": {
+            "anyOf": [
+              { "bsonType": "bool" },
+              { "bsonType": "null" }
```
mattChiaravalloti, line 38: I'm glad to see this works without the need for `!!str`. I thought that was still a requirement but it appears not!

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,132 @@
...
+  {
+    "db": {
+      "foo": {
+        "bsonType": "object",
+        "required": ["_id", "a", "b"],
+        "additionalProperties": false,
+        "properties": {
+          "_id":   { "bsonType": "int" },
+          "a":     { "bsonType": "int" },
+          "b":     { "bsonType": "int" },
```
mattChiaravalloti, line 34: I like this formatting a lot 💯

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,132 @@
...
+            "bsonType": "array",
+            "items": { "bsonType": "int" }
+          }
+        }
+      }
+    }
+  }
+
+tests:
+  - description: "SELECT DISTINCT * from distinct ensures unique rows for entire document"
```
mattChiaravalloti, line 84:
```suggestion
  - description: "SELECT DISTINCT * ensures unique rows for entire document"
```
What does "from distinct" mean here? I think the test description works when we drop those words, but open to adding them back in some way if you had something specific in mind.

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,132 @@
...
+          "c": {
+            "anyOf": [
+              { "bsonType": "bool" },
+              { "bsonType": "null" }
+            ]
+          },
+          "d": {
+            "anyOf": [
+              { "bsonType": "string" },
+              { "bsonType": "null" }
```
mattChiaravalloti, line 44: I don't think we need to say `c` and `d` are possibly `null` since we don't have any data to reflect that. By omitting them from `required`, we already indicate they may be `missing`. Is there a reason you included this in the schema for these fields?

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,132 @@
...
+      - { 'foo': { "_id": 3, "a": 2, "b": 2 } }
+      - { 'foo': { "_id": 4, "a": 1, "b": 2 } }
+      - { 'foo': { "_id": 5, "a": 2, "b": 2, "c": true, "d": "a", "e": null } }
+      - { 'foo': { "_id": 6, "a": 2, "b": 2, "c": true, "d": "a", "e": null } }
+
+  - description: "SELECT DISTINCT * on derived table omitting _id selects unique rows"
+    query: |
+      SELECT DISTINCT *
+      FROM (
+        SELECT a, b, c
```
mattChiaravalloti, line 98: This comment just contains a few musings about the implications of how we support distinct -- no action is required as a result of this comment.

That's funny that we actually can't quite demonstrate that `{a: 1, b: 1}` is distinct from `{b: 1, a: 1}` via `SELECT DISTINCT *`. `*` will always include an `_id` that is unique, and a derived table will always specify the same order... I suppose we could make a view that doesn't include an `_id` but this test framework isn't set up to support that so I'm not sure it's worth it right now.

Also, now that I'm thinking about it, it's kinda weird that MongoSQL will declare those two as distinct but any BI tool would display them as essentially equivalent 💀 not sure if that's something we'll need to worry about though, given the constraints of the previous paragraph. It'll really only be a thing if people try to do `SELECT DISTINCT *` from views that don't produce unique fields in every document.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,32 +953,89 @@ impl<'a> Algebrizer<'a> {
...
-            // All normal Expressions will be mapped as Datasource Bottom, and all Substars will be mapped
-            // as their name as a Datasource.
-            ast::SelectBody::Values(exprs) => {
-                self.algebrize_select_values_body(exprs, source, is_add_fields)
-            }
         }
     }
 
+    fn build_distinct_star_project_stage(&self, source: mir::Stage) -> Result<mir::Stage> {
+        source.schema(&self.schema_inference_state())?;
```
mattChiaravalloti, line 990: Is this a typo? We immediately do this again on line 992.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,32 +953,89 @@ impl<'a> Algebrizer<'a> {
...
+            })
+            .collect::<Vec<(String, Key)>>();
+
+        let (table_name, _) = ds_names.remove(0);
+        let group_stage = mir::Stage::Group(mir::Group {
+            source: Box::new(source),
+            keys: vec![OptionallyAliasedExpr::Aliased(AliasedExpr {
+                alias: table_name.clone(),
+                expr: Expression::Reference(ReferenceExpr {
+                    key: Key::named(table_name.as_str(), self.scope_level),
```
mattChiaravalloti, line 1008: Shouldn't we be using the `key` from line 997 here?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,32 +953,89 @@ impl<'a> Algebrizer<'a> {
...
+        let mut ds_names = source_result_set
+            .schema_env
+            .keys()
+            .filter_map(|key| match &key.datasource {
+                DatasourceName::Named(n) => Some((n.clone(), key.clone())),
+                DatasourceName::Bottom => None,
+            })
+            .collect::<Vec<(String, Key)>>();
+
+        let (table_name, _) = ds_names.remove(0);
```
mattChiaravalloti, line 1002: Why are we only including the first one? What if the source was a join? Shouldn't we group on all datasources as group keys?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,32 +953,89 @@ impl<'a> Algebrizer<'a> {
...
+    fn build_distinct_star_project_stage(&self, source: mir::Stage) -> Result<mir::Stage> {
+        source.schema(&self.schema_inference_state())?;
+
+        let source_result_set = source.schema(&self.schema_inference_state())?;
+        let mut ds_names = source_result_set
+            .schema_env
+            .keys()
+            .filter_map(|key| match &key.datasource {
+                DatasourceName::Named(n) => Some((n.clone(), key.clone())),
+                DatasourceName::Bottom => None,
```
mattChiaravalloti, line 998: Why are we ignoring the Bottom datasource? I think I agree it is not possible for the source of a SELECT to produce one but I'm not sure silently ignoring it is the right move.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +477,61 @@ impl<'a> Algebrizer<'a> {
...
+
+    fn build_distinct_values_group_stage(
+        project_tuple: &BindingTuple<Expression>,
+        source_stage: mir::Stage,
+        scope_level: u16,
+    ) -> mir::Stage {
+        let mut group_keys: Vec<OptionallyAliasedExpr> = Vec::new();
+
+        for (key, expr) in project_tuple.iter() {
+            if key.datasource == DatasourceName::Bottom && key.scope == scope_level {
```
mattChiaravalloti, line 508: I think checking the scope is correct -- we can try some unit or e2e tests that use `SELECT DISTINCT` in subquery contexts to test that -- but I'm not convinced we should only be considering the `Bottom` datasource.

Consider `SELECT DISTINCT VALUES foo.*, bar.* FROM foo JOIN bar`. There is no `Bottom` datasource in that query. Instead, there are 2 `Named` datasources that we should use as the group keys.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +477,61 @@ impl<'a> Algebrizer<'a> {
...
+                                key: key.clone(),
+                            })),
+                            field: field_name.to_string(),
+                            is_nullable: field_expr.is_nullable(),
+                        });
+                        let group_key = OptionallyAliasedExpr::Aliased(AliasedExpr {
+                            alias: alias.clone(),
+                            expr: group_key_expr,
+                        });
+                        group_keys.push(group_key);
```
mattChiaravalloti, line 523: We shouldn't need to depend on the `expr` being a document and we shouldn't need to group by the individual fields. Again, consider my example above, `SELECT DISTINCT VALUES foo.*, bar.* FROM foo JOIN bar`. That should translate roughly to something like:
```
[
  { $project: { _id: 0, foo: "$$ROOT" }},
  { $join: { collection: "bar", join_type: "inner", pipeline: [{ $project: { _id: 0, bar: "$$ROOT" } }] } },
  { $project: { _id: 0, foo: "$foo", bar: "$bar" } },
  { $group: { _id: {foo: "$foo", bar: "$bar" } } },
  { $project: { _id: 0, foo: "$_id.foo", bar: "$_id.bar" } }
]
```
I'm not sure this code will do that since it depends on finding `Bottom` and the `expr` being a `Document`.

Generally, I believe the group keys should just be the datasources themselves!

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +477,61 @@ impl<'a> Algebrizer<'a> {
...
-        Ok(stage)
+        project_stage.schema(&self.schema_inference_state())?;
+        Ok(if is_distinct {
+            let group_stage = Self::build_distinct_values_group_stage(
+                &expression,
+                project_stage,
+                self.scope_level,
+            );
+            group_stage.schema(&self.schema_inference_state())?;
+            group_stage
```
mattChiaravalloti, line 494: Aren't we supposed to follow-up the `$group` stage with a `$project` that removes the `_id` and projects the original datasources?

mattChiaravalloti (changes requested):
Some initial thoughts! I had already left some comments on the spec tests before realizing you only wanted me to focus on the algebrizer changes. Since they were already written, I left them in. I think this is definitely on the right track, but there are some places where we need to loosen conditions and handle more general cases. The testing is spot on! Happy to discuss more if needed.

tests/spec_tests/query_tests/distinct.yml
bucaojit: Oh, I had the collection originally named `distinct` but even that's unnecessary to have here, I'll update it.

mongosql/src/algebrizer/definitions.rs
bucaojit: It is, I removed it 👍

mongosql/src/algebrizer/definitions.rs
bucaojit: I updated this part a bit, I think it's now using the right key.

mongosql/src/algebrizer/definitions.rs
bucaojit: I added some code to take into account the bottom datasource.  Let me know if it's handled correctly.

mongosql/src/algebrizer/definitions.rs
bucaojit: This was super helpful.  I ran the query and it didn't work the way it was implemented.  I changed the logic to use the datasources for the group keys.  I also added a test that has a nested subquery.

mongosql/src/algebrizer/definitions.rs
bucaojit: I added the `$project` to get the original datasources.

mongosql/src/algebrizer/definitions.rs
bucaojit: I made the changes that should address this.  There are still some areas I'm not completely sure about such as adding the `BOTTOM_ALIAS`, and how to set is_nullable.  I'll look into it a bit more but might have to ask you for help tomorrow.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +478,105 @@ impl<'a> Algebrizer<'a> {
...
+        let mut project_expressions = BindingTuple::new();
+        for datasource_alias in datasource_aliases {
+            project_expressions.insert(
+                Key::named(datasource_alias.as_str(), scope_level),
+                Expression::FieldAccess(FieldAccess {
+                    expr: Box::new(Expression::Reference(ReferenceExpr {
+                        key: Key::bot(scope_level).clone(),
+                    })),
+                    field: datasource_alias.clone(),
+                    is_nullable: false,
```
bucaojit, line 553: Setting this to false at the moment, I'll continue working on setting this programmatically.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,30 +998,126 @@ impl<'a> Algebrizer<'a> {
...
+            .collect();
+
+        if let Some(bottom_key) = &bottom {
+            group_keys.push(OptionallyAliasedExpr::Aliased(AliasedExpr {
+                alias: BOTTOM_ALIAS.to_string(),
+                expr: Expression::Reference(ReferenceExpr {
+                    key: bottom_key.clone(),
+                }),
+            }));
+        }
```
mattChiaravalloti, line 1073:
```suggestion
        let mut datasource_counter = 0;
        let mut datasources = Vec::new();

        // Identify data sources from the result set schema
        for key in source_result_set.schema_env.keys() {
            if key.scope == self.scope_level {
                let group_key_name = format!("__groupKey{datasource_count}");
                datasources.push((group_key_name, key.clone());
                datasource_counter++;
            }
        }

        let mut group_keys: Vec<OptionallyAliasedExpr> = datasource_names
            .iter()
            .map(|(group_key_name, key)| {
                OptionallyAliasedExpr::Aliased(AliasedExpr {
                    alias: group_key_name.clone(),
                    expr: Expression::Reference(ReferenceExpr { key: key.clone() }),
                })
            })
            .collect();
```
Heads up: this is a big comment that includes 2 major points. First, it covers why `Bottom` should not be treated specially. Then, it talks about scope-level. Sorry it's one big thread. I was writing it as two threads but realized both of these changes touch the same spot so I combined.

### Bottom
I don't think we need to, or even should, treat `Bottom` specially in this implementation. The names we use for the group keys are totally up to us. While it's nice to use the datasource names themselves in most cases, it's a bit risky. Consider a query that has a user-provided datasource called `__bot`! We'd have a conflict since we're using a hardcoded `BOTTOM_ALIAS` here.

This could happen for a query like `SELECT DISTINCT {'a': a}, "__bot".* FROM "__bot"`; the results would look like `{'': {a: ...}, '__bot': { ... }}` where the empty key is actually representing the `Bottom` datasource and the `"__bot"` key represents the `Named("__bot")` datasource!

Given this example, that's why I proposed using a naming pattern like `__groupKeyX` and simply mapping that generated name to the datasource key from the `schema_env`.

### Scope level
We also need to be wary of `scope` here. Particularly because it is possible for the `schema_env` to include correlated datasources that are not actually outputted by the `source` stage. Here's an example I have in mind
```
SELECT (SELECT DISTINCT * FROM inner LIMIT 1) AS s FROM outer
```
(assumer `inner` has schema info indicating it has exactly 1 field, so the degree of the subquery is guaranteed to be 1.)

When we're algebrizing the `SubqueryExpr` in this query, we'll have the `(outer, 0)` datasource in our `schema_env`. In particular, when algebrizing the `SELECT DISTINCT *` clause in the subquery, the `source` will be the `FROM inner` clause. The `schema_env` being outputted from `FROM inner` will include not only `(inner, 1)` but also `(outer, 0)`. We _don't_ want to group by both of those datasources, we only want to group by the `inner` datasource since we're only looking for distinct results on the inner query.

Given this example, that's why I proposed using the `if key.scope == self.scope_level` conditional in the for loop to determine if a datasource from the env should be considered for grouping.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -904,30 +998,126 @@ impl<'a> Algebrizer<'a> {
...
+                },
+                Expression::FieldAccess(FieldAccess {
+                    expr: Box::new(Expression::Reference(ReferenceExpr {
+                        key: Key::bot(self.scope_level),
+                    })),
+                    field: BOTTOM_ALIAS.to_string(),
+                    is_nullable: false,
+                }),
+            );
+        }
```
mattChiaravalloti, line 1112:
```suggestion
        for (group_key_name, key) in &datasources {
            expression.insert(
                key.clone(), // given the way naming works, we need to use the same key!
                Expression::FieldAccess(FieldAccess {
                    expr: Box::new(Expression::Reference(ReferenceExpr {
                        key: Key::bot(self.scope_level),
                    })),
                    field: group_key_name.clone().to_string(),
                    is_nullable: false,
                }),
            );
        }
```
Consequent change based on the proposal above.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +478,105 @@ impl<'a> Algebrizer<'a> {
...
-            expression,
+            expression: expression.clone(),
             is_add_fields,
             cache: SchemaCache::new(),
         });
-        stage.schema(&self.schema_inference_state())?;
-        Ok(stage)
+        project_stage.schema(&self.schema_inference_state())?;
+        Ok(if is_distinct {
+            let group_stage = Self::build_distinct_values_group_and_project_stage(
```
mattChiaravalloti, line 489: [nit] this will be a `Project` stage with a `Group` source so the name `group_stage` is outdated. Maybe we could call this variable `distinct_stages` or something like that?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +478,105 @@ impl<'a> Algebrizer<'a> {
...
+        scope_level: u16,
+        datasources: BTreeSet<Key>,
+    ) -> mir::Stage {
+        let mut group_keys = Vec::new();
+        let mut bottom: Option<AliasedExpr> = None;
+        let mut datasource_aliases = Vec::new();
+
+        // Identify data sources and build group keys for deduplication
+        for key in datasources.iter() {
+            if key.scope == scope_level {
```
mattChiaravalloti, line 512: As opposed to the `*` case below, in this case we know that `datasources` is exactly the set of datasource bindings being output from the `source` so we actually can safely ignore this `scope_level` check!

mongosql/src/algebrizer/definitions.rs
```diff
@@ -475,14 +478,105 @@ impl<'a> Algebrizer<'a> {
...
+                            alias: name.clone(),
+                            expr: Expression::Reference(ReferenceExpr { key: key.clone() }),
+                        }));
+                        datasource_aliases.push(name.clone());
+                    }
+                    DatasourceName::Bottom => {
+                        bottom = Some(AliasedExpr {
+                            alias: BOTTOM_ALIAS.to_string(),
+                            expr: Expression::Reference(ReferenceExpr { key: key.clone() }),
+                        });
```
mattChiaravalloti, line 525: Similar to what I said for the `*` case, we should (1) not special case `Bottom` and (2) use some other pattern for creating group key names since we're free to use whatever we want as long as we avoid naming collisions -- again, using datasource names _could_ result in a collision with `Bottom` in rare circumstances.

mongosql/src/algebrizer/definitions.rs
mattChiaravalloti: I suppose the only way this could be nullable is if the result set coming into the `SELECT DISTINCT` clause is empty. If it is, inevitably this field will be "missing". However, strictly speaking, the result set would therefore be empty so what does the result set schema even represent at that point? I guess we can't statically know it is empty... I think setting this `true` might be the safest bet since we may never be able to statically determine the appropriate value 🤔 Let's spend more time thinking about this for sure.

mattChiaravalloti (changes requested):
This is shaping up excellently. Thank you for sorting through all of this, it's clearly much more complicated than we initially thought during design time 😓 I think you nailed the essence of what the implementation needs to be. I detailed 2 correctness issues that ultimately reduce the size of the implementation. See the comments below. The big comment is on the `SELECT *` case but it applies to `SELECT VALUE(S)` as well! Let me know if you want to discuss tomorrow.

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,154 @@
...
+      - { 'foo': { "_id": 2, "a": 1, "b": 2 }, 'bar': { "_id": 2, "a": 2, "b": 2 } }
+      - { 'foo': { "_id": 3, "a": 2, "b": 2 }, 'bar': { "_id": 3, "a": 2, "b": 2 } }
+      - { 'foo': { "_id": 4, "a": 1, "b": 2 }, 'bar': { "_id": 4, "a": 2, "b": 1 } }
+
+  - description: "SELECT DISTINCT specific columns from joined tables"
+    query: "SELECT DISTINCT foo.a, bar.b FROM db.foo, db.bar WHERE foo._id = bar._id"
+    result:
+      - { '': { "a": 1, "b": 1 } }
+      - { '': { "a": 2, "b": 2 } }
+      - { '': { "a": 2, "b": 4 } }
```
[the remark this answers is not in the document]
bucaojit: That's a good catch.  I missed that I had switched the order of the keys as an additional test and saw `      - { _id: 4, b: 2, a: 1 }` as b:1.  But it looks like uncovered an issue how the test runner checks the expected vs actual results since it marked it as correct.  I have a fix for the issue, I'll open a separate ticket for that. 
I corrected the results here and another one where I had b:1 in `SELECT DISTINCT * on joined tables unique rows for entire document`

mongosql/src/algebrizer/definitions.rs
bucaojit: Thank you so much for that!  I was having trouble trying to figure out the groupkey name was the way to go and how to handle bottom.  This explains it very clearly how to handle them.  And the scope explanation was helpful as well, I added the check here as proposed.

mongosql/src/algebrizer/definitions.rs
bucaojit: That works, I renamed it to `distinct_stages`.

mongosql/src/algebrizer/definitions.rs
bucaojit: Okay, I added a similar logic here as with the `*` case.

mongosql/src/algebrizer/definitions.rs
bucaojit: Thanks for the explanation.  I'll set it to true to take the safer route.

tests/spec_tests/query_tests/distinct.yml
```diff
@@ -0,0 +1,154 @@
...
+      - { _id: 2, a: 2, b: 2 }
+      - { _id: 3, a: 2, b: 2 }
+      - { _id: 4, a: 2, b: 4 }
+
+    documents:
+      - { _id: 1, doc: { a: 1, b: 2 } }
+      - { _id: 2, doc: { a: 1, b: 2 } }
+      - { _id: 3, doc: { a: 2, b: 2 } }
+      - { _id: 4, doc: { a: 2, b: 3 } }
+      - { _id: 5, doc: { b: 2, a: 1 } }
```
jchemburkar, line 22: I think the behavior should definitely work as expected, but would it be worth adding a doc (and an array below) that are subsets of another? eg `- {_id: 6, doc: {a: 1, b:2, c: 3 } }`? To verify we dont get caught by only checking the overlapping subset of keys.

mongosql/src/algebrizer/test.rs
```diff
@@ -5667,20 +5667,97 @@ mod select_clause {
     }
 
     test_algebrize!(
-        select_distinct_not_allowed,
+        select_values_distinct,
```
jchemburkar, line 5670: apologies if I'm missing it but I think a test with the `NonStarStandardSelectBody` Error could be useful as well, here, in the spec tests, or both. Codifying that we dont allow stuff like `select distinct (foo.*, bar.a)` or (I think) even `select distinct foo.*`

jchemburkar (approved):
The code looks really good to me, I think Matt's last batch of comments addressed the only things that seemed to jump out to me. Also great work with commenting / incorporating some of this context that I certainly didn't understand coming in.

Spent an extra chunk of time thinking about tests when re-reviewing, had a couple of ideas, non blocking. Nice work!

tests/spec_tests/query_tests/distinct.yml
bucaojit: Sure, I added the doc and array with the updated results.

mongosql/src/algebrizer/test.rs
bucaojit: I added an algebrizer e2e test to check that `SELECT DISTINCT *, a` gets that error.  A query like `select distinct foo.*` will go down the `ast::SelectBody::Values(exprs)` branch and works ok.

PR #24, opened 2025-02-13
SQL-2647: Fix ResultSet Schema for UNION ALL

pmeredit:
This came up while debugging things with Alexi. UNION ALL in many cases was causing our odbc/jdbc drivers to crash because it was generating result set schema the drivers assumed to be impossible!

### DIFF
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -2062,14 +2062,14 @@ impl<'a> Algebrizer<'a> {
         // If we find exactly one datasource that May or Must contain
         // the field `i`, we return `datasource.i`. If there is more
         // than one, it is an ambiguous error.
-        let mut i_containing_datasources = dbg!(&self.schema_env)
+        let mut i_containing_datasources = self
+            .schema_env
             .iter()
             .filter(|(_, schema)| {
                 let sat = schema.contains_field(i.as_ref());
                 sat == Satisfaction::May || sat == Satisfaction::Must
             })
             .collect::<Vec<_>>();
-        dbg!(&i_containing_datasources);
         // If there is no datasource containing the field, the field is not found.
         if i_containing_datasources.is_empty() {
             let all_keys = self
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -2062,14 +2062,14 @@ impl<'a> Algebrizer<'a> {
         // If we find exactly one datasource that May or Must contain
         // the field `i`, we return `datasource.i`. If there is more
         // than one, it is an ambiguous error.
-        let mut i_containing_datasources = self
-            .schema_env
+        let mut i_containing_datasources = dbg!(&self.schema_env)
             .iter()
             .filter(|(_, schema)| {
                 let sat = schema.contains_field(i.as_ref());
                 sat == Satisfaction::May || sat == Satisfaction::Must
             })
             .collect::<Vec<_>>();
+        dbg!(&i_containing_datasources);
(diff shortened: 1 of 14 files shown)


pmeredit:
Before:
```
cargo run --bin mongosql-cli -- "SELECT x FROM foo UNION ALL SELECT _id FROM foo" --translation
   Compiling mongosql v0.0.0 (/home/pmeredit/git/mongosql/mongosql)
   Compiling mongosql-cli v0.1.0 (/home/pmeredit/git/mongosql/mongosql-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.07s
     Running `target/debug/mongosql-cli 'SELECT x FROM foo UNION ALL SELECT _id FROM foo' --translation`
target_db: test,
target_collection: Some("foo"),
result set schema:
{
  "bsonType": "object",
  "properties": {
    "": {
      "anyOf": [
        {
          "bsonType": "object",
          "properties": {
            "_id": {
              "bsonType": "objectId"
            }
          },
          "required": [
            "_id"
          ],
          "additionalProperties": false
        },
        {
          "bsonType": "object",
          "properties": {
            "x": {
              "anyOf": [
                {
                  "bsonType": "minKey"
                },
                {
                  "bsonType": "int"
                }
              ]
            }
          },
          "additionalProperties": false
        }
      ]
    }
  },
  "required": [
    ""
  ],
  "additionalProperties": false
}
```


Now:
```cargo run --bin mongosql-cli -- "SELECT x FROM foo UNION ALL SELECT _id FROM foo" --translation
   Compiling mongosql v0.0.0 (/home/pmeredit/git/mongosql/mongosql)
   Compiling mongosql-cli v0.1.0 (/home/pmeredit/git/mongosql/mongosql-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.74s
     Running `target/debug/mongosql-cli 'SELECT x FROM foo UNION ALL SELECT _id FROM foo' --translation`
target_db: test,
target_collection: Some("foo"),
result set schema:
{
  "bsonType": "object",
  "properties": {
    "": {
      "bsonType": "object",
      "properties": {
        "_id": {
          "bsonType": "objectId"
        },
        "x": {
          "anyOf": [
            {
              "bsonType": "minKey"
            },
            {
              "bsonType": "int"
            }
          ]
        }
      },
      "required": [
        "_id"
      ],
      "additionalProperties": true
    }
  },
  "required": [
    ""
  ],
  "additionalProperties": false
}```

pmeredit:
I really have no clue why we now need to manually simplify the schemas for Project and Group. Those should be the only ones since they are the only ones that create new mappings in a way using unions (versus just grabbing them from previous stages or creating a map directly like Collection)

pmeredit:
> I really have no clue why we now need to manually simplify the schemas for Project and Group. Those should be the only ones since they are the only ones that create new mappings in a way using unions (versus just grabbing them from previous stages or creating a map directly like Collection)

This might actually obviate the need to simplify the result set schema now. But we're probably moving to schema generation anyway?

bucaojit (approved):
LGTM

pmeredit:
@mattChiaravalloti It was never actually correct to use MISSING in the type constraint tests because they generate Schema which means that that field is _always missing_, which should correctly return "field not found". This was working because because our Schema were not being simplified. Now that they are, this was causing errors!

MISSING is only viable when a field is not _always_ missing.

mattChiaravalloti (approved):
Thanks for explaining the removal of `missing` from the type constraint tests. That totally makes sense!

The change to `simplify` feels like it should have been done a long time ago so I'm glad to see we got this done 👍

PR #25, opened 2025-02-18
SQL-2646: Implement Limit Movement

terakilobyte:
This PR allows limit to move the same as offset. It also refactors in a few ways:
* Both limit and offset are referred to as "pagination" and methods are updated accordingly.
* Offset and Limit and now pagination invalidating. This is to prevent moving a limit ahead of an offset.
* I attempted to make the comments a bit more clear and pronounced.
* Tests that had a Limit stage have had that stage removed, and a few tests have been added to show that Limit now moves. This is an attempt to show the behavior in more isolation.

### DIFF
--- a/mongosql/src/mir/optimizer/stage_movement/mod.rs
+++ b/mongosql/src/mir/optimizer/stage_movement/mod.rs
@@ -36,9 +36,10 @@ use crate::{
 use std::collections::HashSet;
 
 impl Stage {
-    // This is used for moving Offset as high as possible. They can be moved ahead of Any
-    // Stage that is not defined as offset invalidating.
-    fn is_offset_invalidating(&self) -> bool {
+    /// `is_pagination_invalidating` is used for moving Pagination stages - Offset
+    /// and Limit - as high as possible. They can be moved ahead of any Stage that
+    /// is not defined as pagination invalidating.
+    fn is_pagination_invalidating(&self) -> bool {
         match self {
             // A tautological Filter will not invalidate offset, but we don't have a SAT solver.
             Stage::Filter(_) => true,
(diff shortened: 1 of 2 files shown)


mongosql/src/mir/optimizer/stage_movement/mod.rs
```diff
@@ -36,9 +36,10 @@ use crate::{
 use std::collections::HashSet;
 
 impl Stage {
```
pmeredit, line 38: [nit]: Ugh, I hate to do this, but could we call this bounds instead of pagination? This isn't really correct use of the word pagination.

mongosql/src/mir/optimizer/stage_movement/test.rs
```diff
@@ -140,6 +140,119 @@ macro_rules! test_move_stage_no_op {
     };
 }
 
+test_move_stage!(
```
mattChiaravalloti, line 143: Let's use the `test_move_stage_no_op!` macro for any tests that expect the `expected` to be the same as the `input`. This not only reduces the size of the test file but also -- in addition to the test name -- helps make it clear that the test is asserting movement _does not_ happen.

pmeredit (approved):
Looks great, other than the naming :joy: pagination is the process of splitting text into pages. It would work better with window functions (when we have those) as a name. I just think bounds is a better name.

mongosql/src/mir/optimizer/stage_movement/test.rs
mattChiaravalloti: Ok I see this is the only test in this PR that would require this change! Feel free to just change this one -- no need to audit the rest of the test file for `expected_changed = false`. We can do that another time.

mongosql/src/mir/optimizer/stage_movement/mod.rs
terakilobyte: Sure. I chose pagination since skip and limit are often used for that.

mongosql/src/mir/optimizer/stage_movement/mod.rs
```diff
@@ -36,19 +36,24 @@ use crate::{
...
             // It's possible to have a Group that does not modify cardinality and thus invalidate
             // an offset, but we can consider that a very rare occurrence.
             Stage::Group(_) => true,
+            // ordering of two Limits does not matter. Limit 5, Limit 3 = Limit 3, Limit 5.
+            // MongoDB coalesces two Limits into one
             Stage::Limit(_) => true,
             // ordering of two Offsets does not matter. Offset 5, Offset 3 = Offset 3, Offset 5.
-            Stage::Offset(_) => false,
+            // However, Limit should not be moved above an Offset.
+            // MongoDB coalesces two Offsets($skip) into one
```
mattChiaravalloti, line 55: These comments are incomplete.

Not only do we not want to move a `Limit` above an `Offset`, but also we don't want to move an `Offset` above a `Limit`. For example, for `SELECT * FROM (SELECT a FROM foo LIMIT 10) OFFSET 5` we could push the `OFFSET 5` into the derived query but it actually still come after the `LIMIT 10` or else we'll change the result set.

Can we update these comments to point out both cases?

mattChiaravalloti (changes requested):
Looking good 👍  Minor changes suggested.

mongosql/src/mir/optimizer/stage_movement/mod.rs
terakilobyte: I've updated the function comment since it seemed appropriate to do so there!

mattChiaravalloti (approved):
lgtm 🚢

PR #26, opened 2025-02-18
SQL-2651: Optimize WHERE/ON false queries such that they will always use index scans

pmeredit:
This PR ended up doing two things:
* Change {$match: {$expr: false}} to {$match: {_id: MinKey, $expr: false}}
* Change Join algebrization to always use a separate Filter instead of the Condition
  * This is fine because Stage Movement will still move them into the condition, if it makes sense, but will move it above the join now when it can! (if the condition constant)

### DIFF
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -613,6 +613,9 @@ impl<'a> Algebrizer<'a> {
                 cache: SchemaCache::new(),
             }),
         };
+        // We always add the condition as a separate filter to facilitate optimization of constant
+        // join conditions. The stage movement optimization will move the filter back into the join
+        // condition when necessary.
         let stage = if let Some(condition) = condition {
             mir::Stage::Filter(mir::Filter {
                 source: Box::new(stage),
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -589,7 +589,7 @@ impl<'a> Algebrizer<'a> {
                     join_type: mir::JoinType::Left,
                     left: Box::new(left_src),
                     right: Box::new(right_src),
-                    condition,
+                    condition: None,
                     cache: SchemaCache::new(),
                 })
             }
@@ -601,7 +601,7 @@ impl<'a> Algebrizer<'a> {
                     join_type: mir::JoinType::Left,
                     left: Box::new(right_src),
                     right: Box::new(left_src),
-                    condition,
+                    condition: None,
                     cache: SchemaCache::new(),
                 })
             }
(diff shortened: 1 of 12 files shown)


mongosql/src/algebrizer/definitions.rs
```diff
@@ -605,13 +605,24 @@ impl<'a> Algebrizer<'a> {
...
-            }),
+            ast::JoinType::Cross | ast::JoinType::Inner => {
+                let join = mir::Stage::Join(mir::Join {
+                    join_type: mir::JoinType::Inner,
+                    left: Box::new(left_src),
+                    right: Box::new(right_src),
+                    condition: None,
+                    cache: SchemaCache::new(),
+                });
+                if let Some(condition) = condition {
```
mattChiaravalloti, line 616:
```suggestion
                // The stage_movement optimization will place this condition in the Join if it makes sense. Otherwise,
                // it will move it as early in the pipeline as possible.
                if let Some(condition) = condition {
```
Not sure if I captured this idea exactly correctly, but I think we'd benefit from commenting this to explain why we're doing it.

(This is an amazing idea btw 🔥 )

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
mattChiaravalloti, line 1: Can you update the file-level comment here to note that we also handle folding into `MatchQuery::False` in this optimization?

mongosql/src/codegen/match_query.rs
```diff
@@ -29,6 +29,7 @@ impl MqlCodeGenerator {
             Regex(r) => self.codegen_match_regex(r),
             ElemMatch(em) => self.codegen_match_elem_match(em),
             Comparison(c) => self.codegen_match_comparison(c),
+            False => Ok(bson!({"_id": Bson::MinKey, "$expr": false})),
```
mattChiaravalloti, line 32:
```suggestion
            // Some versions of MongoDB do not properly optimize {$match: { $expr: false } } so we codegen it
            // this way to ensure efficient performance. Otherwise, such a query could result in a needless collection
            // scan.
            False => Ok(bson!({"_id": Bson::MinKey, "$expr": false})),
```
We'd also benefit from a comment here.

mattChiaravalloti (approved):
Looks good 👍 I suggest adding a few more code comments to document why this all exists.

mongosql/src/algebrizer/definitions.rs
pmeredit: looks good to me! (and thanks)

pmeredit:
Ah good, applying suggestions is properly verified. Was a bit worried.

terakilobyte (approved):
LGTM

PR #27, opened 2025-02-19
SQL-2362: implement schema derivation for array ops in $match

jchemburkar:


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -440,6 +440,37 @@ impl DeriveSchema for TaggedOperator {
                     })
             }};
         }
+        macro_rules! array_schema_or_error {
```
jchemburkar, line 443: this is used for `$reduce` and `$map`, which tolerate nullish arguments. This allows for null, missing, or valid anyofs to be included.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -773,22 +804,23 @@ impl DeriveSchema for TaggedOperator {
...
-                let array_schema = match input_schema {
-                    Schema::Array(a) => *a,
-                    _ => {
-                        return Err(Error::InvalidExpressionForField(
-                            format!("{:?}", m.input),
-                            "input",
-                        ))
-                    }
-                };
+                if input_schema.satisfies(&NULLISH.clone()) == Satisfaction::Must {
```
jchemburkar, line 807: as mentioned in the macro above, these array operators in many cases can handle an array _or_ null, but we were treating them strictly as arrays, so these changes simply enable that behavior

terakilobyte (approved):
LGTM. I'd just like to see more rustdoc style comments as we add/refactor code.

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -2771,3 +2771,803 @@ mod string_ops {
...
+        })),
+        input = r#"{"$match": {"$expr": {"$not": {"$last": "$foo"}}}}"#,
+        ref_schema = Schema::Any
+    );
+    test_derive_schema_for_match_stage!(
+        first_null,
+        expected = Ok(Schema::Document(Document {
+            keys: map! {
+                "foo".to_string() => Schema::AnyOf(set!(
+                    Schema::Atomic(Atomic::Null),
```
pmeredit, line 2814: This really has to be NULL, right? I think this is fine though

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -2771,3 +2771,803 @@ mod string_ops {
...
+            keys: map! {
+                "foo".to_string() => Schema::AnyOf(set!(
+                    Schema::Array(Box::new(Schema::Document(Document::any()))),
+                    Schema::Document(Document::any())
+                ))
+            },
+            required: set!("foo".to_string()),
+            ..Default::default()
+        })),
+        input = r#"{"$match": {"$expr": {"$mergeObjects": "$foo"}}}"#,
```
pmeredit, line 3324: I am wholly surprised that this works with $foo being an array. I think most operators don't work that way. I love agg :joy:

pmeredit (approved):
LGTM!

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
jchemburkar: It is really bewildering. It works for an array or document, hence why neither can be assumed. Maybe it is usable in another operator that makes this make snese (like a windowing type deal)
```
derivation> db.aggregate([ { $documents: [ { _id: 1, "foo": { "a": 1 } }] }, { $project: { expr: { $mergeObjects: "$foo" } } }] )
[ { _id: 1, expr: { a: 1 } } ]
derivation> db.aggregate([ { $documents: [ { _id: 1, "foo": [{ "a": 1 }] }] }, { $project: { expr: { $mergeObjects: "$foo" } } }] )
[ { _id: 1, expr: { a: 1 } } ]
```

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -1456,6 +1457,436 @@ impl MatchConstrainSchema for Expression {
             Ok(())
         }
 
+        fn match_derive_first_last(u: &UntaggedOperator, state: &mut ResultSetState) -> Result<()> {
```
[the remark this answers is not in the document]
jchemburkar: Yeah good idea. I can do this now for these operators and we can go back through and comment as part of a cleanup ticket also

PR #28, opened 2025-02-19
SQL-2622: Implement multi-column COUNT

mattChiaravalloti:
This PR adds support for multi-column `COUNT`. It supports this for distinct and non-distinct `COUNT`. It also updates the desugarer to properly handle single-column `COUNT` when that column `MAY`/`MUST` be a document, and to avoid conditionally checking literal values. It also added some new exprs and mql functions to the `air` since we needed them for our desugaring (`$map`, `$objectToArray`, `$allElementsTrue`, etc. you'll see them when you review).

This ended up being a bit larger than originally anticipated mainly because, in addition to implementing multi-column `COUNT`, I covered the other things described above. These changes touch a lot of files but the core of the new work is still in the `accumulators.rs` desugarer, the query spec tests, and a new rewriter pass.


mongosql/src/air/desugarer/accumulators.rs
```diff
@@ -115,44 +162,72 @@ impl AccumulatorsDesugarerVisitor {
...
-                // In the case the condition evaluates to false, we count this argument.
                 Literal(LiteralValue::Integer(1)),
             )
         };
 
+        let arg = match arg {
+            // No need to create a conditional if the argument is a literal value.
+            // We know null will result in the then case and non-null will result in the else.
+            Literal(LiteralValue::Null) => then,
+            Literal(_) => r#else,
```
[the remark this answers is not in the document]
mattChiaravalloti: There is also the obvious future extension here where we optimize at the mir-level whether or not we need to use null-checking semantics at all but that's for another time.

mongosql/src/air/desugarer/testdata/desugar_accumulators.yml
```diff
@@ -59,6 +59,111 @@ tests:
...
+          {
+            "_id": null,
+            "acc":
+              {
+                "$sum":
+                  {
+                    "$cond":
+                      [
+                        {
+                          "$or":
```
mattChiaravalloti, line 83: I flipped this from a conjunction of negative conditions to a disjunction of positive ones. This just helped line up the single- and multiple-column cases more easily, especially now that we account for single-columns that may be documents.

mongosql/src/algebrizer/test.rs
```diff
@@ -5177,6 +5177,7 @@ mod aggregation {
                 function: mir::AggregationFunction::Count,
                 distinct: false,
                 arg: mir::Expression::Literal(mir::LiteralValue::Integer(42)).into(),
+                arg_is_possibly_doc: Satisfaction::Not,
```
mattChiaravalloti, line 5180: This value is irrelevant for testing for the majority of the codebase. As noted where the field is defined, it is only relevant at desugar time. That's why you keep seeing it set to `Not` in most unit tests.

mongosql/src/ast/rewrites/aggregate.rs
```diff
@@ -165,6 +174,59 @@ impl Visitor for AggregateUsageCheckVisitor {
     }
 }
 
+#[derive(Default)]
+pub struct MultiArgCountVisitor {
```
mattChiaravalloti, line 178: This was not in the design doc. The `algebrizer` rejects aggregations with multiple arguments, and the `mir` and `air` only support single-argument accumulator exprs. I figured the smallest change was to syntactically rewrite multi-column `COUNT`s into `COUNT(<doc>)`, thus making it an aggregation with a single argument. The previous ticket already enabled `COUNT` with document arguments so this change requires no further work down the line except desugarer updates.

mongosql/src/ast/rewrites/aggregate.rs
```diff
@@ -165,6 +174,59 @@ impl Visitor for AggregateUsageCheckVisitor {
...
+                ast::Expression::Function(ast::FunctionExpr {
+                    function: ast::FunctionName::Count,
+                    args: ast::FunctionArguments::Args(args),
+                    set_quantifier,
+                }) if args.len() > 1 => {
+                    let doc: Vec<ast::DocumentPair> = args
+                        .into_iter()
+                        .map(|arg| {
+                            let ast::Expression::Identifier(key) = arg.clone() else {
+                                self.error = Some(Error::InvalidMultiArgCountArg);
```
mattChiaravalloti, line 197: I decided for now we will only support field references as the multiple arguments. It is doubtful anyone will want to do something like `COUNT(a+1, WHEN b CASE 7 THEN "yes" ELSE c END, x)`. If someone requests this, we can support it then. For now, this is a very convenient restriction.

tests/spec_tests/query_tests/group_by.yml
```diff
@@ -9,7 +9,7 @@ catalog_data:
 
     multi:
       - { "_id": 1, "a": 1, "b": 2, "c": 1 }
-      - { "_id": 2, "a": 2, "b": 2, "c": 2 }
+      - { "_id": 2, "a": 2, "b": 2, "c": 3 }
```
mattChiaravalloti, line 12: Made this change for the sake of `DISTINCT` testing.

pmeredit (approved):
LGTM! I really like the syntax rewrite here, and now I'm pretty convinced that is how we should handle the new Unwind support, also.

Although, we don't have schema info so we might end up with extra unwinds if we go that way.

mongosql/src/air/desugarer/accumulators.rs
```diff
@@ -4,33 +4,75 @@ use crate::{
...
 use linked_hash_map::LinkedHashMap;
+use mongosql_datastructures::unique_linked_hash_map::UniqueLinkedHashMap;
 
-macro_rules! make_single_expr_count_conditional {
-    ($arg:expr, $then:expr, $else:expr) => {
-        Box::new(make_cond_expr!(
-            MQLSemanticOperator(MQLSemanticOperator {
-                op: MQLOperator::In,
+// Condition used to desugar a $sqlCount when the argument is not a document. This condition asserts
+// that the arg is null or missing.
```
bucaojit, line 18: Nice use of macros, that helped with keeping the code easy to understand 👍

mongosql/src/ast/rewrites/aggregate.rs
bucaojit: Yeah that's a good call, no need to add too much complexity if it's not requested.

bucaojit (changes requested):
The changes look great and the tests are thorough. 
I had one question about supporting `SELECT COUNT(DISTINCT col,...)`.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1106,15 +1107,18 @@ impl<'a> Algebrizer<'a> {
                 return Err(Error::StarInNonCount);
             }
             ast::FunctionArguments::Args(ve) => {
+                let arg = if ve.len() != 1 {
```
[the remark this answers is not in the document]
mattChiaravalloti: Oh that's surprising! I wonder if it's a rewriter order thing 🤔 Like we do the accumulator rewrite before the select clause one. This spec test obviously passes:
```
- description: COUNT distinct multi column correctness test -- only counts rows with at least one non-nullish value
    query: "SELECT * FROM foo.multi AS m GROUP BY a AS a AGGREGATE COUNT(DISTINCT b, c) AS gcount"
```
so I'll need to figure out what's causing the `COUNT(<multi>)` in the select position to fail! Great catch, thank you!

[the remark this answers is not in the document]
mattChiaravalloti: Fixed!

bucaojit (approved):
LGTM! Great work and nice fix.

PR #29, opened 2025-02-19
Chore: remove oidc since it is not used by mongosql

pmeredit:


terakilobyte (approved):
LGTM. If we ever need oidc for the cli we can add it back in!

PR #30, opened 2025-02-20
SQL-2625: Add/Update spec tests for BETWEEN, CASE, and NULLIF

mattChiaravalloti:
This PR updates the spec tests for the `BETWEEN` and simple `CASE` operators and the `NULLIF` scalar function. As [noted in the design doc](https://docs.google.com/document/d/1htydMszlKy_nym2_gQsZvgkBuJQ_6qyt175m20Bc_fQ/edit?tab=t.0#bookmark=id.e625i71i29nn), since `CASE` and `NULLIF` perform implicit equality comparisons, we do not exhaustively test all types in the query tests for them. Instead, we just have type constraint tests for those 2. This PR updates the type constraint tests for those to account for document and array operands. For `BETWEEN`, we do include testing for common types, particularly those that can be written by users using literal syntax, so this PR adds `BETWEEN` query spec testing for documents and arrays.


tests/spec_tests/query_tests/operators_comparison.yml
```diff
@@ -1052,3 +1148,34 @@ tests:
       - {'': { "a": { "$date": { "$numberLong": "10" } }, "c": { "$date": { "$numberLong": "15" } }, "r": null } }
       - {'': { "a": { "$date": { "$numberLong": "10" } }, "b": { "$date": { "$numberLong": "5" } }, "r": null } }
 
+  - description: correctness test for BETWEEN for documents
```
jchemburkar, line 1151: not the point of this ticket at all but in looking at this file i noticed (and correct me if there is a reason this should be the case) it looks like we have no "greater than upper bound" test for between strings, ie, the standard _id:  4 document in these collections

jchemburkar (approved):
Your changes look good! I noticed one unrelated thing, can file a ticket if it makes sense

tests/spec_tests/query_tests/operators_comparison.yml
mattChiaravalloti: Good catch. I can drive-by add that here.

bucaojit (approved):
Looks good

PR #31, opened 2025-02-20
SQL-2365: Add schema derivation for Lookup, GraphLookup, and UnionWith

pmeredit:
This also includes an implementation of include-only $project derivation for testing purposes.

### DIFF
--- a/agg-ast/schema_derivation/src/schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/schema_derivation.rs
@@ -106,6 +106,7 @@ impl DeriveSchema for Stage {
                         &mut required_doc,
                         field_schema.clone(),
                         path.clone(),
+                        false,
                     );
                 }
             });
@@ -182,15 +183,15 @@ impl DeriveSchema for Stage {
                 .ok_or_else(|| Error::UnknownReference(from_field.to_string()))?
                 .clone();
             if let Some(ref depth_field) = graph_lookup.depth_field {
-                let depth_field_schema = Schema::Atomic(Atomic::Long);
                 insert_required_key_into_document(
                     &mut from_schema,
-                    depth_field_schema,
+                    Schema::Atomic(Atomic::Long),
                     depth_field
                         .as_str()
                         .split('.')
                         .map(|s| s.to_string())
                         .collect(),
+                    true,
                 );
             }
             insert_required_key_into_document(
@@ -202,6 +203,7 @@ impl DeriveSchema for Stage {
                     .split('.')
                     .map(|s| s.to_string())
                     .collect(),
+                true,
             );
             Ok(state.result_set_schema.to_owned())
         }
@@ -292,6 +294,7 @@ impl DeriveSchema for Stage {
                     .split('.')
                     .map(|s| s.to_string())
                     .collect(),
+                true,
             );
             Ok(state.result_set_schema.to_owned())
         }
@@ -357,6 +360,7 @@ impl DeriveSchema for Stage {
                 &mut state.result_set_schema,
                 Schema::Array(Box::new(lookup_schema.clone())),
                 as_var.split('.').map(|s| s.to_string()).collect(),
+                true,
             );
             Ok(state.result_set_schema.to_owned())
         }
@@ -461,12 +465,14 @@ impl DeriveSchema for Stage {
                                         Schema::Atomic(Atomic::Null)
                                     )),
                                     path,
+                                    true,
                                 );
                             } else {
                                 insert_required_key_into_document(
                                     &mut state.result_set_schema,
                                     Schema::Atomic(Atomic::Integer),
                                     path,
+                                    true,
                                 );
                             }
                         }
--- a/agg-ast/schema_derivation/src/schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/schema_derivation.rs
@@ -342,14 +342,15 @@ impl DeriveSchema for Stage {
                         .catalog
                         .get(&from_name)
                         .ok_or_else(|| Error::UnknownReference(from_name))?;
-                    let state = &mut ResultSetState {
+                    let pipeline_state = &mut ResultSetState {
                         catalog: state.catalog,
                         variables: state.variables.clone(),
                         result_set_schema: from_schema.clone(),
                         current_db: state.current_db.clone(),
                         null_behavior: state.null_behavior,
                     };
-                    let pipeline_schema = derive_schema_for_pipeline(p.pipeline.clone(), state)?;
+                    let pipeline_schema =
+                        derive_schema_for_pipeline(p.pipeline.clone(), pipeline_state)?;
                     state.result_set_schema = state.result_set_schema.union(&pipeline_schema);
                     Ok(state.result_set_schema.to_owned())
                 }
--- a/agg-ast/schema_derivation/src/schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/schema_derivation.rs
@@ -3,8 +3,9 @@ use crate::{
     schema_difference, schema_for_type_numeric, schema_for_type_str, Error, Result,
 };
 use agg_ast::definitions::{
-    Densify, Expression, LiteralValue, Ref, Stage, TaggedOperator, UntaggedOperator,
-    UntaggedOperatorName, Unwind,
+    ConciseSubqueryLookup, Densify, EqualityLookup, Expression, GraphLookup, LiteralValue, Lookup,
+    LookupFrom, ProjectItem, ProjectStage, Ref, Stage, SubqueryLookup, TaggedOperator, UnionWith,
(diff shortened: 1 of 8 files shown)


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +151,186 @@ impl DeriveSchema for Stage {
...
+                keys.insert(
+                    "_id".to_string(),
+                    state
+                        .result_set_schema
+                        .get_key("_id")
+                        .cloned()
+                        .unwrap_or(Schema::Missing),
+                );
+            }
+            Ok(Schema::Document(Document {
```
pmeredit, line 230: note, if _id is Missing simplify will remove it from required. I change the test code to always call simplify, and when we linked this up to the schema manager we need to make sure we simplify

pmeredit: I guess I could just call simplify here

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +169,184 @@ impl DeriveSchema for Stage {
...
+                insert_required_key_into_document(
+                    &mut from_schema,
+                    depth_field_schema,
+                    vec![depth_field.clone()],
+                );
+            }
+            insert_required_key_into_document(
+                &mut state.result_set_schema,
+                Schema::Array(Box::new(from_schema.clone())),
+                vec![graph_lookup.r#as.to_string()],
```
jchemburkar, line 195: should we just make this _as like in other stages?

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +169,184 @@ impl DeriveSchema for Stage {
...
+                .catalog
+                .get(&from_field)
+                .ok_or_else(|| Error::UnknownReference(from_field.to_string()))?
+                .clone();
+            if let Some(ref depth_field) = graph_lookup.depth_field {
+                let depth_field_schema = Schema::Atomic(Atomic::Long);
+                insert_required_key_into_document(
+                    &mut from_schema,
+                    depth_field_schema,
+                    vec![depth_field.clone()],
```
jchemburkar, line 189:
```suggestion
                    depth_field.as_str().split('.').map(|s| s.to_string()).collect(),
```
This might not be perfect with rustfmt/references but I think this is the right idea. I tested it and it looks like if you have a document field in the from collection and the depth_field is nested, it will treat the field name as a path rather than a literal. It is ugly but here is a demonstration from my local mongo:
```
derivation> db.aggregate([
...   {$documents: [{_id: 1, "foo": 1,},]},
...   {"$graphLookup": {"from": "bar", "startWith": "$foo", "connectFromField": "foo", "connectToField": "baz", "depthField": "DEPTH", "as": "arr"}}
... ])
[
  {
    _id: 1,
    foo: 1,
    arr: [ { _id: 0, baz: 1, doc: { a: 0 }, DEPTH: Long('0') } ]
  }
]
derivation> db.aggregate([
...   {$documents: [{_id: 1, "foo": 1,},]},
...   {"$graphLookup": {"from": "bar", "startWith": "$foo", "connectFromField": "foo", "connectToField": "baz", "depthField": "doc.DEPTH", "as": "arr"}}
... ])
[
  {
    _id: 1,
    foo: 1,
    arr: [ { _id: 0, baz: 1, doc: { a: 0, DEPTH: Long('0') } } ]
  }
]
```

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +169,184 @@ impl DeriveSchema for Stage {
...
+                .flatten()
+                .map(|(k, v)| Ok((k.clone(), v.derive_schema(state)?)))
+                .collect::<Result<BTreeMap<String, Schema>>>()?;
+            let from_ns = from_to_ns(from.as_ref().ok_or(Error::MissingFromField)?, state);
+            let from_schema = state
+                .catalog
+                .get(&from_ns)
+                .ok_or_else(|| Error::UnknownReference(from_ns.into()))?;
+            variables.extend(state.variables.clone());
+            let mut lookup_state = ResultSetState {
```
jchemburkar, line 334: really cool stuff here, creating a new result set state. I imagine this is one of the only times well do this as opposed to cloning the original

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: I assume this will also hold for the `_as` fields in normal $lookups

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +169,184 @@ impl DeriveSchema for Stage {
...
+                catalog: state.catalog,
+                variables,
+                result_set_schema: from_schema.clone(),
+                current_db: state.current_db.clone(),
+                null_behavior: state.null_behavior,
+            };
+            let lookup_schema = derive_schema_for_pipeline(pipeline.to_owned(), &mut lookup_state)?;
+            insert_required_key_into_document(
```
jchemburkar, line 342: I am forgetting the exact mechanics of how this helper will work, but we should be overwriting the field if it exists in the result_set_schema. Maybe this does this already? but either way, a concise lookup test overwriting the `as` field would be useful.

jchemburkar (changes requested):
The overall stage derivation looks good :) I just had a few more peripheral comments on how were handling the outputs

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -345,3 +345,204 @@ mod unwind {
...
+    test_derive_stage_schema!(
+        union_simple,
+        expected = Ok(Schema::Document(Document {
+            keys: map! {
+                "_id".to_string() => Schema::Atomic(Atomic::ObjectId),
+                "food".to_string() => Schema::Atomic(Atomic::String),
+                "baz".to_string() => Schema::Atomic(Atomic::String),
+                "qux".to_string() => Schema::Atomic(Atomic::Integer),
+            },
+            required: set!(),
```
mattChiaravalloti, line 516: Could there be a case where a field is required in the output, like if it appears on both sides of the union? Should we test that if it is possible?

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -149,6 +169,184 @@ impl DeriveSchema for Stage {
...
+            project: &ProjectStage,
+            state: &mut ResultSetState,
+        ) -> Result<Schema> {
+            if project
+                .items
+                .iter()
+                .all(|(k, p)| matches!(p, ProjectItem::Exclusion) || k == "_id")
+            {
+                // TODO: handle exclusion $project stages
+                todo!();
```
mattChiaravalloti, line 210: Is this something we intend to do for this ticket, or is there a follow-up ticket for it? Should we note the follow-up in the comment if there is one?

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: Yea, I'm not a fan of # fields.

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: Yes, good point

mattChiaravalloti (approved):
looks good 👍

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
pmeredit: That's really just testing schema union.

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: There is a ticket for $project already! I just added this for testing.

mattChiaravalloti: Oooh that makes sense. Thanks for letting me know.

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
mattChiaravalloti: True, good point.

pmeredit:
> looks good 👍

You know, I can make one of the two tests use the same schema without adding _another_ test though, so I might as well :joy:

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: I almost left this exact same comment 😂

jchemburkar (approved):
lgtm!

PR #32, opened 2025-02-21
SQL-2645: avoid excessive nesting of $setFields

terakilobyte:
I changed the logic of how we translate documents, only nesting $setField when necessary.  I also added a test that shows our ability to deal with this.

Before the changes:

```sh
failures:

---- generated::spec_tests_query_tests_select::SELECT_many_fields_with_some_aliases_containing_dots_and_starting_with_dollars stdout ----
thread 'generated::spec_tests_query_tests_select::SELECT_many_fields_with_some_aliases_containing_dots_and_starting_with_dollars' panicked at e2e-tests/src/generated/spec_tests_query_tests_select.rs:2527:59:
called `Result::unwrap()` on an `Err` value: MongoDBErr(Error { kind: Command(CommandError { code: 15, code_name: "Overflow", message: "BSONObj exceeds maximum nested object depth in element with field name 'pipeline.2.$project.__bot.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField.input.$setField' in object with unknown _id", topology_version: None }), labels: {}, wire_version: Some(25), source: None })
```


tests/spec_tests/query_tests/select.yml
```diff
@@ -109,3 +109,9 @@ tests:
...
+    result:
+      - { '':  { '$1': 1, '2.0': 2, '$3': 3, '$4': 4, '5.0': 5, '$6.': 6, '7': 7, '8': 8, '9': 9, '10': 10, '11': 11, '12': 12, '13': 13, '14': 14, '15': 15, '16': 16, '17': 17, '18': 18, '19': 19, '20': 20, '21': 21, '22': 22, '23': 23, '24': 24, '25': 25, '26': 26, '27': 27, '28': 28, '29': 29, '30': 30, '31': 31, '32': 32, '33': 33, '34': 34, '35': 35, '36': 36, '37': 37, '38': 38, '39': 39, '40': 40, '41': 41, '42': 42, '43': 43, '44': 44, '45': 45, '46': 46, '47': 47, '48': 48, '49': 49, '50': 50, '51': 51, '52': 52, '53': 53, '54': 54, '55': 55, '56': 56, '57': 57, '58': 58, '59': 59, '60': 60, '61': 61, '62': 62, '63': 63, '64': 64, '65': 65, '66': 66, '67': 67, '68': 68, '69': 69, '70': 70, '71': 71, '72': 72, '73': 73, '74': 74, '75': 75, '76': 76, '77': 77, '78': 78, '79': 79, '80': 80, '81': 81, '82': 82, '83': 83, '84': 84, '85': 85, '86': 86, '87': 87, '88': 88, '89': 89, '90': 90, '91': 91, '92': 92, '93': 93, '94': 94, '95': 95, '96': 96, '97': 97, '98': 98, '99': 99 } }
```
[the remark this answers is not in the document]
pmeredit: I really don't think this test needs to be this big :joy:, but it's cool

pmeredit (approved):
Looks great. I'm actually kind of surprised we didn't do it like this to begin with.

tests/spec_tests/query_tests/select.yml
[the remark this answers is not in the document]
terakilobyte: It needed to be this big to fail initially without the changes, unfortunately :(

tests/spec_tests/query_tests/select.yml
```diff
@@ -109,3 +109,9 @@ tests:
     result:
       - {'a': {'_id': 0, 'a': 1, 'b': 1}, '': {'a': 1}}
       - {'a': {'_id': 1, 'a': 2, 'b': 2}, '': {'a': 2}}
+
+  - description: SELECT many fields with some aliases containing dots and starting with dollars
```
mattChiaravalloti, line 113: This should not be a spec test, it should be an e2e test. It is not demonstrating anything relevant to the specification, rather it is demonstrating an implementation-specific issue based on the mongodb back-end. Can you move this to the `e2e` test suite?

mongosql/src/translator/expressions.rs
```diff
@@ -187,48 +187,55 @@ impl MqlTranslator {
...
-        Ok(air::Expression::Document(doc_expr))
+        // set fields are required to be set inside a SetField expression.
+        // The document expression calculated above is the initial input to the
+        // SetField expression(s). The SetField expression is
+        // chained with other SetField expressions for each set field.
+        let mut set_field_expr = air::Expression::Document(doc_expr);
+        for (k, v) in set_fields {
+            if k.is_empty() {
+                return Err(Error::InvalidDocumentKey(k));
+            }
```
mattChiaravalloti, line 230: I understand why we have this as a "just-in-case" but shouldn't this be impossible? The `partition` method separated any empty strings into the first group, `normal_fields`, leaving only strings that start with `$` or include `.` in this set.

mattChiaravalloti (changes requested):
Looks good except where the test lives, and we can possibly remove an error conditional that is impossible to reach.

mongosql/src/translator/expressions.rs
terakilobyte: Updated to be handled during initial partitioning

tests/spec_tests/query_tests/select.yml
terakilobyte: moved

mongosql/src/translator/expressions.rs
```diff
@@ -199,17 +199,25 @@ impl MqlTranslator {
...
-        let (normal_fields, set_fields): (Vec<_>, Vec<_>) = mir_document
-            .into_iter()
-            .partition(|(k, _)| !k.starts_with('$') && !k.contains('.'));
+        // Empty fields are invalid.
+        let (normal_fields, set_fields): (Vec<_>, Vec<_>) = mir_document.into_iter().try_fold(
+            (Vec::new(), Vec::new()),
+            |(mut normal_fields, mut set_fields), (k, v)| {
+                if k.is_empty() {
+                    return Err(Error::InvalidDocumentKey(k));
+                }
```
mattChiaravalloti, line 208: Sorry, I guess my previous comment was unclear. If you wanted to keep using the `partition` method to keep this more concise, that's totally fine. I just meant we don't need the `if k.is_empty()` check in both for loops (for `normal_fields` and for `set_fields`) since all empty keys would by definition be included in the `normal_fields` vector.

This current version is totally correct and fine though so no need to change unless you have a preference!

mattChiaravalloti (approved):
looks good 💯

PR #33, opened 2025-02-25
SQL-2664: Fix document comparison logic in test runner to check for matching key

bucaojit:
Change to the document comparison logic to also check that they keys match. 
It required some minor changes to our existing tests, many of which are expected with the fix.  One is unexpected related to count(distinct *) which I've mentioned to @mattChiaravalloti

### DIFF
--- a/mongosql/src/air/desugarer/remove_id.rs
+++ b/mongosql/src/air/desugarer/remove_id.rs
@@ -28,6 +28,6 @@ impl Visitor for RemoveIdDesugarerVisitor {
                 .insert(id, ProjectItem::Exclusion)
                 .unwrap();
         }
-        node
+        node.walk(self)
     }
 }
--- a/mongosql/src/air/desugarer/testdata/desugar_all.yml
+++ b/mongosql/src/air/desugarer/testdata/desugar_all.yml
@@ -62,7 +62,10 @@ tests:
                 },
                 {
                   "$project":
-                  {  "_id": 0, "eca58228-b657-498a-b76e-f48a9161a404": 0 },
+                    {
+                      "_id": 0,
+                      "eca58228-b657-498a-b76e-f48a9161a404": 0
+                    },
                 },
               ],
             "as": "__subquery_result_0",
@@ -71,6 +74,7 @@ tests:
       - {
         "$project":
           {
+            "_id": 0,
             "expr":
               {
                 "$let":
@@ -111,8 +115,8 @@ tests:
             "let": { "bar_0": "$bar" },
             "pipeline":
               [
-                { "$project": { "coll": "$$ROOT" } },
-                { "$project": { "coll": "$coll" } },
+                { "$project": { "_id": 0, "coll": "$$ROOT" } },
+                { "$project": { "_id": 0, "coll": "$coll" } },
                 {
                   "$match":
                     { "$expr":
@@ -271,9 +275,9 @@ tests:
                                     },
                                     {
                                       "$project": {
-                                          "_id": 0,
-                                          "eca58228-b657-498a-b76e-f48a9161a404": 0,
-                                        },
+                                        "_id": 0,
+                                        "eca58228-b657-498a-b76e-f48a9161a404": 0,
+                                      },
                                     },
                                   ],
                                 "as": "__subquery_result_0",
@@ -282,6 +286,7 @@ tests:
                           {
                             "$project":
                               {
+                                "_id": 0,
                                 "expr":
                                   {
                                     "$let":
@@ -301,7 +306,7 @@ tests:
                           },
                           {
                             "$project":
-                              { "__subquery_result_0": 0 },
+                              { "_id": 0, "__subquery_result_0": 0,  },
                           },
                         ],
                       "as": "eca58228-b657-498a-b76e-f48a9161a404",
@@ -332,6 +337,7 @@ tests:
       - {
         "$project":
           {
+            "_id": 0,
             "expr":
               {
                 "$let":
@@ -441,6 +447,7 @@ tests:
       - {
         "$project":
           {
+            "_id": 0,
             "expr":
               {
                 "$let":
@@ -501,6 +508,7 @@ tests:
                 {
                   "$project":
                     {
+                      "_id": 0,
                       "expr":
                         {
                           "$let":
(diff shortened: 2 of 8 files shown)


tests/e2e_tests/ext_json_implicit_type_conversion.yml
```diff
@@ -56,13 +56,13 @@ tests:
     current_db: test_db
     query: "SELECT '{\"$numberInt\": \"1\"}' AS s FROM [{}] AS t"
     result:
-      - { 't': { 's': '{"$numberInt": "1"}' } }
+      - { '': { 's': '{"$numberInt": "1"}' } }
```
mattChiaravalloti, line 59: Wow yes, this is a huge difference! Excellent find and fix 💯

It's great, of course, that the implementation was actually correct all along and just the test runner was buggy but still it's obviously so critical to have a correct test framework 🤦‍♀️

mattChiaravalloti (approved):
lgtm 💯

test-utils/src/query/mod.rs
```diff
@@ -218,8 +218,8 @@ pub fn compare_documents(expected: &Document, actual: &Document, type_compare: b
     if expected.len() != actual.len() {
         return false;
     }
-    expected.iter().all(|(ek, ev)| {
-        actual.iter().any(|(ak, av)| {
+    expected.iter().all(|(ek, ev)| match actual.get(ek) {
```
terakilobyte, line 221: I think we could sort this up since each `if` clause returns the result of its evaluation when it matches. Not a required change.

```suggestion
    expected
        .iter()
        .all(|(ek, expected_value)| match actual.get(ek) {
            Some(actual_value) => match (expected_value, actual_value) {
                (Bson::Document(expected_document), Bson::Document(actual_document)) => {
                    compare_documents(expected_document, actual_document, type_compare)
                }
                (Bson::Array(expected_array), Bson::Array(actual_array)) => {
                    compare_arrays(expected_array, actual_array, type_compare)
                }
                _ if type_compare => expected_value.element_type() == actual_value.element_type(),
                _ if is_numeric(expected_value) && is_numeric(actual_value) => {
                    compare_doubles_for_test(
                        numeric_to_double(expected_value),
                        numeric_to_double(actual_value),
                    )
                }
                _ => expected_value == actual_value,
            },
            None => false,
        })
}
```

terakilobyte (changes requested):
I found when testing this locally, a test was failing. It appears this is unique to running this test on macos.

```sh
failures:

---- generated::spec_tests_query_tests_order_by::Sorting_by_column_from_RHS_of_LEFT_JOIN_orders_the_entire_result_set stdout ----

thread 'generated::spec_tests_query_tests_order_by::Sorting_by_column_from_RHS_of_LEFT_JOIN_orders_the_entire_result_set' panicked at e2e-tests/src/generated/spec_tests_query_tests_order_by.rs:3514:17:
unexpected query result for Sorting by column from RHS of LEFT JOIN orders the entire result set at index 5, 
expected: Document({"": Document({"_id": Int32(0), "x": Int32(2)})})
actual: Document({"": Document({"_id": Int32(1), "x": Int32(2)})})
```

I dug into it more, and the difference is in the order of results between 6.0.6 and 8.0.1. This only appears to be different on mac. Looking at the query and result, *both* results are correct since the query only sorts on `x`. Could we update the query to
`SELECT l._id, f.x FROM local AS l LEFT JOIN foreign AS f ON l._id = f.l_id ORDER BY x, l._id ASC` so that testing agrees universally?

We may want to file a ticket to look at all of our sort tests to ensure they are 100% deterministic.

test-utils/src/query/mod.rs
mattChiaravalloti: This is a fun use of `_ <condition>` patterns

mattChiaravalloti:
> One is unexpected related to count(distinct *) which I've mentioned to @mattChiaravalloti

We resolved this one offline. I'll take a look again when that is in with unit tests!

mattChiaravalloti:
> I dug into it more, and the difference is in the order of results between 6.0.6 and 8.0.1. This only appears to be different on mac. Looking at the query and result, _both_ results are correct since the query only sorts on `x`. Could we update the query to `SELECT l._id, f.x FROM local AS l LEFT JOIN foreign AS f ON l._id = f.l_id ORDER BY x, l._id ASC` so that testing agrees universally?

Good suggestion. It's very interesting that our dependence on implicit ordering hasn't been apparent earlier than this. Generally, though, you're totally right we should ensure we order by all relevant selected fields in tests that have `ordered: true`.

test-utils/src/query/mod.rs
bucaojit: Cool, I added this implementation.

bucaojit:
@mattChiaravalloti I added some unit tests and updated the desugarer tests to have them pass with the remove _id change.  Re-tagging you for review on those.

mongosql/src/air/desugarer/remove_id.rs
```diff
@@ -28,6 +28,6 @@ impl Visitor for RemoveIdDesugarerVisitor {
                 .insert(id, ProjectItem::Exclusion)
                 .unwrap();
         }
-        node
+        node.walk(self)
```
mattChiaravalloti, line 31: The comment at the top of this file says "This removes _id using the last Project stage in the pipeline". Can we update that to accurately reflect what we are doing now?

For Nathan: it seems at some point we changed our translations to go from excluding `_id` in every `$project` stage to only doing it in the _last_ `$project` stage. Excluding in only the last `$project` breaks some of the critical assumptions we relied on for our `DISTINCT` implementations. Chiefly, that only the specifically selected fields would be projected within the pipeline. This change is necessary to restore that invariant. The good news is that an exclusion is very cheap (and the "unnecessary" further ones down the pipeline should even be free).

mongosql/src/air/desugarer/testdata/desugar_all.yml
mattChiaravalloti, line 1: Glad to see this is the only spot that needed to be updated. It's nice we got rid of translations for every spec test 😌

mattChiaravalloti (approved):
lgtm 👍

terakilobyte (approved):
LGTM, nice updates/fixes.

PR #35, opened 2025-02-25
SQL-2369: Add $group, $setWindowFields, $addFields, $count, $skip, $limit, $collection schema derivation

pmeredit:
It was weird that we were using normal Tagged and UntaggedOps for $setWindowFields but GroupAccumulators for $group, so I undid the GroupAccumulator thing and moved us over to Tagged and Untagged ops for Group. Part of the motivation is going through the window ops I realized we were missing quite a few GroupAccumlators, and this was the easiest way to support this, and also the cleanest, imo.

### DIFF
--- a/agg-ast/ast/src/custom_serde.rs
+++ b/agg-ast/ast/src/custom_serde.rs
@@ -1,11 +1,10 @@
 use crate::{
     definitions::{
         Cond, Convert, DateExpression, DateFromParts, DateFromString, DateToString, Expression,
-        GroupAccumulator, GroupAccumulatorExpr, GroupAccumulatorName, LiteralValue,
-        MatchArrayExpression, MatchArrayQuery, MatchBinaryOp, MatchElement, MatchExpression,
-        MatchField, MatchNot, MatchNotExpression, MatchRegex, MatchStage, ProjectItem,
-        ProjectStage, Ref, SetWindowFieldsOutput, Trim, UntaggedOperator, UntaggedOperatorName,
-        VecOrSingleExpr, Window,
+        LiteralValue, MatchArrayExpression, MatchArrayQuery, MatchBinaryOp, MatchElement,
+        MatchExpression, MatchField, MatchNot, MatchNotExpression, MatchRegex, MatchStage,
+        ProjectItem, ProjectStage, Ref, SetWindowFieldsOutput, Trim, UntaggedOperator,
+        UntaggedOperatorName, VecOrSingleExpr, Window,
     },
     map,
 };
@@ -85,7 +84,6 @@ macro_rules! try_from_serde {
 }
 
 try_from_serde!(UntaggedOperatorName);
-try_from_serde!(GroupAccumulatorName);
 try_from_serde!(MatchBinaryOp);
 
 impl MatchArrayQueryVisitor {
@@ -946,59 +944,6 @@ impl<'de> Visitor<'de> for UntaggedOperatorVisitor {
         Err(serde_err::custom(
             "fail when there are no keys; this lets empty doc be parsed as Document",
         ))
-    }
-}
-
-impl<'de> Deserialize<'de> for GroupAccumulator {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        /// Custom map visitor for identifying and deserializing Accumulators.
-        struct AccumulatorVisitor;
-
-        impl<'de> Visitor<'de> for AccumulatorVisitor {
-            type Value = GroupAccumulator;
-
-            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-                formatter.write_str("{\"$op\": <expression or struct>}")
-            }
-
-            fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
-            where
-                M: MapAccess<'de>,
-            {
-                let kv = access.next_entry::<String, GroupAccumulatorExpr>()?;
-                if let Some((key, value)) = kv {
-                    let function = key.try_into().map_err(serde_err::custom)?;
-                    // Immediately return when we see one key that starts with a "$".
-                    // In a general environment, this would be very brittle, however in this
-                    // controlled test environment, we safely make the assumption that
-                    // a single key that starts with a "$" is present and indicates an operator.
-                    // let value = value.get_as_vec();
-                    return Ok(GroupAccumulator {
-                        function,
-                        expr: value,
-                    });
-                }
-
-                Err(serde_err::custom("no accumulator could be parsed"))
-            }
-        }
-
-        const FIELDS: &[&str] = &["function", "expr"];
-        deserializer.deserialize_struct("GroupAccumulator", FIELDS, AccumulatorVisitor)
-    }
-}
-
-impl ser::Serialize for GroupAccumulator {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: ser::Serializer,
-    {
-        let mut map = serializer.serialize_map(Some(1))?;
-        map.serialize_entry(&self.function, &self.expr)?;
-        map.end()
     }
 }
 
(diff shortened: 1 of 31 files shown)


agg-ast/ast/src/custom_serde.rs
```diff
@@ -949,59 +947,6 @@ impl<'de> Visitor<'de> for UntaggedOperatorVisitor {
     }
 }
```
pmeredit, line 949: Love to see code removed

agg-ast/ast/src/definitions.rs
```diff
@@ -22,7 +22,7 @@ use serde::{Deserialize, Serialize};
 /// Stage represents an aggregation pipeline stage.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum Stage {
```
pmeredit, line 24: collection isn't actually used by mongodb, but there is no reason not to support it.

agg-ast/ast/src/definitions.rs
```diff
@@ -1278,6 +1261,28 @@ pub enum TaggedOperator {
     Shift(Shift),
     #[serde(rename = "$cond")]
     Cond(Cond),
```
pmeredit, line 1263: These made sense to just add as TaggedOperators, fun thing is they all use the same underlying struct

agg-ast/ast/src/serde_test.rs
```diff
@@ -1089,8 +1089,8 @@ mod stage_test {
     mod group_test {
         use crate::{
```
pmeredit, line 1090: Ignore this file, this is just $group refactor

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -725,11 +823,11 @@ impl DeriveSchema for TaggedOperator {
                 state,
                 Schema::Atomic(Atomic::Double),
             ),
```
pmeredit, line 825: This seems to have been a mistake, $percentile always returns arrays in all contexts

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1111,6 +1211,56 @@ impl DeriveSchema for TaggedOperator {
                 }
                 Ok(Schema::Array(Box::new(array_schema)))
             }
```
pmeredit, line 1213: This is unfortunate, I tried to use a macro, but you can't use macros for pattern expansions, sadly.

agg-ast/ast/src/definitions.rs
mattChiaravalloti: We must have used this in the desugarer at some point and just kept it around all this time. The desugarer does not currently use this so it's unclear what it's original purpose was.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -208,6 +208,88 @@ impl DeriveSchema for Stage {
...
+                .collect::<Result<BTreeMap<_, _>>>()?;
+            let Schema::Document(Document {
+                keys: ref current_keys,
+                ..
+            }) = state.result_set_schema
+            else {
+                // This should never actually happen
+                return Err(Error::InvalidResultSetSchema(
+                    state.result_set_schema.to_string(),
+                ));
```
mattChiaravalloti, line 253: Should we `panic` here instead since this is considered an unreachable state? If we got here, it would mean there is a mistake in our code, not somehow an invalid input.

mattChiaravalloti (approved):
This looks great 👍

agg-ast/ast/src/definitions.rs
pmeredit: I think it's supported on ADF?

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: That's true, was just curious to see this if it ever happened, but it really shouldn't.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -208,6 +208,88 @@ impl DeriveSchema for Stage {
...
+                keys,
+                ..Default::default()
+            }))
+        }
+
+        // add_fields_derive_schema is near set_window_fields_derive_schema because they are necessarily
+        // identical in semantics: they both add fields to the schema incoming from the previous
+        // pipeline stage. The only difference is that add_fields is a map of expressions, while set_window_fields
+        // is a map of window functions.
+        fn add_fields_derive_schema(
```
jchemburkar, line 234: perhaps I'm missing a little nuance here but wouldn't we just directly mutate the result set schema and use the insert_required_... helper?
```
for (field, expression) in add_fields.iter() {
    let expr_schema = expression.derive_schema(state)?;
    let path = field
        .split(".")
        .map(|s| s.to_string())
        .collect::<Vec<String>>();
    insert_required_key_into_document(
        &mut state.result_set_schema,
        expr_schema,
        path,
        true,
    );
}
Ok(Schema::simplify(&state.result_set_schema.to_owned()))
```

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -483,30 +565,46 @@ impl DeriveSchema for Stage {
...
+                    .catalog
+                    .get(&ns)
+                    .ok_or(Error::UnknownReference(ns.to_string()))?
+                    .clone();
+                Ok(state.result_set_schema.to_owned())
+            }
+            Stage::Count(c) => {
+                state.result_set_schema = Schema::Document(Document {
+                    keys: map! {
+                        c.clone() => Schema::AnyOf(set!{Schema::Atomic(Atomic::Integer), Schema::Atomic(Atomic::Long), Schema::Atomic(Atomic::Double)})
```
jchemburkar, line 586: is this right, count can return a double?

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -7,6 +7,62 @@ use mongosql::{
 };
 use std::collections::BTreeMap;
 
+mod add_fields {
+    use super::*;
+
+    test_derive_stage_schema!(
+        add_fields,
```
jchemburkar, line 14: I took a stab at this as part of SQL-2366, had a couple test cases in mind that might be worth adding: https://github.com/mongodb/mongosql/pull/37/files#diff-99e52b8280797b2fa94ad9864f86606367010126938e62b9d30fe2e3525d62d9

I think the add fields with $$REMOVE and add fields with a nested field would be the ones I'm most interested in adding here (I just took the examples from the docs and tested each of them)

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -649,6 +767,106 @@ mod graphlookup {
...
+                "set".to_string(),
+                "push".to_string(),
+                "avg".to_string(),
+                "bottom".to_string(),
+                "bottomN".to_string(),
+                "count".to_string()
+            ),
+            ..Default::default()
+        })),
+        input = r#"{"$setWindowFields": {
```
jchemburkar, line 816: nightmare stage 😂

jchemburkar (approved):
lgtm, awesome stuff. Just had a couple thoughts on addFields

agg-ast/schema_derivation/src/schema_derivation.rs
mattChiaravalloti: https://www.mongodb.com/docs/manual/reference/operator/aggregation/count/#behavior

Yes it can!

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: Nah, that seems like a good idea.

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: It seems like it can, but not for realistic collections, I'm going to back off on this even though it technically can return double if we overflow 2^63 documents.

PR #37, opened 2025-02-27
SQL-2366: implement schema derivation for field setting stages

jchemburkar:


pmeredit (approved):
LGTM, great tests!

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1141,3 +1141,243 @@ mod union {
         })
     );
 }
+
+mod replace {
+    use super::*;
+
+    test_derive_stage_schema!(
+        replace_root_nested_doc,
```
nbagnard, line 1149: Can you add a test where replace root is using a field which does not appear in the source schema to make sure things are working as expected.

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1141,3 +1141,243 @@ mod union {
...
+                    ..Default::default()
+                }),
+                "foo".to_string() => Schema::Atomic(Atomic::Decimal),
+            },
+            required: set!("foo".to_string(), "name".to_string()),
+            ..Default::default()
+        })
+    );
+
+    test_derive_stage_schema!(
```
nbagnard, line 1206: Can you add a test where replace_with is using a field which does not appear in the source schema to make sure things are working as expected.

nbagnard (changes requested):
Just a few edge cases that I think are worth testing out.
Otherwise, it looks great :)

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
jchemburkar: Would the test below cover what you are looking for already? it creates new fields using a separate operator and tests that those are correctly in the replaced schema

jchemburkar: if so this would be the same question for replaceRoot, which also has a test where an expression is being used to replaceRoot rather than simply a field ref

nbagnard: You are right, it covers the same idea and I found the unit test for field_ref_missing, so that should be alright 👍

nbagnard (approved):
LGTM!

PR #38, opened 2025-02-27
SQL-2626: Add spec tests for subquery comparison expressions

mattChiaravalloti:
This PR adds type constraint spec tests for subquery comparison expressions. In the design doc, we initially said we would query test the new document and array comparison features for subquery expressions. However, as I audited the existing spec query and spec type constraint tests, I realized (1) we do not have any other type-related query tests for subquery comparison expressions and (2) we did not have any existing spec type constraint tests for subquery comparison expressions. Given that, I decided to make these new type constraint tests and they demonstrate exactly what we wanted to check.

### DIFF
--- a/tests/spec_tests/type_constraint_tests/README.md
+++ b/tests/spec_tests/type_constraint_tests/README.md
@@ -59,10 +59,10 @@ Here is an example of such a test, followed by a brief explanation:
 ```
 In this example, the query contains the expression `arg1 + arg2`, an addition operation.
 There is one map in the `valid_types` sequence. It indicates that `arg1` can be any
-numeric type, `NULL`, or missing; it also indicates that `arg2` can be any numeric type,
-`NULL`, or missing. Since these lists are included in the same map, that means any
-combination from the cross product of the lists is valid (i.e. `INT, INT`, `LONG, DECIMAL`,
-and `DOUBLE,  NULL` are all valid, among others).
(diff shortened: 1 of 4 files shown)


tests/spec_tests/type_constraint_tests/README.md
```diff
@@ -59,10 +59,10 @@ Here is an example of such a test, followed by a brief explanation:
 ```
 In this example, the query contains the expression `arg1 + arg2`, an addition operation.
 There is one map in the `valid_types` sequence. It indicates that `arg1` can be any
-numeric type, `NULL`, or missing; it also indicates that `arg2` can be any numeric type,
-`NULL`, or missing. Since these lists are included in the same map, that means any
```
mattChiaravalloti, line 63: Patrick recently realized we can't actually support `MISSING` in these tests using the test runner as implemented because it is invalid for a field's sole schema to be `MISSING`. So we removed missing from the test suite.

This change is just some lingering clean up to remove references to missing.

tests/spec_tests/type_constraint_tests/join.yml
mattChiaravalloti, line 1: Since I added the same `keys`/`required` for `foo` to `bar`, we need to disambiguate in the queries now. That's not a bad tradeoff by any means.

bucaojit (approved):
LGTM!

terakilobyte (approved):
LGTM :shipit:

PR #39, opened 2025-02-28
SQL-2541: implement schema derivation for $filter

jchemburkar:


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1146,7 +1146,25 @@ impl DeriveSchema for TaggedOperator {
                     set! {then_schema, else_schema},
                 )))
             }
-            TaggedOperator::Filter(_) => todo!(),
+            TaggedOperator::Filter(f) => {
+                let input_schema = f.input.derive_schema(state)?;
+                let is_nullable = input_schema.satisfies(&NULLISH.clone()) != Satisfaction::Not;
+                let array_schema = array_schema_or_error!(input_schema, f.input);
```
mattChiaravalloti, line 1152:
```suggestion
                let array_element_schema = array_schema_or_error!(input_schema, f.input);
```
If I understand the macro correctly, this is getting the schema from within a `Schema::Array`. So that represents the schema of the elements of the array, right? Could we consider this rename for the variable, and even possibly changing the name of the macro (and adding a comment to it) to reflect that?

<hr>

Side note: is the `array_schema_or_error` macro complete? Why does it only take the first `Array` schema found in an `AnyOf` as opposed to collecting all possible array schemas? For example, what about something like `AnyOf([Array(Int), Array(String)])`?

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1146,7 +1146,25 @@ impl DeriveSchema for TaggedOperator {
...
+                let filter_schema = state.variables.remove(&var_name);
+                if let Some(schema) = filter_schema {
+                    let mut schema = Schema::Array(Box::new(schema));
+                    if is_nullable {
+                        schema = schema.union(&Schema::Atomic(Atomic::Null))
+                    }
+                    Ok(schema)
+                // this should be unreachable, since we manually add and remove the variable from the state
+                } else {
+                    Ok(Schema::Unsat)
```
mattChiaravalloti, line 1165: Should we panic instead of return `Unsat` if it's unreachable? I know Patrick updated an error case to a panic in one of his PRs I reviewed recently. Not sure what our convention is in this situation throughout the library (or if it even matters what choice we make).

agg-ast/schema_derivation/src/schema_derivation_tests/tagged_ops.rs
```diff
@@ -1409,3 +1409,95 @@ mod convert {
         string_rep = "maxKey"
     );
 }
+
+mod filter {
```
mattChiaravalloti, line 1413: Can you add a test demonstrating the filter expression reducing the schema? For example, consider a starting schema like
```
starting_schema = Document {
    keys: {
        _id: Atomic(Int),
        items: Array(Document {
            keys: {
                item_id => AnyOf([Int, String]),
                ...
            }
        })
    },
    ...
}
```
The condition could be something like `{$isNumber: "$$item.item_id"}` (or similar expression that indicates type info), so the expected result would reduce `item_id`'s schema from `AnyOf([Int, String])` to just `Int`.

This test would be unique from the existing tests since the existing tests would pass even if we just directly returned the `input` field's inner schema and we didn't derive the schema for the condition. This new test would demonstrate we do use the condition expression to determine a more precise schema. It looks like the implementation is already set up to do this so this new test should pass without implementation changes.

mattChiaravalloti (changes requested):

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: `AnyOf([Array(Int), Array(String)])` _should_ be impossible, but that's something we may want to audit. Simplify should definitely collapse them. Union already does.

pmeredit: I know `ArrayOf([AnyOf([Int, String])])` is actually less precise, but that's one of those things that is fine to keep sizes down.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1146,7 +1146,25 @@ impl DeriveSchema for TaggedOperator {
...
-            TaggedOperator::Filter(_) => todo!(),
+            TaggedOperator::Filter(f) => {
+                let input_schema = f.input.derive_schema(state)?;
+                let is_nullable = input_schema.satisfies(&NULLISH.clone()) != Satisfaction::Not;
+                let array_schema = array_schema_or_error!(input_schema, f.input);
+                let var_name = f._as.clone().unwrap_or("this".to_string());
+                state.variables.insert(var_name.clone(), array_schema);
+                f.cond.derive_schema(state)?;
+                let filter_schema = state.variables.remove(&var_name);
+                if let Some(schema) = filter_schema {
```
pmeredit, line 1157: I really like the idea of inserting and removing the variable, but this is unfortunately (very rarely) incorrect. If that variable already exists in the state, we are losing it forever here. We, unfortunately, need to clone the state and revert to it rather than this much more efficient implementation... or we can just check for key collision and only restore that one key. Up to you.

pmeredit (changes requested):
This looks great other than that one rare small bug. Once we fix that, should be good!

agg-ast/schema_derivation/src/schema_derivation.rs
mattChiaravalloti: Ooh ok yes I should have checked `simplify`. I had a feeling it would combine those into `Array(AnyOf([Int, String]))` but I decided to just not check and assume since that's a different definition we didn't do it. Since we do do that, though, that's totally fine. No reason to change the macro implementation!

I still think a name change for the macro and the variable could be useful for reading purposes.

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: ahhhh good call, I'll handle this with the clone

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: Yeah this is a good callout. I think the compass I have mostly been following is error when you have a reachable issue -- eg an operator that takes an `Expression` but that expression must be an array literal or field reference but cannot be something else; and panicing when a pipeline is in fact unreachable. I'll do that here!

agg-ast/schema_derivation/src/schema_derivation_tests/tagged_ops.rs
jchemburkar: This is a good shout and revealed some interesting implications! First off, it revealed what I suggested should be the implementation but failed to do correctly here -- we should have been doing match_derive_schema to filter the documents that come in.

But more critically, we haven't treated variables as paths in any of our helpers! Using the docs example, we get the nested $$items.price, which failed because we just stored the "items" variable and would try to get "items.price" as a single key.

I updated this to split on '.' and get the schema document from the first part of the path, then use the existing get_schema_for_path_mut to get the nested key from the top level schema document. This comes up for both the expression schema derivation and for the intersect_if_exists function

agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -3388,32 +3388,39 @@ mod array_ops {
...
+            keys: map! {
+                "foo".to_string() => Schema::AnyOf(set!(
+                    Schema::Array(Box::new(Schema::Any)),
+                    Schema::Atomic(Atomic::Null)
+                ))
+            },
+            ..Default::default()
+        })),
+        input = r#"{"$match": {"$expr": {"$ne": [[], {"$filter": {"input": "$foo", "as": "item", "cond": {"$gte": ["$$item", 1]}}}]}}}"#,
+        ref_schema = Schema::AnyOf(set!(
```
jchemburkar, line 3403: forgot to uncomment these originally and noticed as I was addressing feedback. ref schema changed to be a maximal schema that still compiles

agg-ast/schema_derivation/src/schema_derivation_tests/tagged_ops.rs
mattChiaravalloti: Oh wow that's great that we sorted this out by testing. I vaguely recall when we decided to introduce variables we realized we wouldn't actually need to access the path components (at the time). Looks like the time has come to make that a more structured type. Looking forward to checking back in on the updates for this.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -693,10 +694,21 @@ impl DeriveSchema for Expression {
             Expression::Ref(Ref::VariableRef(v)) => match v.as_str() {
                 "REMOVE" => Ok(Schema::Missing),
                 "ROOT" => Ok(state.result_set_schema.clone()),
```
pmeredit, line 696: There are a few more variables in mongo I forgot we should add:

```
NOW: A variable that returns the current datetime value. NOW returns the same value for all members of the deployment and remains the same throughout all stages of the aggregation pipeline.
CLUSTER_TIME: A variable that returns the current timestamp value. CLUSTER_TIME is only available on replica sets and sharded clusters. CLUSTER_TIME returns the same value for all members of the deployment and remains the same throughout all stages of the pipeline.
ROOT: References the root document, i.e., the top-level document, currently being processed in the aggregation pipeline stage.
CURRENT: References the start of the field path being processed in the aggregation pipeline stage. Unless documented otherwise, all stages start with CURRENT the same as ROOT. CURRENT is modifiable. However, since $ is equivalent to $$CURRENT., rebinding CURRENT changes the meaning of $ accesses.
REMOVE: A variable which evaluates to the missing value, allowing for the conditional exclusion of fields. In a $project, a field set to the variable REMOVE is excluded from the output.
DESCEND, PRUNE, and KEEP: These are allowed results of a $redact expression.
SEARCH_META: A variable that stores the metadata results of an Atlas Search query. In all supported aggregation pipeline stages, a field set to the variable $$SEARCH_META returns the metadata results for the query.
USER_ROLES: Returns the roles assigned to the current user. This variable is new in version 7.0.
```

We should probably file a ticket for these

pmeredit: This is the entire list, we've obviously done ROOT and REMOVE already.

PR #40, opened 2025-03-01
SQL-2620: Add support for UNION distinct

bucaojit:
Implementing UNION distinct.  It differs a bit from the initial design, I applied some of the things we did for SELECT DISTINCT such as using the `__groupKey{}` alias.  
Added tests to confirm behavior, removed `skip_reasons` from the existing union distinct tests and moved them into `distinct.yml`

### DIFF
--- a/mongosql/src/algebrizer/test.rs
+++ b/mongosql/src/algebrizer/test.rs
@@ -7454,20 +7454,256 @@ mod limit_or_offset_clause {
 }
 
 mod set_query {
-    use super::{catalog, mir_source_bar, mir_source_foo, AST_QUERY_BAR, AST_QUERY_FOO};
-    use crate::{ast, mir, mir::schema::SchemaCache, usererror::UserError};
+    use super::{
+        catalog, mir_source_bar, mir_source_foo, AST_QUERY_BAR, AST_QUERY_FOO, AST_SOURCE_BAR,
+        AST_SOURCE_FOO,
+    };
+    use crate::schema::{Document, Schema};
+    use crate::{
+        ast, map, mir, mir::schema::SchemaCache, multimap, schema, set,
+        unchecked_unique_linked_hash_map,
+    };
+    use mongosql_datastructures::binding_tuple::DatasourceName;
 
     test_algebrize!(
-        union_distinct_not_allowed,
+        union_distinct_star,
         method = algebrize_set_query,
-        expected = Err(Error::DistinctUnion),
-        expected_error_code = 3006,
+        expected = Ok(mir::Stage::Project(mir::Project {
+            source: Box::new(mir::Stage::Group(mir::Group {
+                source: Box::new(mir::Stage::Set(mir::Set {
+                    operation: mir::SetOperation::UnionAll,
+                    left: Box::new(mir::Stage::Project(mir::Project {
+                        source: Box::new(mir::Stage::Collection(mir::Collection {
+                            db: "test".into(),
+                            collection: "foo".into(),
+                            cache: SchemaCache::new()
+                        })),
+                        expression: map! {
+                            (DatasourceName::Named("foo".into()), 0u16).into() => mir::Expression::Reference((DatasourceName::Named("foo".into()), 0u16).into())
+                        },
+                        is_add_fields: false,
+                        cache: SchemaCache::new()
+                    })),
+                    right: Box::new(mir::Stage::Project(mir::Project {
+                        source: Box::new(mir::Stage::Collection(mir::Collection {
+                            db: "test".into(),
+                            collection: "bar".into(),
+                            cache: SchemaCache::new()
+                        })),
+                        expression: map! {
+                            (DatasourceName::Named("bar".into()), 0u16).into() => mir::Expression::Reference((DatasourceName::Named("bar".into()), 0u16).into())
+                        },
+                        is_add_fields: false,
+                        cache: SchemaCache::new()
+                    })),
+                    cache: SchemaCache::new()
+                })),
+                keys: vec![
+                    mir::OptionallyAliasedExpr::Aliased(mir::AliasedExpr {
+                        alias: "__groupKey0".to_string(),
+                        expr: mir::Expression::Reference(
+                            (DatasourceName::Named("bar".to_string()), 0u16).into()
+                        )
+                    }),
+                    mir::OptionallyAliasedExpr::Aliased(mir::AliasedExpr {
+                        alias: "__groupKey1".to_string(),
+                        expr: mir::Expression::Reference(
+                            (DatasourceName::Named("foo".to_string()), 0u16).into()
+                        )
+                    })
+                ],
+                aggregations: vec![],
+                cache: SchemaCache::new(),
+                scope: 0
+            })),
+            expression: map! {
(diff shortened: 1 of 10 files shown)


tests/spec_tests/query_tests/union.yml
```diff
@@ -63,15 +44,6 @@ tests:
       - {'': {'a': 2}}
       - {'': {'a': 2}}
 
-  - description: union is left associative
-    current_db: mydb
-    query: "SELECT VALUE {'a': a} FROM foo UNION ALL SELECT VALUE {'a': a} FROM foo UNION SELECT VALUE {'a': a} FROM foo"
-    ordered: false
-    skip_reason: "distinct UNION not currently supported"
```
mattChiaravalloti, line 70: Not sure why this one was skipped since it's doing a `UNION ALL`. Should we leave this one here? When you cut and pasted it into the other file, you left it as `UNION ALL` so it's out of place in the `distinct.yml` file.

tests/spec_tests/query_tests/distinct.yml
mattChiaravalloti, line 1: Is `SELECT a, b FROM foo UNION SELECT a, b FROM bar` a valid query? If yes, can we also include a test for that demonstrating that the duplicate rows from the two result sets are deduplicated? We demonstrate this for selectin specific columns from the same collection (`SELECT a FROM foo UNION SELECT a FROM foo`) and we demonstrate this for same column names in a different order (`SELECT a, b FROM foo UNION SELECT b, a FROM bar`) and for `SELECT *` on both sides, but we don't demonstrate deduplication for same columns in the same order from different datasources on the two sides of a UNION. A test covering that would therefore be unique and useful.

mongosql/src/algebrizer/test.rs
```diff
@@ -7454,20 +7454,256 @@ mod limit_or_offset_clause {
...
     test_algebrize!(
-        union_distinct_not_allowed,
+        union_distinct_star,
         method = algebrize_set_query,
-        expected = Err(Error::DistinctUnion),
-        expected_error_code = 3006,
+        expected = Ok(mir::Stage::Project(mir::Project {
+            source: Box::new(mir::Stage::Group(mir::Group {
+                source: Box::new(mir::Stage::Set(mir::Set {
+                    operation: mir::SetOperation::UnionAll,
```
mattChiaravalloti, line 7474: Ah very interesting revelation here: do we even need `mir::SetOperation::Union`? If we algebrize in such a way that all unions become `UnionAll`s in the `mir`, can we do away with `mir`'s representation of distinct `Union`?

mongosql/src/algebrizer/test.rs
```diff
@@ -7454,20 +7454,256 @@ mod limit_or_offset_clause {
...
+                            db: "test".into(),
+                            collection: "bar".into(),
+                            cache: SchemaCache::new()
+                        })),
+                        expression: map! {
+                            (DatasourceName::Named("bar".into()), 0u16).into() => mir::Expression::Reference((DatasourceName::Named("bar".into()), 0u16).into())
+                        },
+                        is_add_fields: false,
+                        cache: SchemaCache::new()
+                    })),
```
mattChiaravalloti, line 7498:
```suggestion
                    left: Box::new(mir_source_foo()),
                    right: Box::new(mir_source_bar()),
```
We could reuse some of the global helpers here to reduce the size of these tests! We could even go a step further and simplify these two helpers into one that is parametrized by collection name. Up to you if you want to do that kind of refactoring here or not.

mongosql/src/algebrizer/test.rs
```diff
@@ -7454,20 +7454,256 @@ mod limit_or_offset_clause {
...
+                    cache: SchemaCache::new()
+                })),
+                keys: vec![
+                    mir::OptionallyAliasedExpr::Aliased(mir::AliasedExpr {
+                        alias: "__groupKey0".to_string(),
+                        expr: mir::Expression::Reference(
+                            (DatasourceName::Named("bar".to_string()), 0u16).into()
+                        )
```
mattChiaravalloti, line 7506:
```suggestion
                        expr: mir::Expression::Reference(("bar", 0u16).into())
```
We have a `impl<T: Into<Key>> From<T> for ReferenceExpr` which utilizes `impl<D, S> From<(D, S)> for Key where D: Into<DatasourceName>, S: Into<u16>` which utilizes `impl<S> From<S> for DatasourceName where S: Into<String>`.

All of that to say, we ultimately allow converting `(&str, u16)` tuples into `ReferenceExpr`s directly! So you could remove the `DatasourceName::Named("...".to_string())` wherever you create an `Expression::Reference` to help cut down on the size of these tests!

mongosql/src/algebrizer/test.rs
```diff
@@ -7454,20 +7454,256 @@ mod limit_or_offset_clause {
...
+                aggregations: vec![],
+                cache: SchemaCache::new(),
+                scope: 0
+            })),
+            expression: map! {
+                (DatasourceName::Named("bar".to_string()), 0u16).into() => mir::Expression::FieldAccess(mir::FieldAccess {
+                    expr: Box::new(mir::Expression::Reference((DatasourceName::Bottom, 0u16).into())),
+                    field: "__groupKey0".to_string(),
+                    is_nullable: true
+                }),
```
mattChiaravalloti, line 7524:
```suggestion
                ("bar", 0u16).into() => crate::util::mir_field_access("__bot__", "__groupKey0", true),
```
We have some more test helpers in `crate/util/mod.rs` that could again help reduce test size.

mattChiaravalloti (changes requested):
This looks fantastic. The implementation is super slick! I left some comments directing you to some test helpers we have hidden away in a util file, as well as a couple suggestions for spec testing coverage and organization.

pmeredit (approved):
LGTM!

mongosql/src/algebrizer/test.rs
pmeredit: I think a good skunkworks project would be adding serde for mir so that we can make these tests less hideous :joy:

mattChiaravalloti: That would be wonderful 🙏

tests/spec_tests/query_tests/union.yml
bucaojit: Hm, there's a UNION distinct on the right side to show that it is executing the `UNION ALL` on the left first to duplicate `foo` and then the UNION distinct will deduplicate it and so a:1 and a:2 only appears once.  Double checking if you missed the second UNION?  Otherwise sure I can move it back here in `union.yml`.

tests/spec_tests/query_tests/distinct.yml
bucaojit: That's a good point, I had the mixed order test but not the same order `a,b`. I added a test to cover it.

mongosql/src/algebrizer/test.rs
bucaojit: Sounds good.  I made this change and added a helper function to hopefully make things easier to follow.

mongosql/src/algebrizer/test.rs
bucaojit: I added a parameterized helper function here.

mattChiaravalloti (approved):

PR #43, opened 2025-03-06
Sql 2628

bucaojit:
Updated the sections related to DISTINCT based on our changes for SELECT, COUNT, UNION.  

Also, check the `Documentation Changes Summary` that was added to change the user facing documentation on [SQL-2628.](https://jira.mongodb.org/browse/SQL-2628)

### DIFF
--- a/semantics.md
+++ b/semantics.md
@@ -44,8 +44,6 @@
     - [Collations](#collations)
     - [ORDER BY arbitrary expressions](#order-by-arbitrary-expressions)
     - [Supporting Non-Document BSON Values In Query Results](#supporting-non-document-bson-values-in-query-results)
-    - [SELECT DISTINCT](#select-distinct)
-    - [UNION DISTINCT](#union-distinct)
     - [USING CLAUSE](#using-clause)
     - [Unify arrays and subqueries](#unify-arrays-and-subqueries)
   - [Appendix](#appendix)
@@ -924,6 +922,14 @@ SELECT \* will return all binding tuples from the input stream
 unmodified. Other expressions may not be present in the select list
 alongside \*.
 
+#### Semantics of SELECT DISTINCT
+
+The DISTINCT keyword specifies that duplicate rows in the result set will be removed.
+
+Document and array comparison for DISTINCT follows MongoDB's equality semantics, where:
+- Document comparison includes field order
+- Array comparison requires equal length and equal elements in the same order
+
 <div id="select-grammar" />
 
 #### Grammar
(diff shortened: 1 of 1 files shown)


semantics.md
```diff
@@ -1663,11 +1669,14 @@ group as determined by the group key value.
...
-  values for which the expression does not result in NULL or
-  MISSING.
+- COUNT - Counts the number of elements. 
+  - COUNT(\*) counts all values unconditionally.
+  - COUNT([\<expression\>](#expressions)) counts all
+    values for which the expression does not result in NULL or
+    MISSING.
+  - COUNT(DISTINCT [\<expression\>](#expressions)) conditionally counts distinct values of the expression.
+  - COUNT(DISTINCT \*) unconditionally counts the number of distinct documents in the result set.
+  - COUNT(DISTINCT col1, col2, ...) counts distinct combinations of the specified columns.
```
mattChiaravalloti, line 1679: We should list the non-distinct version of this too.

Also could you note that `COUNT(DISTINCT? col1, col2, ..., coln)` is syntactically equivalent to `COUNT(DISTINCT? {'_1': col1, '_2': col2, ..., '_n': coln})`, but it is up to implementations whether or not to implement that as a syntactic rewrite. (That last bit is relevant since I did not include the rewrite as part of the spec tests! It isn't, strictly speaking, part of the spec to do the rewrite -- we just chose to implement it that way for convenience!)

semantics.md
```diff
@@ -2514,8 +2525,10 @@ deviates from MongoDB\'s comparison semantics.
 
 The binary comparison operators \<, \<=, \<\>, =, \>, and \>= specify
 less than, less than or equals, not equals, equals, greater than, and
-greater than or equals, respectively. MongoSQL does not support
-comparison operations on structured data (documents and arrays).
+greater than or equals, respectively. Document comparison follows MongoDB's 
+equality semantics, requiring exact field name and value matches, with 
+field order being significant. Array comparison requires equal length 
+and equal elements in the same order.
```
mattChiaravalloti, line 2531: In the Rejected Alternatives for this section (roughly lines 2750 - 2755) we note that "For now, we rejected comparisons for structured data (documents and arrays)". Can you update that section to say "_Initially_, we rejected comparisons for structured data ..." (and any other little bits of clean up in that paragraph to indicate we previously rejected this but no longer do).

semantics.md
mattChiaravalloti, line 1: The "Document Key Ordering Semantics" section could use some touching up since it discusses the disallowance of document comparisons at length. We may want to update it to say something like "Although MongoSQL does not provide any guarantees about the ordering of keys in documents, it does support document comparison. Literal documents are treated as ordered and document comparison follows MongoDB's document comparison rules."

The second paragraph of that section could possibly be fully deleted.

semantics.md
bucaojit: I added the non-distinct version and a note about the syntactic rewrite.

semantics.md
bucaojit: Oh good catch, I've updated it.

semantics.md
bucaojit: Okay, I added the part you suggested and removed the second paragraph since it is now outdated.

mattChiaravalloti (approved):

EthanHardyMongo (approved):
LGTM!

PR #44, opened 2025-03-06
SQL-2684: Add system variable support

pmeredit:
This does not include the redact variables since those are specific to the redact stage (redact will just have to bind those variables).

### DIFF
--- a/agg-ast/schema_derivation/src/schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/schema_derivation.rs
@@ -1253,7 +1253,6 @@ impl DeriveSchema for TaggedOperator {
                 remove_field(&mut input_schema, vec![u.field.clone()]);
                 Ok(input_schema)
             }
-            TaggedOperator::Accumulator(_) => todo!(),
             TaggedOperator::Bottom(b) => b.output.derive_schema(state),
             TaggedOperator::BottomN(b) => {
                 Ok(Schema::Array(Box::new(b.output.derive_schema(state)?)))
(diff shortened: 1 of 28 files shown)


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -743,7 +743,13 @@ impl DeriveSchema for Expression {
             Expression::Literal(ref l) => derive_schema_for_literal(l),
             Expression::Ref(Ref::FieldRef(f)) => {
                 let path = f.split(".").map(|s| s.to_string()).collect::<Vec<String>>();
-                let schema = get_schema_for_path_mut(&mut state.result_set_schema, path);
+                // If the user has rebound the CURRENT variable, we should use that schema instead of the result set schema to find any
+                // path.
```
mattChiaravalloti, line 747:
```suggestion
                // If the user has rebound the CURRENT variable, we should use that schema instead of the result set schema to find any
                // path. The reason we do this is because a field reference `$<field>` is equivalent to `$$CURRENT.<field>.
```
First of all, wow, great catch here 🔥 .

Feel free to change the wording. I think we should explain why this is relevant since it is not immediately obvious. Most people maintaining this codebase (well, most people in general), probably won't realize that all field references are implicitly `$$CURRENT`-prefixed variables references!

agg-ast/schema_derivation/src/schema_derivation_tests/expression.rs
```diff
@@ -122,11 +122,66 @@ mod literal {
...
+                    })
+                },
+                required: set!["foo".to_string()],
+                ..Default::default()
+            })
+        }
+    );
+}
+
+mod variable_ref {
```
mattChiaravalloti, line 180: Looks like we're missing a test for just `$$CURRENT` (being equivalent to `$$ROOT`)

mattChiaravalloti (changes requested):
Looks good. Marking as changes requested since I think we need one more unit test, but then we'll be good 👍

agg-ast/schema_derivation/src/schema_derivation.rs
pmeredit: Sadly I can't just commit suggestions since they won't be signed. Really annoying.

mattChiaravalloti: Can you still push your own commit with an updated comment here?

mattChiaravalloti (approved):
lgtm 👍

jchemburkar (approved):

PR #46, opened 2025-03-08
SQL-2667: Ensure ordered tests order on all items in the select list

EthanHardyMongo:
This PR updates the `ORDER BY` tests to never rely on implicit ordering. Additionally, I removed the `missing sorts before null` test and updated the spec because MongoDB doesn't work this way anymore.

### DIFF
--- a/tests/spec_tests/query_tests/order_by.yml
+++ b/tests/spec_tests/query_tests/order_by.yml
@@ -8,10 +8,6 @@ catalog_data:
       - { '_id': 0, 'a': 3, 'b': { 'c': 1 } }
       - { '_id': 1, 'a': 2, 'b': { 'c': 3 } }
       - { '_id': 2, 'a': 1, 'b': { 'c': 2 } }
-    nullAndMissing:
-      - { '_id': 0, 'a': null }
-      - { '_id': 1 }
-      - { '_id': 2, 'a': null }
     nullAndNonNull:
       - { '_id': 0, 'a': null }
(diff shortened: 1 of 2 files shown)


tests/spec_tests/query_tests/order_by.yml
```diff
@@ -170,7 +170,7 @@ tests:
 
   - description: missing sorts before null
     current_db: mydb
-    query: "SELECT * FROM nullAndMissing AS nullAndMissing ORDER BY a ASC"
+    query: "SELECT * FROM nullAndMissing AS nullAndMissing ORDER BY a ASC, _id ASC"
     ordered: true
     result:
       - { 'nullAndMissing': { '_id': 0, 'a': null } }
```
EthanHardyMongo, line 176: The description for this test is `missing sorts before null`; however, that's not what the result is.

EthanHardyMongo: This test is now failing with the following error:

```
[2025/03/07 19:34:05.826] thread 'generated::spec_tests_query_tests_order_by::missing_sorts_before_null' panicked at e2e-tests/src/generated/spec_tests_query_tests_order_by.rs:619:49:
[2025/03/07 19:34:05.826] called `Result::unwrap()` on an `Err` value: Translation(Algebrize(FieldNotFound("_id", Some(["a"]), OrderBy, 0)))
```

mattChiaravalloti (approved):

PR #48, opened 2025-03-10
SQL-2686:  Implement UNWIND rewrite to satisfy SQL Ergonomics Improvement Document

pmeredit:
fixing the fuzz tester took a bit.

### DIFF
--- a/mongosql/src/ast/rewrites/extended_unwind_rewrite.rs
+++ b/mongosql/src/ast/rewrites/extended_unwind_rewrite.rs
@@ -1,0 +1,225 @@
+use crate::ast::{
+    self,
+    definitions::{
+        Datasource, Expression, ExtendedUnwindOption, ExtendedUnwindSource, UnwindOption,
+        UnwindPathPart, UnwindPathPartOption, UnwindSource,
+    },
+    rewrites::{Error, Pass, Result},
+    visitor::Visitor,
+    SubpathExpr,
+};
+
+pub struct ExtendedUnwindRewritePass;
+
+impl Pass for ExtendedUnwindRewritePass {
+    fn apply(&self, query: ast::Query) -> Result<ast::Query> {
+        let mut visitor = ExtendedUnwindRewriteVisitor::default();
+        let res = query.walk(&mut visitor);
+        if let Some(error) = visitor.error {
+            return Err(error);
+        }
+        Ok(res)
+    }
+}
+
+/// The visitor that performs the rewrites for the `ExtendedUnwindRewriteRewritePass`.
+#[derive(Default)]
+struct ExtendedUnwindRewriteVisitor {
+    error: Option<Error>,
+}
+
+impl Visitor for ExtendedUnwindRewriteVisitor {
+    fn visit_datasource(&mut self, data_source: ast::Datasource) -> ast::Datasource {
+        match data_source {
+            ast::Datasource::ExtendedUnwind(ExtendedUnwindSource {
+                datasource: source,
+                options,
+            }) => {
+                let (mut paths, mut global_index, mut global_outer) = (None, None, false);
+                for option in options {
+                    match option {
+                        ExtendedUnwindOption::Paths(path) => {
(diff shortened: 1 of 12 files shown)


terakilobyte (approved):
LGTM, nice work. We should make another ticket to capture some of this in our query tests.

pmeredit:
@terakilobyte I'll file the ticket

mongosql/src/ast/rewrites/test.rs
```diff
@@ -900,6 +900,79 @@ mod optional_parameters {
...
+        );
+
+        test_rewrite!(
+            unwind_local_outer_overwrites_global_outer,
+            pass = ExtendedUnwindRewritePass,
+            expected = Ok(
+                "SELECT a.b.c AS c, a_ix, b_ix, b.d AS d FROM UNWIND(UNWIND(UNWIND(foo WITH INDEX => FOOO, OUTER => true, PATH => a) \
+                   WITH OUTER => false, INDEX => a_b_ix, PATH => a.b) WITH INDEX => b_ix, OUTER => true, PATH => b)"),
+            input = "SELECT a.b.c AS c, a_ix, b_ix, b.d as d FROM UNWIND(foo WITH PATHS => (a[INDEX=>FOOO].b[OUTER=>false], b[]), INDEX => ix, OUTER => true)",
+        );
```
bucaojit, line 973: How about we add a test to confirm the backward compatibility logic where the last part path is missing options (if we don't have one already).
Such as:
```
        test_rewrite!(
            unwind_missing_options_adds_unwind,
            pass = ExtendedUnwindRewritePass,
            expected =
                Ok("SELECT a.b.c AS c FROM (SELECT * FROM UNWIND(UNWIND(foo WITH PATH => a) WITH PATH => a.b)) AS zaz"),
            input = "SELECT a.b.c AS c FROM (SELECT * FROM UNWIND(foo WITH PATHS => (a[].b))) AS zaz",
        );
```

mongosql/src/ast/rewrites/extended_unwind_rewrite.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+
+    if !found_index && global_index.is_some() {
+        ret.push(UnwindOption::Index(format!(
+            "{}_{}",
+            index_prefix,
+            global_index.unwrap()
+        )));
+    }
```
bucaojit, line 147:
```suggestion
    if !found_index {
        if let Some(gi) = &global_index {
            ret.push(UnwindOption::Index(format!("{}_{}", index_prefix, gi)));
        }
    }
```
nit: this could help avoid having the `unwrap()` call

mongosql/src/ast/rewrites/extended_unwind_rewrite.rs
```diff
@@ -0,0 +1,245 @@
...
+            );
+            ret = Datasource::Unwind(UnwindSource {
+                datasource: Box::new(ret),
+                options,
+            });
+        }
+    }
+    // for backward compatibility, if the last path part has no options, we need to unwind it
+    // anyway. Even if the path part was specified as part[], the options will still consist of
+    // one empty vector, rather than being and empty vector, so we do not insert an extra unwind.
```
bucaojit, line 230:
```suggestion
    // one empty vector, rather than being an empty vector, so we do not insert an extra unwind.
```

bucaojit (approved):
LGTM! 
Nice work implementing the UNWIND improvements.  The tests are thorough and helped confirm the logic is correct 👍

mongosql/src/ast/rewrites/test.rs
pmeredit: thought I already had one, thanks!

mongosql/src/ast/rewrites/extended_unwind_rewrite.rs
pmeredit: Surprised clippy didn't complain

PR #49, opened 2025-03-10
SQL-2687: Implement WITH rewrite to satisfy SQL Ergonomics Improvement Document

pmeredit:


terakilobyte (approved):
LGTM. Sorry it took so long to review, it took me a while to write a bunch of local tests around this. Nice work!

mongosql/src/ast/rewrites/mod.rs
```diff
@@ -86,6 +88,7 @@ pub fn rewrite_query(query: ast::Query) -> Result<ast::Query> {
         &OptionalParameterRewritePass,
         &NotComparisonRewritePass,
         &ScalarFunctionsRewritePass,
+        &WithQueryRewritePass,
```
pmeredit, line 91: Note, we actually want to do WithQueryRewrite last because it can introduce duplicate queries, simple example:

```
WITH foo AS (...),
(SELECT * FROM foo JOIN foo)
```

If there are rewrites that apply in `...`, then they will be applied twice if we don't do WithQueryRewrite last.

nbagnard (approved):
LGTM!

pmeredit:
signature nonsense

PR #50, opened 2025-03-11
SQL-2688: Disallow order by missing cols for SELECT DISTINCT

bucaojit:
This resolves an issue where schema checking was failing when order by missing columns option is true and the query has `SELECT DISTINCT` and `ORDER BY`. 
It follows the behavior of Postgres that order by expressions must appear in the select list. 
Added a query test to confirm the behavior. 
Also added a fix in the mongosql-cli, for it to set the expected default of `allow_order_by_missing_columns = true`.

### DIFF
--- a/mongosql-cli/src/main.rs
+++ b/mongosql-cli/src/main.rs
@@ -59,8 +59,10 @@ fn main() -> Result<(), CliError> {
     let query = args.query;
     let namespaces = mongosql::get_namespaces(current_db.as_str(), query.as_str())?;
     let catalog = get_schema_catalog(uri.as_str(), current_db.as_str(), namespaces)?;
-    let mut options = mongosql::options::SqlOptions::default();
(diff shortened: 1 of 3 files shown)


pmeredit:
> This resolves an issue where schema checking was failing when order by missing columns option is true and the query has `SELECT DISTINCT` and `ORDER BY`. It follows the behavior of Postgres that order by expressions must appear in the select list. Added a query test to confirm the behavior. Also added a fix in the mongosql-cli, for it to set the expected default of `allow_order_by_missing_columns = true`.

To be clear, must be in the select list when SELECT DISTINCT is specified. This actually makes sense because it isn't clear what the value of something not in the select list would be when DISTINCT is used.

pmeredit (approved):
Looks great!

jchemburkar (approved):

PR #55, opened 2025-03-14
SQL-2675: add tests for query correctness tests

jchemburkar:


agg-ast/ast/src/definitions.rs
```diff
@@ -681,7 +681,7 @@ pub enum UnionWith {
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct UnionWithPipeline {
-    pub collection: String,
+    pub coll: String,
```
jchemburkar, line 684: 🚗 due to pipeline not deserializing

terakilobyte (approved):

pmeredit:
@jchemburkar I don't think git is storing the entire yaml files because at least tests_1 just seems to end for me on line 109405 in the middle of the catalog. I think we might need to split these up in order to store them in git?

pmeredit:
> @jchemburkar I don't think git is storing the entire yaml files because at least tests_1 just seems to end for me on line 109405 in the middle of the catalog. I think we might need to split these up in order to store them in git?

tests_2 looks fine, but isn't parsing with the python yaml parser

terakilobyte:
@pmeredit opening them locally in vim, they both look complete and format for me?

pmeredit (approved):
LGTM, we can split the files in another PR

pmeredit:
> @jchemburkar I don't think git is storing the entire yaml files because at least tests_1 just seems to end for me on line 109405 in the middle of the catalog. I think we might need to split these up in order to store them in git?

This was the middle of a result set schema, not the catalog, so it was indented farther than I expected

jchemburkar:
@pmeredit I just split them up here, and verified they still deserialize. I got most if not all of the weird line breaks too, I think the formatting looks better

terakilobyte (approved):

PR #56, opened 2025-03-14
SQL-2668: Support Arrays in get_schema_for_path_mut and get_or_create_schema_for_path_mut

pmeredit:
This is fun. It really simplified to use tail recursion in stead of a loop. No more &** etc for the most part. I just couldn't appease the borrow checker without doing this, and it also makes the path iteration easier to work with since we only advance the path when it's _not_ an array. 

I didn't bother to test single arrays because double arrays, I think, so that it's working.


agg-ast/schema_derivation/src/test.rs
```diff
@@ -77,6 +77,28 @@ mod get_schema_for_path {
         path = vec!["a".to_string(), "b".to_string(), "c".to_string()]
     );
 
+    test_schema_for_path!(
```
jchemburkar, line 80: I'm rocking with these tests. I think it might also be worth adding these in practice, perhaps not everywhere, but also in at least one likely operator. Perhaps we can add an $unwind test that relies on this?

jchemburkar (changes requested):
This turned out pretty sick. I have mostly housekeeping ideas to try to document this for future me 😂

agg-ast/schema_derivation/src/test.rs
```diff
@@ -274,4 +324,41 @@ mod get_schema_for_path {
         }),
         path = vec!["a".to_string(), "b".to_string()]
     );
+
+    test_get_or_create_schema_for_path!(
```
mattChiaravalloti, line 328: Test idea / behavior question: should the following work?
```
test_get_or_create_schema_for_path!(
    get_or_create_multiple_possible_paths_with_arrays,
    expected = Some(&mut Schema::AnyOf(set!(Schema::Atomic(Atomic::Integer), Schema::Atomic(Atomic::String)))),
    output = Schema::Document(Document {
        keys: map! {
            "a".to_string() => Schema::AnyOf(set!(
                Schema::Array(Box::new(Schema::Document(Document {
                    keys: map! {
                        "b".to_string() => Schema::Atomic(Atomic::Integer),
                    },
                    required: BTreeSet::new(),
                    additional_properties: false,
                    ..Default::default()
                }))),
                Schema::Array(Box::new(Schema::Array(Box::new(Schema::Document(Document {
                    keys: map! {
                        "b".to_string() => Schema::Atomic(Atomic::String),
                    },
                    required: BTreeSet::new(),
                    additional_properties: false,
                    ..Default::default()
                }))))),
            )),
        },
        required: BTreeSet::new(),
        additional_properties: false,
        ..Default::default()
    }),
    input = <same as output>,
    path = vec!["a".to_string(), "b".to_string()]
);
```

I think this is written correctly... my point is: what if there is an `AnyOf` with two valid alternative paths? In this case, we're saying `a` is a either (1) an array of documents with `b` as an `int`, or (2) an array of arrays of documents with `b` as a `string`.

If this is a valid case, can we add a test to demonstrate it?

mattChiaravalloti (changes requested):
Let me know if my suggested test case is just something we're not interested in implementing at this time!

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -118,16 +119,33 @@ pub(crate) fn get_schema_for_path_mut(
...
+}
+
+#[tailcall]
+fn get_schema_for_path_mut_aux(
+    schema: &mut Schema,
+    mut path: Vec<String>,
+    current_field: Option<String>,
+    field_index: usize,
+) -> Option<&mut Schema> {
+    if path.len() == field_index {
```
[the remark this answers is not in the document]
pmeredit: correct

agg-ast/schema_derivation/src/test.rs
pmeredit: That's not a valid case, it only looks at arrays if there is no document schema in an anyOf. That would require mapping across multiple paths which creates an exponential blow up possibly. If we ensure schemas are always minimized, it's sort of safe, but this is very difficult while also pretty rare. I actually have a comment in there that we can revisit this idea later if it comes up a lot.

mattChiaravalloti: There isn't a Document in the `AnyOf`, at least not at the top-level. This test would trigger the "We only investigate Arrays if the AnyOf contains no Document schemas" case:
```
if !schemas.iter().any(|s| matches!(s, &Schema::Document(_))) {
// <-- evaluates to true since the 2 top-level schema types in the AnyOf are Array
                let schemas = std::mem::take(schemas);
                if let Some(a) = schemas.into_iter().find_map(|s| { // <-- evaluates to the first Array schema (Array(Document))
                    if let Schema::Array(a) = s {
                        Some(a)
                    } else {
                        None
                    }
                }) {
                    *schema = Schema::Array(a);
                    return get_or_create_schema_for_path_mut_aux(
                        schema,
                        path,
                        Some(field),
                        field_index,
                    );
                }
```

I see what you're saying though: this is not a case we plan to support fully.

jchemburkar (approved):
lgtm!

PR #60, opened 2025-03-21
Make mongosql-cli pipelines copy-pasteable

mattChiaravalloti:
This PR makes `mongosql-cli` pipelines copy-pasteable. I know the tool already runs the pipelines for us, but if I want to re-run the pipelines or make minor edits and run the pipelines, I have to copy them into an editor and manually add commas between the stages. This update puts the commas in the output for us.

I'm just going to wait on 1 approval for this one since it's so small.


terakilobyte (approved):
lgtm, this will make copying and using elsewhere a bit easier and faster.

PR #62, opened 2025-03-21
SQL-2537: Remove todos

pmeredit:
Noticed our Document schema intersection was actually wrong in that requiring a key is smaller than not.

### DIFF
--- a/agg-ast/schema_derivation/src/match_schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/match_schema_derivation.rs
@@ -2149,7 +2149,7 @@ impl MatchConstrainSchema for Expression {
                 TaggedOperator::DateDiff(d) => match_derive_date_diff(d, state)?,
                 TaggedOperator::DateSubtract(d) => match_derive_date_subtract(d, state)?,
                 TaggedOperator::DateTrunc(d) => match_derive_date_trunc(d, state)?,
-                _ => todo!(),
+                _ => {}
             },
 
             Expression::UntaggedOperator(u) => match u.op {
@@ -2248,7 +2248,7 @@ impl MatchConstrainSchema for Expression {
                 UntaggedOperatorName::IndexOfCP | UntaggedOperatorName::IndexOfBytes => {
                     match_derive_index_of(u, state)?
                 }
-                _ => todo!(),
+                _ => {}
             },
             _ => {}
         }
--- a/agg-ast/schema_derivation/src/match_schema_derivation.rs
+++ b/agg-ast/schema_derivation/src/match_schema_derivation.rs
@@ -7,12 +7,13 @@ use crate::{
 use agg_ast::definitions::{
     DateAdd, DateDiff, DateExpression, DateFromParts, DateFromString, DateSubtract, DateToParts,
     DateToString, DateTrunc, Expression, Filter, Let, Map, MatchBinaryOp, MatchExpr,
-    MatchExpression, MatchField, MatchLogical, MatchNotExpression, MatchStage, Median, NArrayOp,
-    Reduce, Ref, RegexAggExpression, Replace, SortArray, Switch, TaggedOperator, Trim,
+    MatchExpression, MatchField, MatchLogical, MatchMisc, MatchNotExpression, MatchStage, Median,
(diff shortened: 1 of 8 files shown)


agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -241,6 +242,31 @@ impl MatchConstrainSchema for MatchField {
...
+                    return Err(Error::InvalidJsonSchema);
+                };
+                let schema: Schema = JsonSchema::from_document(d)
+                    .map_err(|_| Error::InvalidJsonSchema)?
+                    .try_into()
+                    .map_err(|_| Error::InvalidJsonSchema)?;
+                state.result_set_schema = state.result_set_schema.intersection(&schema);
+                Ok(())
+            }
+            MatchMisc::Text(_) => Ok(()),
```
jchemburkar, line 264: worth maybe leaving a comment just specifying these are no ops

agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -2123,7 +2149,7 @@ impl MatchConstrainSchema for Expression {
                 TaggedOperator::DateDiff(d) => match_derive_date_diff(d, state)?,
                 TaggedOperator::DateSubtract(d) => match_derive_date_subtract(d, state)?,
                 TaggedOperator::DateTrunc(d) => match_derive_date_trunc(d, state)?,
-                _ => todo!(),
+                _ => {}
```
jchemburkar, line 2152: did you consider throwing an error here? I suppose for match its probably better default behavior to do nothing then error, if we eg add a new op but forget to add it to match, but curious if you considered it any deeper.

agg-ast/schema_derivation/src/negative_normalize.rs
```diff
@@ -228,30 +260,108 @@ impl NegativeNormalize<Expression> for Expression {
...
+                    | UntaggedOperatorName::Last | UntaggedOperatorName::Literal
+                    | UntaggedOperatorName::Max | UntaggedOperatorName::Min => {
                         let null_check = wrap_in_null_or_missing_check!(self.clone());
                         let zero_check = wrap_in_zero_check!(self.clone());
                         let false_check = wrap_in_false_check!(self.clone());
                         (UntaggedOperatorName::Or, vec![null_check, zero_check, false_check])
                     }
+                    // the following can only evaluate to true or false
+                    | UntaggedOperatorName::MQLBetween => {
+                        let false_check = wrap_in_false_check!(self.clone());
```
jchemburkar, line 328: nit but this and the following null check only can probably just happen on one line, that's how we do it in TaggedOperator for types that only have one check

agg-ast/schema_derivation/src/negative_normalize.rs
```diff
@@ -228,30 +260,108 @@ impl NegativeNormalize<Expression> for Expression {
...
+                    | UntaggedOperatorName::SQLIs | UntaggedOperatorName::SQLLt
+                    | UntaggedOperatorName::SQLLte | UntaggedOperatorName::SQLNe
+                    | UntaggedOperatorName::SQLNot => {
+                        let null_check = wrap_in_null_or_missing_check!(self.clone());
+                        let false_check = wrap_in_false_check!(self.clone());
+                        (UntaggedOperatorName::Or, vec![null_check, false_check])
+                    }
+                    // the following operators can be expensive and evaluate to null, 0, or false,
+                    // so we bind them to a let.
+                    | UntaggedOperatorName::Cond => {
```
jchemburkar, line 350: 🔥

jchemburkar (approved):
a couple cleanup nits, but overall looks great 😄

terakilobyte (approved):
LGTM, nothing to add on top of @jchemburkar comments

agg-ast/schema_derivation/src/match_schema_derivation.rs
pmeredit: Yes, the idea is that we are conservative here.

terakilobyte (approved):

PR #64, opened 2025-03-24
SQL-2676: add schema derivation tests for spec query tests

jchemburkar:
pipelines generated by using the catalog schema and calling translate_sql

result set schemas generated by loading the result fields into a local mongod and using the schema manager

no need for catalog_data, since we wont be loading / using data, just the result set schema.


terakilobyte (approved):
Lots of tests here

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1307,6 +1307,44 @@ mod union {
...
+                "out".to_string() => Schema::Atomic(Atomic::Decimal),
+            },
+            required: set!("food".to_string(), "out".to_string()),
+            ..Default::default()
+        })
+    );
+    test_derive_stage_schema!(
+        union_not_enough_args,
+        expected = Err(crate::Error::NotEnoughArguments("$unionWith".to_string())),
+        input = r#"{"$unionWith": {}}"#,
```
mattChiaravalloti, line 1338: So did you run into this as an actual pipeline or is this just preventative in case we encounter this in the future? Does mongodb not reject views that contain this type of stage?

tests/schema_derivation_tests/query_spec_tests/alias.yml
```diff
@@ -0,0 +1,284 @@
...
+        "properties": {
+          "a": {
+            "bsonType": "int"
+          }
+        }, 
+        "bsonType": "object"
+      }
+    }, 
+    "bsonType": "object"
+  }
```
mattChiaravalloti, line 284:
```suggestion
  }

```
[completely optional nit] Let's add newlines to any test files that are missing them at the end. A scroll through this change set makes it look like only this file is missing it but I might have missed some (admittedly, I did not inspect every file closely 🙈 )

mattChiaravalloti (approved):
lgtm 👍 one small suggestion that might impact multiple files before merging, but otherwise this is perfect. Glad it worked out so well.

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
jchemburkar: you are correct that we should never run into this, and pipeline correctness is out of scope for this project. this won't actually run on server! We should never hit this error. I keep waffling on this case by case on if we should simply panic, but given we have an existing descriptive error for this case, it seemed free to me to point out the user needs to add args

PR #65, opened 2025-03-27
SQL-2720: add optional current_collection field to ResultSetState

jchemburkar:
only tagging one since its really small, but if you think it merits I'll add a second. Essentially, we need to capture the "from" collection that we're getting from either a translation or a view definition to know where to start.


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -71,6 +71,7 @@ pub(crate) struct ResultSetState<'a> {
     pub variables: BTreeMap<String, Schema>,
     pub result_set_schema: Schema,
     pub current_db: String,
+    pub current_collection: Option<String>,
```
jchemburkar, line 74: this is the only file worth reviewing. All the test updates are just adding this field as necessary

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -558,6 +572,7 @@ impl DeriveSchema for Stage {
                 variables,
                 result_set_schema: from_schema.clone(),
                 current_db: state.current_db.clone(),
+                current_collection: None,
```
jchemburkar, line 575: I used the convention that any internal calls to this function will set current collection to None, for now. And any external calls (i.e. schema manager) will use the collection from the view definition / the current_collection from the translate_sql response.

mattChiaravalloti: (No need to tag me as a reviewer after this comment, just taking a look since I saw the `@sql-engines-team` email). If the only time this is used is when calling `derive_schema_for_pipeline`, should we consider making it an argument to the function rather than a field of the `ResultSetState`? This could help reduce the number of places we need to make the change (just `derive_schema_for_pipeline` callsites, as opposed to all instances of `ResultSetState`).

jchemburkar: Yeah this is a good shout. I had some ideas of how this could abstractly be useful otherwise, but I think they either aren't real or aren't impacted by this switch. so I changed to match this suggestion.

pmeredit (approved):
Looks great!

PR #68, opened 2025-03-31
SQL-2725: handle missing variable refs appropriately

jchemburkar:
the update to $match: $reduce was (sort of) a driveby addressing the same core error message we were getting back from deserializing pipelines. This is the one array op we'll _need_ to handle this for -- for $filter, for example, we just investigate the input, but dont bother scanning the array, so $$this won't come up there!


nbagnard (approved):
LGTM!

pmeredit (approved):
LGTM!

PR #71, opened 2025-04-03
SQL-2531: Change SQL and MQL in type names to Sql and Mql to be consistent with rust

pmeredit:
Should be good now. I'm not worried about case in comments, it doesn't matter, but I wanted to make sure we don't accidentally change anything user facing.


pmeredit (commented):
Should be good now

mongosql/src/codegen/mod.rs
```diff
@@ -15,8 +15,8 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, Error, PartialEq, Eq)]
 pub enum Error {
-    #[error("cannot generate MQL for {0:?} operator")]
-    UnsupportedOperator(air::SQLOperator),
+    #[error("cannot generate Mql for {0:?} operator")]
```
mattChiaravalloti, line 18: This is an interesting one. Do customers know what "Mql" means? Should we say "cannot generate Aggregation Language for ..."? Do they know what "Aggregation Language" is? 🤷‍♀️ Idk, we can probably just leave this as-is.

mattChiaravalloti (approved):
lgtm 👍  there is really no need to make a change for the one comment I left. It's really just a musing on something that stood out to me as interesting, but it's ultimately inconsequential.

mongosql/src/codegen/mod.rs
pmeredit: This is a good point. This should never actually hit customers, to be fair, since we don't generate any operators that cannot generate. This would actually mean we had a code error (which is another thing haha)

terakilobyte (approved):

PR #72, opened 2025-04-04
SQL-2732: Allow Distinct queries to return select order

bucaojit:
Added SetQuantifier::Distinct to the pattern matching for getting select order. 

This is a fix for the issue where results were not being shown when running a DISTINCT query through DBeaver.

### DIFF
--- a/mongosql/src/lib.rs
+++ b/mongosql/src/lib.rs
@@ -151,13 +151,13 @@ pub fn get_namespaces(current_db: &str, sql: &str) -> Result<BTreeSet<Namespace>
 }
 
 // get_select_order uses pattern matching to parse the select body from the rewritten AST.
-// Only parse if it is a non-distinct SelectQuery, otherwise return none (we won't attempt reordering).
+// Parses both distinct and non-distinct SelectQuery
 pub fn get_select_order(ast: &ast::Query) -> Option<ast::SelectBody> {
     match ast {
         ast::Query::Select(ast::SelectQuery {
             select_clause:
                 ast::SelectClause {
-                    set_quantifier: ast::SetQuantifier::All,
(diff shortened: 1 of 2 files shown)


mongosql/src/lib.rs
```diff
@@ -151,13 +151,13 @@ pub fn get_namespaces(current_db: &str, sql: &str) -> Result<BTreeSet<Namespace>
...
 // get_select_order uses pattern matching to parse the select body from the rewritten AST.
-// Only parse if it is a non-distinct SelectQuery, otherwise return none (we won't attempt reordering).
+// Parses both distinct and non-distinct SelectQuery
 pub fn get_select_order(ast: &ast::Query) -> Option<ast::SelectBody> {
     match ast {
         ast::Query::Select(ast::SelectQuery {
             select_clause:
                 ast::SelectClause {
-                    set_quantifier: ast::SetQuantifier::All,
+                    set_quantifier: ast::SetQuantifier::All | ast::SetQuantifier::Distinct,
```
mattChiaravalloti, line 160:
```suggestion
                    set_quantifier: _,
```
Could/should we use a wildcard here to indicate we don't even actually care what the value is? I know we typically avoid wildcards so we get compiler warnings when new variants are added but (1) this is not a context that will produce such as warning and (2) I'm not sure there will ever be more variants for `SetQuantifier` 🤔 

Your call either way! This looks good to me 👍

mattChiaravalloti (approved):
lgtm 👍

mongosql/src/lib.rs
bucaojit: Yeah sure that sounds like a good suggestion I'll commit the suggestion.

jchemburkar (approved):

PR #73, opened 2025-04-07
SQL-2734: Split up algebrizer tests into separate files to make my nvim not so terribly slow

pmeredit:


nbagnard (approved):
LGTM!

Feel free to request a second pair of eyes though, because it is easy to miss a test that we was removed.
Did you check that the same number of tests are running as a baseline, spot check test?

PR #74, opened 2025-04-07
SQL-2733: handle missing in expression schema derivation

jchemburkar:
for field ref expressions, we were using the get_schema_for_path_mut func, which handles arrays and docs but cannot handle anyofs. tHis becomes an issue for, amongst other things, Missing. We promote missing (so a non required Int becomes an AnyOf(MISSING, INT), for example). 

There was no compiler friendly way to handle this in get_schema_for_path_mut, and just getting the schema (non-mut) is more forgiving / we can cheat in other ways (such as around Schema::Any). So I added a new helper and use that for field ref schema derivation.


agg-ast/schema_derivation/src/lib.rs
```diff
@@ -155,6 +155,49 @@ fn get_schema_for_path_mut_aux(
     }
 }
 
+/// Gets a mutable reference to a specific field or document path in the schema.
```
mattChiaravalloti, line 158: Outdated comment? This is the one that does not return a mutable ref

agg-ast/schema_derivation/src/schema_derivation_tests/expression.rs
```diff
@@ -213,6 +238,129 @@ mod field_ref {
...
+    // the input schema to this test is surely a bit hard to follow. The
+    // the below pipeline illustrates what type of cases we are covering:
+    // [
+    //  {$documents: [
+    //    {foo: [{a: {b: 1}}]},
+    //    {foo: [{a: [{b: 1}]}]},
+    //    {foo: {a: {b: 2}}},
+    //    {foo: {a: [{b: 2}]}},
+    //  ]},
+    //  {$project: {foo: \"$foo.a.b\"}}
```
mattChiaravalloti, line 304: Super helpful 🙏

mattChiaravalloti (approved):
lgtm 👏

agg-ast/schema_derivation/src/lib.rs
jchemburkar: yup just copy and pasted the original func 🤦

pmeredit (approved):
LGTM!

PR #78, opened 2025-04-21
SQL-2741: schema derivation fixes for server correctness tests

jchemburkar:
Please ignore the changes to the test files. Many were either skipped or rewritten to project relevant fields, greatly reducing the scope of the result set schema. A handful were also intentionally removed, because other tests were a superset of the stages/operators they covered (if curious why I'm happy to explain).


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -231,32 +231,35 @@ impl DeriveSchema for Stage {
...
                     FillOutput::Value(e) => {
                         let fill_schema = e.derive_schema(state)?;
-                        let path_schema = get_path_schema!();
-                        // Null is possible if the value expression can evaluate to null, but
-                        // missing will always be impossible. We subtract NULL so it can only be
-                        // reintroduced via the union of the fill_schema
-                        *path_schema = std::mem::take(path_schema)
-                            .subtract_nullish()
-                            .union(&fill_schema);
+                        match get_schema_for_path_mut(
```
jchemburkar, line 237: for $fill, we were correctly marking existing keys as required; but we were not inserting as required new keys.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -309,11 +312,12 @@ impl DeriveSchema for Stage {
                 .aggregations
                 .iter()
                 .map(|(k, e)| {
-                    let field_schema = e.derive_schema(state)?;
+                    let field_schema = e.derive_schema(state)?.upconvert_missing_to_null();
```
jchemburkar, line 315: there's a few upconvert missing to nulls sprinkled around this pr, a common mistake / fix needed for window or group funcs.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -408,31 +412,16 @@ impl DeriveSchema for Stage {
...
-                keys,
-                ..Default::default()
-            }))
+            let mut result_doc = state.result_set_schema.clone();
+            for (k, e) in set_windows.output.clone() {
+                let field_schema = e
+                    .window_func
+                    .derive_schema(state)?
+                    .upconvert_missing_to_null();
+                let path = k.split('.').map(|s| s.to_string()).collect::<Vec<String>>();
```
jchemburkar, line 421: This change just a. upconverts missing to null and b. ensures field name is being treated as a path, not a literal

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -455,47 +444,58 @@ impl DeriveSchema for Stage {
                 state.result_set_schema = Schema::simplify(&state.result_set_schema);
                 return Ok(state.result_set_schema.to_owned());
             }
+            let mut result_doc = Schema::Document(Document::empty());
```
jchemburkar, line 447: still work to be done ofc, but the crux of the $project changes here is rather than collecting key-schema pairs, we just create an empty document and insert the full paths as we go. I suppose we could collect path-schema pairs, but its the same either way 🤷

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1998,7 +1971,7 @@ impl DeriveSchema for UntaggedOperator {
             UntaggedOperatorName::ArrayElemAt => {
                 let input_schema = self.args[0].derive_schema(state)?;
                 match input_schema.clone() {
-                    Schema::Array(a) => Ok(a.as_ref().clone()),
+                    Schema::Array(a) => Ok(Schema::simplify(&a.as_ref().union(&Schema::Atomic(Atomic::Null)))),
```
jchemburkar, line 1974: null needed for out of bounds indexes!

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -2098,28 +2073,12 @@ impl DeriveSchema for UntaggedOperator {
...
-                let schema = self.args[0].derive_schema(state)?;
-                let schema = match schema {
-                    Schema::AnyOf(a) => {
-                        // Unsat should be impossible, since we should never see AnyOf(empty_set)
-                        let schema = a.iter().min().unwrap_or(&Schema::Unsat);
-                        numeric_filter(schema.clone(), a)
-                    }
-                    _ => schema,
-                };
+            UntaggedOperatorName::Max
```
jchemburkar, line 2076: max and min had greedy implementations...if you have an AnyOf, you can't just take the max value, because for example, $max: $foo is just going to return whatever foo is for a given row, even if its just MinKey.

There is probably something to be done here if args is an array, and a specific argument is strictly > than the rest, but that was more work than it was worth for now.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -2327,13 +2286,17 @@ impl DeriveSchema for UntaggedOperator {
...
             UntaggedOperatorName::First | UntaggedOperatorName::Last => {
-                let schema = self.args[0].derive_schema(state)?;
+                let schema = self.args[0].derive_schema(state)?.upconvert_missing_to_null();
                 Ok(match schema {
                     Schema::Array(a) => *a,
                     schema => schema
                 })
             }
             UntaggedOperatorName::Literal => self.args[0].derive_schema(state),
+            UntaggedOperatorName::Split => handle_null_satisfaction(
```
jchemburkar, line 2296: incorrectly returned just string as opposed to array(string)!

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1268,6 +1268,31 @@ mod project {
...
+    //     )),
+    //     input = r#"{"$project": {"foo.a": 1, "foo.b": 1}}"#,
+    //     ref_schema = Schema::Array(Box::new(Schema::Document(Document {
+    //         keys: map! {
+    //             "a".to_string() => Schema::Atomic(Atomic::String),
+    //             "b".to_string() => Schema::Atomic(Atomic::Integer),
+    //         },
+    //         ..Default::default()
+    //     })))
+    // );
```
terakilobyte, line 1295: Just a reminder about ignore for anything you might not get to right away. 

```rs
macro_rules! test_derive_stage_schema {
    // ref_schema and starting_schema are mutually exclusive. ref_schema should be used when only
    // one reference is needed, while starting_schema should be used when the schema needs multiple
    // fields.
    ($func_name:ident, expected = $expected:expr, input = $input:expr$(, starting_schema = $starting_schema:expr)?$(, ref_schema = $ref_schema:expr)?$(, variables = $variables:expr)?$(, ignore_reason = $ignore_reason:expr)?) => {
        $(#[ignore = $ignore_reason])?
        #[test]
        ...
```

terakilobyte (approved):
LGTM.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -455,47 +444,58 @@ impl DeriveSchema for Stage {
...
             // and then union them together to get the resulting schema.
-            let mut keys = project
-                .items
-                .iter()
-                .filter(|(_k, p)| !matches!(p, ProjectItem::Exclusion))
-                .map(|(k, p)| match p {
+            for (k, p) in project.items.clone() {
+                let path = k.split('.').map(|s| s.to_string()).collect::<Vec<String>>();
+                if path[0] == "_id" {
+                    include_id = false;
```
mattChiaravalloti, line 458: Can you expand on this in code comments? Why is this the case?

mattChiaravalloti (approved):
lgtm mod one more code comment, unless you think that comment can/should wait until the follow-up ticket which is fine with me!

PR #80, opened 2025-04-23
SQL-2789: make schema derivation more ergonomic for external callers

jchemburkar:
we had 5 different definitions of Namespace 😔

### DIFF
--- a/agg-ast/ast/src/definitions.rs
+++ b/agg-ast/ast/src/definitions.rs
@@ -2,6 +2,7 @@ use crate::custom_serde::{deserialize_mql_operator, serialize_mql_operator};
 use bson::Bson;
 use linked_hash_map::LinkedHashMap;
 use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
 
 // This module contains an aggregation pipeline syntax tree that implements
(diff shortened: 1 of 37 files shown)


agg-ast/ast/src/definitions.rs
```diff
@@ -532,10 +533,50 @@ pub struct SubqueryLookup {
     pub as_var: String,
 }
 
-#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize, Eq, PartialOrd, Ord)]
 pub struct Namespace {
-    pub db: String,
-    pub coll: String,
+    #[serde(alias = "db")]
```
jchemburkar, line 538: the benefit of aliasing here is that it allows us to read in aggregation pipelines as is ("db", "coll" for lookups) but it also allows us to serialize to "database" "collection" (preserving translation format) and deserialize again

terakilobyte (approved):
Code changes LGTM, thanks for cleaning this up!

nbagnard (approved):
LGTM!
Great idea to remove all the different struct definition and use a common one 👍

PR #81, opened 2025-04-24
SQL-2785: handle nested array fields in $project properly

jchemburkar:


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -429,11 +429,157 @@ impl DeriveSchema for Stage {
             Ok(result_doc)
         }
 
+        #[derive(Debug)]
+        struct ProjectPathNode {
+            pub children_nodes: LinkedHashMap<String, ProjectPathNode>,
```
terakilobyte, line 434: I know there's some gravity to using LinkedHashMap, but wondering if we actually need to here. I don't see any use of the linked list funcionality. Additionally and super super nitty, but it might be better to use `Option` to wrap to save a few bytes.

agg-ast/schema_derivation/src/lib.rs
```diff
@@ -418,9 +418,18 @@ pub(crate) fn insert_required_key_into_document(
...
+            Some(Schema::Document(d)) => {
+                d.keys.remove(field);
+                d.required.remove(field);
+            }
+            Some(Schema::Array(a)) => {
+                if let Schema::Document(d) = a.as_mut() {
+                    d.keys.remove(field);
+                    d.required.remove(field);
+                }
+            }
```
mattChiaravalloti, line 431:
```suggestion
            Some(Schema::Array(Schema::Document(d))) => {
                d.keys.remove(field);
                d.required.remove(field);
            }
```
Is this legal rust? Could be nice to condense this directly into the pattern as opposed to using an `if` inside the pattern branch.

tests/schema_derivation_tests/server_correctness_tests/correctness_test_2_3.yml
```diff
@@ -11,12 +10,10 @@ result_set_schema:
         "items": {
           "bsonType": "object",
           "properties": {
-            "_id": {
-              "bsonType": "array",
-              "items": {
```
mattChiaravalloti, line 16: Love to see it

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1347,6 +1503,84 @@ mod project {
...
+                    keys: map! {
+                        "a".to_string() => Schema::Atomic(Atomic::String),
+                    },
+                    ..Default::default()
+                })))
+            },
+            required: set!("foo".to_string()),
+            ..Default::default()
+        })),
+        input = r#"{"$project": {"foo.a": 0, "foo.a": 1}}"#,
```
mattChiaravalloti, line 1543: Reviewing tests before implementation so I may revisit this comment, but wow this test and the previous one surprise me. The order of the exclude and include matter? Whichever you specify last is the one that takes effect? MongoDB is so weird.

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
```diff
@@ -1268,31 +1268,187 @@ mod project {
...
-    //     ref_schema = Schema::Array(Box::new(Schema::Document(Document {
-    //         keys: map! {
-    //             "a".to_string() => Schema::Atomic(Atomic::String),
-    //             "b".to_string() => Schema::Atomic(Atomic::Integer),
-    //         },
-    //         ..Default::default()
-    //     })))
-    // );
+    test_derive_stage_schema!(
+        project_multiple_fields_array_of_docs,
```
mattChiaravalloti, line 1272: Can you also a add a test very similar to this one but with extra keys in the `ref_schema` that are not projected in the `input`. As in, what if the `ref_schema` had a `c` field and the `input` stayed the same as it is? The output should be the same as well in that case. We should have a test to assert this since it could be possible for us to mess up the implementation where we retain all fields.

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -429,11 +429,157 @@ impl DeriveSchema for Stage {
...
+        /// the parent schema keeps a reference to the schema for the current node from the
+        /// result set schema -- that is, we traverse the result set schema in parallel in
+        /// order to properly evaluate inclusions.
+        fn get_schema_for_project_node(
+            node: &ProjectPathNode,
+            parent_schema: Option<&Schema>,
+            state: &mut ResultSetState,
+        ) -> Schema {
+            if let Some(schema) = parent_schema {
+                // the former case is not likely, but if we have reached a terminal arm of the
```
mattChiaravalloti, line 500: It's "not likely" or it's not possible? Can you give an example of how we could end up in this state?

mattChiaravalloti: Oh wait, would is just be an empty `$project` stage?

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -429,11 +429,157 @@ impl DeriveSchema for Stage {
...
+                            insert_required_key_into_document(
+                                &mut path_document,
+                                child_schema,
+                                vec![child_field.clone()],
+                                true,
+                            );
+                        }
+                        schemas.insert(path_document);
+                    }
+                    _ => {}
```
mattChiaravalloti, line 564: Sorry I'm a bit out of date, but do we expand `Any` to an `AnyOf` before this? Is that relevant?

mattChiaravalloti (changes requested):
Looks excellent. Really nice job on this. I suggest an extra test and have some questions, but I don't think the implementation needs much updating, if any at all.

agg-ast/schema_derivation/src/lib.rs
jchemburkar: That would be legal, however, array types are wrapped in a `Box` which isn't possible to pattern match

agg-ast/schema_derivation/src/schema_derivation_tests/stage.rs
jchemburkar: it is pretty annoying! we actually just overwrite during deserialization _anyways_, so this should work out of the box. But it revealed the issue with not handling `remove_field` with arrays properly 😄 s/o @terakilobyte for the test suggestion

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: This is a good callout, this is correct, we need to do this. This also is making me realize we should add tests to handle additionalproperties correctly. Will add

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: Yeah that is about correct. I just removed it, and added a new error for empty project, which will deserialize but we shouldn't entertain

terakilobyte (approved):

PR #82, opened 2025-04-24
SQL-2793: Fix match null filtering optimization to properly set nullability of filtered exprs

mattChiaravalloti:
This PR updates the `match_null_filtering` optimization. The bug was actually not specific to `char_length` or `case`! It turns out the problem was actually that any expression that was "match_null_filtered" had all of the exprs in its tree set to `is_nullable: false` which is obviously inaccurate in the case of the query in the ticket. In fact, that's the case for plenty of expression trees -- anything that contains a literal null, or even a function that may return null such as `coalesce` or `nullif`, could still be nullable even after filtering field refs! Ultimately, the incorrect assumption of the `NullableSetter` visitor was that all expr nullability is dictated by field access expressions. That is not true.

This PR fixes that assumption. I added a unit test to the optimization to show the fix, as well as an e2e test demonstrating this particular query works now (it is technically representative of the general issue).


terakilobyte (commented):
Really nice find and fix here. Just a question about the expected results of the test.

tests/e2e_tests/queries.yml
```diff
@@ -891,3 +906,8 @@ tests:
...
+
+  - description: Nullable expression is treated as nullable inside an operator that requires SQL null semantics
+    current_db: test_db
+    query: "SELECT * FROM bar WHERE CHAR_LENGTH(CASE WHEN a < 5 THEN NULL WHEN a < 10 THEN '' END) <> 0"
+    result: []
```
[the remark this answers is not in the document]
mattChiaravalloti: There should not -- this filter could never be true! The result of the `CASE` expr is one of two values: `NULL` or `''` (note there is an implicit `ELSE NULL`). `CHAR_LENGTH(NULL)` is `NULL`, and `NULL <> 0` is `NULL`, and `WHERE NULL` is effectively `WHERE FALSE`; `CHAR_LENGTH('')` is `0`, and `0 <> 0` is `FALSE`, so that's also `WHERE FALSE`. It doesn't even really matter what the inputs are since this filter is always false!

mattChiaravalloti:
I want to look at this index failure in more detail. I will follow up when I have a fix for it.

mattChiaravalloti:
Oh you know what, that index usage test is quite unique -- a division _can_ be null if the divisor is `0` at runtime! We can't know that statically, so the test is failing as expected. I'm going to update the test expression rather update the optimization implementation.

tests/e2e_tests/queries.yml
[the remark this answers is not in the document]
terakilobyte: oh right, I had a brain fart!

terakilobyte (approved):

mongosql/src/mir/optimizer/match_null_filtering/mod.rs
```diff
@@ -182,21 +184,48 @@ impl Visitor for NullableSetter {
...
+            node.is_nullable = false;
+        }
         node
     }
 
     fn visit_expression(&mut self, node: Expression) -> Expression {
-        let mut node = node.walk(self);
-        node.set_is_nullable(false);
-        node
+        match node {
```
nbagnard, line 203: I am not sure why only these functions?
Looking at our semantics doc, a lot more functions apply here.

nbagnard: @mattChiaravalloti What about a query like `select * from foo where char_length(SUBSTRING((case when s = 'a' then null else s end) from start to 5)) > 0` for example. Is it handled somewhere else?

mattChiaravalloti: Which other ones do you have in mind? Remember these are expressions that evaluate to null independently.

Here's my breakdown for nullable expressions (note that the definition is recursive and I'm claiming NullIf, Coalesce, and the Null literal are the base cases):
```
Array -> can never be null
Cast -> can be null if either "on null" or "on error" are a nullable expression
DateFunction -> can be null if any of the arguments are a nullable expression
Document -> can never be null
Exists -> can never be null
FieldAccess -> technically can be null, but at this point we've guaranteed all possibly null field accesses are never null
Is -> can never be null
Like -> can be null if any of the arguments are a nullable expression
Literal -> only the null literal can be null
Reference -> can never be null
ScalarFunction -> <listed below, separately>
SearchedCase -> can be null if any of the branches (including else) are a nullable expression
SimpleCase -> can be null if any of the branches (including else) are a nullable expression
Subquery -> can be null if the result value is a nullable expression
SubqueryComparison -> can be null if either operand is a nullable expression
TypeAssertion -> nullability related to it's underlying type
```
For ScalarFunctions:
```
// String operators
Concat -> can be null if any operand is a nullable expression
// Unary arithmetic operators
Pos -> can be null if the operand is a nullable expression
Neg -> can be null if the operand is a nullable expression
// Arithmetic operators
Add -> can be null if any operand is a nullable expression
Sub -> can be null if any operand is a nullable expression
Mul -> can be null if any operand is a nullable expression
Div -> can be null if any operand is a nullable expression, or if the "on error" value is a nullable expression
// Comparison operators
Lt -> can be null if any operand is a nullable expression
Lte -> can be null if any operand is a nullable expression
Neq -> can be null if any operand is a nullable expression
Eq -> can be null if any operand is a nullable expression
Gt -> can be null if any operand is a nullable expression
Gte -> can be null if any operand is a nullable expression
Between -> can be null if any operand is a nullable expression
// Boolean operators
Not -> can be null if any operand is a nullable expression
And -> can be null if any operand is a nullable expression
Or -> can be null if any operand is a nullable expression
// Computed Field Access operator when the field is not known until runtime.
ComputedFieldAccess ->
// Conditional scalar functions
NullIf -> null if the operands are equal -- this is not something we cannot easily/fully check statically, therefore we must always assume this one is nullable
Coalesce -> null if none of the operands are non-null -- I'm actually realizing we can skip this one since we can rely on the recursive definition of nullable expression here (we'll either walk to an arg that must be nullable, or if none exist than this is definitely not null. The only case that would not cover is an empty arg list, which we could cover with constant folding... or might not even be syntactically valid.)
// Array scalar functions
Slice -> can be null if any operand is a nullable expression
Size -> can be null if any operand is a nullable expression
// Numeric value scalar functions
Position -> can be null if any operand is a nullable expression
CharLength -> can be null if the operand is a nullable expression
OctetLength -> can be null if the operand is a nullable expression
BitLength -> can be null if the operand is a nullable expression
Abs -> can be null if the operand is a nullable expression
Ceil -> can be null if the operand is a nullable expression
Cos -> can be null if the operand is a nullable expression
Degrees -> can be null if the operand is a nullable expression
Floor -> -> can be null if the operand is a nullable expression
Log -> can be null if any operand is a nullable expression
Mod -> can be null if any operand is a nullable expression
Pow -> can be null if any operand is a nullable expression
Radians -> can be null if the operand is a nullable expression
Round -> can be null if the operand is a nullable expression
Sin -> can be null if the operand is a nullable expression
Sqrt -> can be null if the operand is a nullable expression
Tan -> can be null if the operand is a nullable expression
// String value scalar functions
Replace -> can be null if any operand is a nullable expression
Substring  -> can be null if any operand is a nullable expression
Upper  -> can be null if the operand is a nullable expression
Lower -> can be null if the operand is a nullable expression
BTrim -> can be null if the operand is a nullable expression
LTrim -> can be null if the operand is a nullable expression
RTrim -> can be null if the operand is a nullable expression
Split -> can be null if the operand is a nullable expression
// Datetime value scalar function
CurrentTimestamp -> can never be null
Year -> can be null if the operand is a nullable expression
Month -> can be null if the operand is a nullable expression
Day -> can be null if the operand is a nullable expression
Hour -> can be null if the operand is a nullable expression
Minute -> can be null if the operand is a nullable expression
Second -> can be null if the operand is a nullable expression
Millisecond -> can be null if the operand is a nullable expression
Week -> can be null if the operand is a nullable expression
DayOfWeek -> can be null if the operand is a nullable expression
DayOfYear -> can be null if the operand is a nullable expression
IsoWeek -> can be null if the operand is a nullable expression
IsoWeekday -> can be null if the operand is a nullable expression
// MergeObjects merges an array of objects
MergeObjects -> can be null if any operand is a nullable expression
```

mattChiaravalloti: One way to think of this is to consider this example: `(a < 3) = null` where a is a possibly null field ref. Consider this optimization independently (i.e. no constant folding occurs). We filter out nullable `a` values and that leaves us with `(a < 3) = null` to update the nullability.

Is  `(a < 3) = null` nullable? To determine that we walk it's children.
Is `a < 3` nullable? To determine that we walk it's children.
Is `a` nullable? No.
Is `3` nullable? No.
=> `a < 3` is not nullable.
Is `null` nullable? Yes.
=> `(a < 3) = null` is nullable.

This implementation demonstrates that we can statically determine if an expression is nullable based on either it's inputs or it's intrinsic properties. The majority of expressions are only nullable based on their inputs. A small number of expressions are implicitly nullable: specifically, the null literal which must be null, and the NullIf function which may be null dynamically (and the Coalesce function, but I've talked myself out of that one).

mattChiaravalloti: > @mattChiaravalloti What about a query like `select * from foo where char_length(SUBSTRING((case when s = 'a' then null else s end) from start to 5)) > 0` for example. Is it handled somewhere else?

This will work because there is a null literal present in this expression tree. The null literal is intrinsically nullable, which makes the `case` epxr nullable, which makes the `substring` expr nullable, which makes the `char_length` expr nullable, which makes the `>` expr nullable. Notice that this visitor is written bottom-up, so we evaluate the leaf nodes first! (And we treat `NullIf` as a leaf since we know it is also intrinsically nullable, regardless of its inputs (we could optimize this further, but that should be follow-up work)).

mattChiaravalloti: You might also notice that this is a blanket approach that's actually too greedy. I'm going to file follow-up work to fine-tune it. Specifically, if a single intrinsic null value is anywhere in the expr tree, we say _all_ exprs in the tree visited after that one are nullable, but that's not totally true in reality.

My example above could be flipped to demonstrate that: `(a < 3) = null` could be flipped to `null = (a < 3)`. The visitor will visit the `null` before it visits `a < 3`, and it'll set the flag to `true` when it does. Then when it visits `a < 3`, the flag will already be true so it'll say `a < 3` is nullable even though it is not.

The fix for this is to reset the flag for each child of an expression before walking the child.

nbagnard (approved):
LGMT!
And, yes, you are right, and coaslesce can be skipped.

mattChiaravalloti:
Updating the Coalesce removal.

PR #83, opened 2025-04-28
SQL-2632: improve precision for numeric operators

jchemburkar:


agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1747,6 +1746,70 @@ impl DeriveSchema for UntaggedOperator {
...
+                        type_set.insert(Schema::Atomic(Atomic::Null));
+                    }
+                    Satisfaction::Not => {}
+                }
+                arg_schemas.insert(arg_schema.intersection(&NUMERIC.clone()));
+            }
+            // we can set the lower bound of numeric args as the largest type that is the minimum
+            // of one of the arguments. For example, if the lowest ordered type an argument can take
+            // on is a Double, the set that as the lower bound, as the operator will never return int or long.
+            let type_bound = arg_schemas
```
mattChiaravalloti, line 1772:
```suggestion
            let lower_bound = arg_schemas
```
Let's use the word "lower" in the name

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1747,6 +1746,70 @@ impl DeriveSchema for UntaggedOperator {
...
+            }
+            let mut schema = Schema::simplify(&Schema::AnyOf(type_set));
+            // the operators calling this, like multiply and add, will return a long if the
+            // inputs are integers and the value is sufficiently large, so we add that type
+            if schema.satisfies(&Schema::Atomic(Atomic::Integer)) != Satisfaction::Not
+                && schema != Schema::Unsat
+            {
+                schema = schema.union(&Schema::Atomic(Atomic::Long));
+            }
+            Ok(schema)
```
mattChiaravalloti, line 1810: 1. Do we need to simplify again here? If `schema` was `AnyOf([Int, Long])` already, will the union result in `AnyOf([Int, Long, Long])`?
2. We need to handle overflow for all types (int overflow -> long, long overflow -> double, double/decimal overflow -> also double (I'm pretty sure +/- inf is always a double although we should check that). ([link](https://www.mongodb.com/docs/manual/reference/operator/aggregation/add/#behavior))

agg-ast/schema_derivation/src/schema_derivation_tests/untagged_ops.rs
mattChiaravalloti, line 1: I think it is still worth adding a test where the two operands both have schema `AnyOf([Int, Double])` and `AnyOf([Int, Double])`. As it stands right now, I'm not sure any of these tests demonstrate having an `Int` lower bound _and_ adding `Long` because of that. In all tests, either the lower bound is already `Long`, or `Long` is present in the input type schemas so we don't add it based on the presence of `Int`.

mattChiaravalloti (changes requested):

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: 1. union of anyof and atomic with insert the atomic directly into the underlying btree set. so we're good there :) 
2. Decimal does have an inf type, which came up a lot in the fuzz tests. I'll add long overflow to double and add a similar test case to the one you described below

mattChiaravalloti: 1. Oh right right right, embarrassing 😅 of course we use a set not a vec for that
2. Ok thank you

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -1892,6 +1891,77 @@ impl DeriveSchema for UntaggedOperator {
...
+                && schema != Schema::Unsat
+            {
+                schema = schema.union(&Schema::Atomic(Atomic::Double));
+            }
+            if schema.satisfies(&Schema::Atomic(Atomic::Integer)) != Satisfaction::Not
+                && schema != Schema::Unsat
+            {
+                println!("{:?}", schema);
+                schema = schema.union(&Schema::Atomic(Atomic::Long));
+                println!("{:?}", schema);
```
mattChiaravalloti, line 1960:
```suggestion
                schema = schema.union(&Schema::Atomic(Atomic::Long));
```

mattChiaravalloti (approved):
lgtm mod the lingering prints

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: devastating

agg-ast/schema_derivation/src/schema_derivation.rs
```diff
@@ -2096,43 +2153,17 @@ impl DeriveSchema for UntaggedOperator {
...
-                if input_schema.satisfies(&Schema::Atomic(Atomic::Long)) != Satisfaction::Not {
-                    types.insert(Schema::Atomic(Atomic::Long));
-                }
-                if input_schema.satisfies(&Schema::Atomic(Atomic::Integer)) != Satisfaction::Not {
-                    types.insert(Schema::Atomic(Atomic::Integer));
-                    types.insert(Schema::Atomic(Atomic::Long));
-                };
-                handle_null_satisfaction(args, state, Schema::simplify(&Schema::AnyOf(types)))
-            }
             // mod returns the maximal numeric type of its inputs
```
bucaojit, line 2156: Here it states we're returning the maximal numeric and in the `$mod` docs that seems to be what we should do:
> Starting in version 7.2, the output data type of the $mod operator is the larger of the two input data types.
In the code I see it's combining all the numeric types, I could be missing something?

bucaojit (changes requested):
The changes look good, I had one question.

agg-ast/schema_derivation/src/schema_derivation.rs
jchemburkar: You are right! I perhaps over simplified code here. I just pushed a change that incorporates a similar lower bound approach. We can't just re-use the same helper idt because we don't care about overflow here (ie if hte resulting type is int, there is no point making it int or long)

bucaojit (approved):
Nice fix! LGTM

PR #85, opened 2025-05-01
SQL-2572: handle minkey properly in comparison

jchemburkar:


agg-ast/schema_derivation/src/schema_derivation_tests/match_expr.rs
```diff
@@ -280,6 +280,26 @@ mod comparison_ops {
...
+            keys: map! {
+                "foo".to_string() => Schema::AnyOf(set!(
+                    Schema::Atomic(Atomic::MinKey),
+                    Schema::Atomic(Atomic::Null),
+                ))
+            },
+            required: set!(),
+            ..Default::default()
+        })),
+        input = r#"{"$match": {"$expr": {"$lt": [{"$max": "$foo"}, null]}}}"#,
```
jchemburkar, line 295: this test case demonstrates why this matters in the first place. if foo was just a reference, or a literal, it would be trivial to evaluate foo < null (we'd constrain it directly). However, if foo is a expression, we need to get the null behavior correctly before recursing match_derive_schema on the sub expression. 

Here, because foo can be minkey, we must preserve minkey in the possible types (rather than saying it must be null or missing). The precise answer to this test would be AnyOf(MinKey, Missing) not inclusive of null, but we treat null and missing as the same, so its impossible as currently implemented to filter one but not the other just based on the null behavior.

terakilobyte (approved):
This LGTM. It's nice the improvement is so isolated.

PR #87, opened 2025-05-02
SQL-2821: Update tpch queries to use extjson syntax

mattChiaravalloti:
This PR updates the tpch queries to use extjson syntax for date literals. We previously converted date strings to datetimes via `::TIMESTAMP`. However, this was problematic since the implicit `ON ERROR` value for that `CAST` expression is the literal `NULL` value. That prevents any filters that use that `CAST` expressions by being treated as an MQL-semantic operator.

That translation is actually semantically correct but does lead to performance hits in cases like these tpch queries. The best way to work around it is to remove statically nullable non-field ref exprs from the filters. We did this here by using extjson syntax to represent literals instead of non-foldable `CAST` expressions. We will advise the same strategy to customers. I'll make that a "docs changes needed" note on this ticket when I close it.

### DIFF
--- a/mongosql/src/mir/optimizer/match_null_filtering/mod.rs
+++ b/mongosql/src/mir/optimizer/match_null_filtering/mod.rs
@@ -184,8 +184,6 @@ impl Visitor for NullableSetter {
         node
     }
 
-    // todo: only set nullable for expressions that don't include a null literal
-
     fn visit_scalar_function_application(
         &mut self,
         node: ScalarFunctionApplication,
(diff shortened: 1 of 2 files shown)


mongosql/src/mir/optimizer/match_null_filtering/mod.rs
```diff
@@ -184,8 +184,6 @@ impl Visitor for NullableSetter {
         node
     }
 
-    // todo: only set nullable for expressions that don't include a null literal
```
mattChiaravalloti, line 187: drive-by -- forgot to remove this from the previous pr that worked on this optimization 🚗

nbagnard (approved):
LGTM! Thank you for looking into this so quickly :)

terakilobyte (approved):

PR #89, opened 2025-05-07
SQL-2801: handle mutation of anyof schemas

jchemburkar:
When approaching this ticket, the question was whether there was an obvious way to solve this or if we can find workarounds for now until we change the underlying datastructure of anyofs. 

Ultimately, it was mostly the latter. for example, for $unwind, we can constrain the schema by removing any non Document schemas if preserve_null_and_empty_arrays is false. And if preserve_null_and_empty_arrays is true, we can get around anyofs by removing the field, mutating it, and unioning it back in.

The implication of this piecemeal is occasional lack of precision with required fields, which can be cleaned up when we can better handle AnyOfs.


agg-ast/schema_derivation/src/match_schema_derivation.rs
```diff
@@ -329,6 +329,7 @@ fn match_derive_schema_for_op(
         }
         MatchBinaryOp::Exists => {
             if is_exists_true_bson(b) {
+                state.result_set_schema = promote_missing(&state.result_set_schema);
```
jchemburkar, line 332: result set schema difference uses get_or_create_schema_for_path_mut to get field schemas to be modified, which is why this operator came up in the first place. result_set_schema_difference only removes explicitly named types, so we have to promote missing to make it visible to the function.

pmeredit (approved):
Looks great!

terakilobyte (approved):
LGTM

PR #90, opened 2025-05-07
SQL-2769: Update repo to use new common data loader for integration tests

mattChiaravalloti:
This PR primarily updates the repo to use the new common data loader from `sql-engines-common-test-infra`. Secondarily, it updates this repo to depend on that one as a module in evergreen, replaces all of the deduplicated functions and tasks from that repo (1), and updates the existing integration test generator in this repo to expect data/catalog to be pre-loaded. This last bit will be updated further in the follow-up ticket that migrates this repo to use the common test generation framework.

Sorry this is so big! The majority of files are just testdata files that rip the `catalog_data` and `catalog_schema` info from the various test files into separate files and in the new standard format. Once you see one file, you've basically seen them all -- the data and schema is moved over exactly as it was (mod some minor cosmetic updates for consistency).

(1) `install rust toolchain`, `clippy`, `rustfmt`, `unusued-dependencies`, `run rust tests`, `bootstrap mongo-orchestration`, `prepare resources`, `stop mongo-orchestration`


testdata/e2e_tests/exclude_namespaces.yml
mattChiaravalloti, line 1: This starts the list of "testdata" files. Again, these are essentially character-for-character copies for `catalog_data` and `catalog_schema` from all the various integration test files. The only changes are:
- formatting (all YAML instead of mixed YAML/JSON)
- quote characters (consistent use of one type, went with `"` instead of `'` since we mixed both previously)
- database names (can't keep attempting to create `foo.bar` or `mydb.test_coll` for each integration test file! now the database names, as you see here, are representative of which tests they are used for)

Feel free to look as many or as few of these in detail as you want. I don't think it's quite worthwhile to review them all but I won't stop you if you think there is good reason.

tests/e2e_tests/exclude_namespaces.yml
mattChiaravalloti, line 1: For the most part, all integration test files (e2e, errors, index, spec/query) were updated to not include `catalog_data` and `catalog_schema`, and to use the new database names as appropriate. There is one test I'll call out in particular that is worthwhile to reconsider, but otherwise these files only have deletions and nothing else of interest.

tests/spec_tests/query_tests/information_schema_lowercase.yml
```diff
@@ -1,20 +1,6 @@
...
-      'mytbl': {
-        'bsonType': 'object',
-        'additionalProperties': true,
-      }
-    }
-  }
-
 tests:
   - description: a user db called information_schema should require no special handling
+    skip_reason: "SQL-xxx: Should this actually be part of the spec?"
```
mattChiaravalloti, line 3: I added this skip_reason because this test in particular causes a bit of a dilemma... MongoSQL is case sensitive, but _MongoDB_ is not. As in, we can't create 2 databases called `INFORMATION_SCHEMA` and `information_schema` in mongod. The old way of running these was fine since we dropped all data before running a test. But now that we load everything upfront, we get an error when we attempt to create `information_schema` after creating `INFORMATION_SCHEMA`.

I argue that this test is not even worth keeping. The CAPS-only version is sufficient to demonstrate the "information schema" db does not exist in MongoSQL. What do you all think?

mattChiaravalloti:
Looks like I have some Windows debugging to do, and I need to update the `evergreen/mongotranslate.yml` stuff to use the `sql-engines-common-test-infra` module. I'll follow-up with those fixes soon. But otherwise, this is still ready for review.

mattChiaravalloti:
> Looks like I have some Windows debugging to do, and I need to update the `evergreen/mongotranslate.yml` stuff to use the `sql-engines-common-test-infra` module. I'll follow-up with those fixes soon. But otherwise, this is still ready for review.

Seemingly resolved the Windows issue by updating `sql-engines-common-test-infra` to set the `cargo_bin` expansion more appropriately based on OS. I'll put a PR up for that on that repo then retest on this one with the update.

jchemburkar (approved):
thanks for including the patch. The actual changes lgtm! parsing hte schemas looked pretty elegant

bucaojit (approved):
LGTM!  It's nice to see the improvements already with the cleaner evergreen and test files.

tests/spec_tests/query_tests/information_schema_lowercase.yml
bucaojit: I think that's a good reasoning.  Seemed like it's more of a difference in how the tests were run itself so I'm okay with removing it.

PR #92, opened 2025-05-08
SQL-2752: improve $literal ergonomics

jchemburkar:
At implementation time, it became clear that there weren't as many operators / stages effected as previously thought, and as such, the easiest way to implement this was to handle during deserialization rather than during derivation / translation. 

If, eg, we expect a string argument, we can get that string by deserializing the expression, and getting the value from either a. the literal value or b. the untagged $literal operator, where the first (and only) argument will be a literal value.

### DIFF
--- a/agg-ast/ast/src/custom_serde.rs
+++ b/agg-ast/ast/src/custom_serde.rs
@@ -1,10 +1,10 @@
 use crate::{
     definitions::{
         Cond, Convert, DateExpression, DateFromParts, DateFromString, DateToString, Expression,
-        LiteralValue, MatchArrayExpression, MatchArrayQuery, MatchBinaryOp, MatchElement,
+        GetField, LiteralValue, MatchArrayExpression, MatchArrayQuery, MatchBinaryOp, MatchElement,
         MatchExpression, MatchField, MatchNot, MatchNotExpression, MatchRegex, MatchStage,
-        ProjectItem, ProjectStage, Ref, SetWindowFieldsOutput, Trim, UntaggedOperator,
-        UntaggedOperatorName, VecOrSingleExpr, Window,
+        ProjectItem, ProjectStage, Ref, SetField, SetWindowFieldsOutput, Trim, UnsetField,
+        UntaggedOperator, UntaggedOperatorName, VecOrSingleExpr, Window,
     },
     map,
 };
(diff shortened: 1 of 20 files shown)


agg-ast/ast/src/custom_serde.rs
```diff
@@ -1312,3 +1331,184 @@ impl<'de> Deserialize<'de> for Convert {
...
+                        op: UntaggedOperatorName::Literal,
+                        args: a,
+                    })) => {
+                        if let Some(Expression::Literal(LiteralValue::String(s))) = a.first() {
+                            s.clone()
+                        } else {
+                            return Err(serde_err::custom("field to setField must be a string"));
+                        }
+                    }
+                    _ => return Err(serde_err::custom("field to setField must be a string")),
```
mattChiaravalloti, line 1355: The ticket mentions we should also allow `$toString` (and presumably `$convert`) expressions for the `field` argument. Is that something we'll do in a different ticket?

mattChiaravalloti (approved):
Looks good to me!

agg-ast/ast/src/custom_serde.rs
jchemburkar: Ah yes, I did skip over this and did not settle on what we shoudl actually do. The only way schema derivation can handle this properly is if the expression in a `$toString` is just a string literal. If it is non-literal, such as a field ref, then we wont be able to statically determine what the output type should be. I suppose we could just set it to Any? but that feels bad / unexpected. So this is technically incorrect for now, but I'm not sure adding it is of much value. wdyt?

nbagnard: We can add a follow up ticket for this.
Reading the description field can be "any valid expression that resolves to a string constant", which means that not only $toString, but other expressions apply to, no? 
And just to make sure that I understand correctly here.  If I have a valid pipeline like `setField( field: $dateToString(dateVal), input: xx, output: xxx)`, with DateVal being a constant, we will fail deserialization instead of passing and having the field type be any?

jchemburkar: Yes, that is the correct interpretation of the issue

agg-ast/ast/src/custom_serde.rs
```diff
@@ -1294,6 +1294,25 @@ impl<'de> Deserialize<'de> for Convert {
...
+                            | Some(Expression::Literal(LiteralValue::Int32(_)))
+                            | Some(Expression::Literal(LiteralValue::Int64(_)))
+                            | Some(Expression::Literal(LiteralValue::Double(_)))
+                            | Some(Expression::Literal(LiteralValue::Decimal128(_))) => {
+                                args[0].to_owned()
+                            }
+                            _ => return Err(serde_err::custom("invalid to format for convert")),
+                        }
+                    } else {
+                        to
```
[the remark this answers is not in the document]
jchemburkar: I suppose we can/should. I'm really toeing a line here -- because for $literal expressions, I was thinking, if we are unpacking already we could just verify its the correct format while we are there. But this leaks into "validating pipeline correctness" which is out of scope.

tests/schema_derivation_tests/query_spec_tests/array.yml
```diff
@@ -20,7 +20,7 @@ catalog_schema:
 tests:
   - description: SELECT empty array literal
     current_db: test
-    skip_reason: SQL-2752 improve $literal ergonomics
+    skip_reason: "cannot reprsent Array(unsat) in JSON schema"
```
nbagnard, line 23: Could you reference an JIRA here?
Create a new one if there isn't already one.

tests/schema_derivation_tests/query_spec_tests/operators_precedence.yml
```diff
@@ -883,7 +883,7 @@ tests:
   }
 - description: unary ops bind tighter than mul/div
   current_db: db
-  skip_reason: SQL-2752 improve $literal ergonomics
+  skip_reason: "negative integer literals deserialize as long"
```
nbagnard, line 886: Here too, can you have a JIRA reference please.
This is just to make it easier to know that we are aware of it and if we ever fix this, it is easy to find the test and "unskip" it.

tests/schema_derivation_tests/query_spec_tests/literal.yml
```diff
@@ -332,7 +332,7 @@ tests:
   }
 - description: Signed integer literals
   current_db: test
-  skip_reason: SQL-2752 improve $literal ergonomics
+  skip_reason: "negative integer literals deserialize as long"
```
nbagnard, line 335: Here too, can you have a JIRA reference please.

nbagnard (approved):
This is looking good 💯 
Just a couple of copy-paste left over and JIRA refs missing.
No need to retag me :)

PR #93, opened 2025-05-08
SQL-2648: handle $redact stage

jchemburkar:
After some investigation, there is not much we can do here. It seems the most common (only real?) use case is conditional redcating, which means we cannot constrain the rss, because a. it is difficult to statically evaluate conditions impacts on specific fields and b. the rss is only constrained based on access but could be unchanged if you have full access.


bucaojit (approved):
LGTM

PR #95, opened 2025-05-21
SQL-2834: Fix nested array assignment for project

jchemburkar:
please sanity check but I think the diff is fake, I just wrapped a block in an if statement.


terakilobyte (approved):

terakilobyte:
Yeah diff just appears to be the if to me, and the extra indentation

PR #97, opened 2025-05-22
SQL-2770: Update integration test generation to use common test generation framework

mattChiaravalloti:
This PR updates the e2e, error, spec/query, and schema derivation tests to use the new test generation framework. These tests all previously used the old repo-specific generation framework.

~This depends on #96 and on [this PR](https://github.com/10gen/sql-engines-common-test-infra/pull/15) in `sql-engines-common-test-infra`.~

~Sorry for the current massive diff (I rebased on #96's branch to check if this would work with schema derivation -- when reviewing, you can choose all but the first commit to get the smaller view). I confirmed locally that when the two dependencies are used in conjunction, all tests pass.~


README.md
```diff
@@ -45,27 +48,33 @@ Fuzz tests for pretty-printing.
 Index-usage assertion tests. These tests specify queries and expected index utilization.
 Requires a running mongod.
 
-`cargo test --features=index,generated --package=e2e-tests -- --test-threads=1` from the main directory
```
mattChiaravalloti, line 48: We no longer require single-threaded testing since all tests are now thread-safe 🎉

bucaojit (approved):
LGTM
That's great that single-threaded testing is no longer needed 🔥

test-utils/src/templates/schema_derivation_test_body_template
```diff
@@ -1,16 +1,13 @@
+
```
jchemburkar, line 1: nit: is this intentional?

test-utils/src/templates/query_test_header_template
```diff
@@ -11,30 +11,28 @@ use mongosql::{{
...
-    static mut TEST_FILE: Option<QueryYamlTestFile> = None;
+    static mut TEST_FILE: Option<YamlTestFile<QueryTestCase>> = None;
     static mut CATALOG: Option<Catalog> = None;
 
     INIT.call_once(|| {{
         unsafe {{
             CLIENT = Some(Client::with_uri_str(MONGODB_URI.clone()).map_err(Error::CannotCreateMongoDBClient).unwrap());
-            let test_file = parse_query_yaml_file(PathBuf::from({path})).unwrap();
-            let db_names = test_file.tests.iter().flat_map(|t| {{
+            let test_file = parse_yaml_test_file("{path}").unwrap();
```
jchemburkar, line 28: nit -- do we still need double curlys here?

test-utils/src/templates/schema_derivation_test_header_template
```diff
@@ -15,29 +15,27 @@ use mongosql::{{
...
+pub fn initialize_test() -> (&'static YamlTestFile<SchemaDerivationTestCase>, &'static BTreeMap<String, BTreeMap<String, json_schema::Schema>>) {{
+    static mut TEST_FILE: Option<YamlTestFile<SchemaDerivationTestCase>> = None;
     static mut CATALOG: Option<BTreeMap<String, BTreeMap<String, json_schema::Schema>>> = None;
 
     INIT.call_once(|| {{
         unsafe {{
             let client: Client = Client::with_uri_str(MONGODB_URI.clone()).map_err(Error::CannotCreateMongoDBClient).unwrap();
-            let test_file = parse_schema_derivation_yaml_file(PathBuf::from({path})).unwrap();
-            let db_names = test_file.tests.iter().flat_map(|t| {{
+            let test_file = parse_yaml_test_file("{path}").unwrap();
```
jchemburkar, line 31: similar q

test-utils/src/templates/query_test_body_template
```diff
@@ -1,23 +1,23 @@
+
```
jchemburkar, line 1: similar nit

jchemburkar (approved):
I had a few nits scattered but otherwise changes look good to me. This feels like a a huge upgrade / much cleaner to get this all done 😄

test-utils/src/templates/schema_derivation_test_body_template
mattChiaravalloti: The newline at the beginning is intentional, although I realize now it's probably not the best way to achieve this. My goal was to put a space between each function in the generated files (I was looking at them a lot during development, so for my own sanity it was useful to break them apart a bit).

I can either change this to be double newlines at the end of the file, or I can just remove this since now we're back to the case where it's unlikely someone needs to read the generated files. Or I can leave this as-is, if you don't care enough either way.

test-utils/src/templates/query_test_header_template
mattChiaravalloti: Can you clarify which braces you're asking about? The line the comment is associated with implies you are asking about `{path}`, is that right? If so, this had single-braces before this pr and continues to have them now as part of this one. The single-braces mean the entire expression should be replaced with its named mapping, in this case `"path"`.

If you mean other braces, it seems to me all of the double-ones are appropriate. I don't see any instances of `{{` or `}}` in any of the generated files.

PR #100, opened 2025-05-29
SQL-2840: add schemas for schema derivation tests

jchemburkar:


testdata/schema_derivation_tests/order_by.yml
```diff
@@ -32,7 +41,15 @@ dataset:
...
-      additionalProperties: true
+      required: [ "_id", "a" ]
+      additionalProperties: false
+      properties:
+        _id:
+          bsonType: "int"
+        a:
+          anyOf:
+            - bsonType: "int"
+            - bsonType: "null"
```
mattChiaravalloti, line 52: This works with `!!str`?

mattChiaravalloti (commented):
2 questions:
1 - see my comment about `!!str` before the `"null"` string literal
2 - There are several schemas that do not include info for `_id` (it's not in `required` or in `properties`). Do we need/want to update those to include `_id` schema? Relevant files/collections: `operators_control_flow.yml - "bar"`, `scopes.yml - "schema_coll" & "schema_foo"`, and `with.yml - "baz"`.

testdata/schema_derivation_tests/order_by.yml
jchemburkar: `  {
    _id: 'nullAndNonNull',
    type: 'collection',
    schema: {
      bsonType: 'object',
      required: [ '_id', 'a' ],
      additionalProperties: false,
      properties: {
        _id: { bsonType: 'int' },
        a: { anyOf: [ { bsonType: 'int' }, { bsonType: 'null' } ] }
      }
    },
    lastUpdated: ISODate('2025-05-29T02:55:09.119Z')
  },` 🤷‍♂️

jchemburkar:
@mattChiaravalloti re your second question -- it can't hurt. It looks like all of those are schemas where I just copied the schema from the spec tests. A broader question would be if we should update these there, and if its worth filling out the schemas for the Any schemas in the spec tests more broadly. I am thinking yes to the former and no the latter

testdata/schema_derivation_tests/order_by.yml
mattChiaravalloti: This is great news

mattChiaravalloti:
> This LGTM, but I'm unclear as to why these were incomplete to begin with and am now curious :)

I suspect those are some of the oldest spec test data sets where schema didn't really matter because we'd use type assertions in the queries. There are probably other reasons too, but I remember us pretty liberally using `::!` early on for spec testing.

PR #105, opened 2025-07-17
SQL-2852: Eliminate $addFields before $project in very specific circumstances

pmeredit:
This eliminates an $addFields before a $project when they define the exact same values only. We need to improve substitution if we want to do more.


mongosql/src/mir/optimizer/dead_code_elimination/mod.rs
```diff
@@ -39,81 +43,127 @@ struct DeadCodeEliminationVisitor {
...
-                                        expression: new_expr,
-                                        is_add_fields: false,
-                                        cache: SchemaCache::new(),
-                                    })
-                                }
-                                // It is possible for substitution to fail if the Group clause
-                                // contains Subqueries. This will be very rare.
-                                Err(n) => *n,
-                                _ => unreachable!(),
+impl DeadCodeEliminationVisitor {
```
pmeredit, line 46: changes here were just to reduce cloning

mongosql/src/mir/optimizer/dead_code_elimination/mod.rs
```diff
@@ -39,81 +43,127 @@ struct DeadCodeEliminationVisitor {
...
+                Err(g) => Stage::Group(g),
+            }
+        } else {
+            // If the Group uses Keys that are not defined by the Project source,
+            // do not perform the substitution or eliminate the Project.
+            Stage::Group(og)
+        }
+     }
+
+    fn project_deadcode_elimination(&mut self, mut p: Project) -> Stage {
```
pmeredit, line 119: This is all the actual new stuff

jchemburkar (approved):
lgtm 🫡

PR #106, opened 2025-09-30
SQL-2953: handle different server versions in integration tests

jchemburkar:
companion pr [here](https://github.com/mongodb/sql-engines-common-test-infra/pull/21). Copying the description: 

Ultimately the simplest approach I found was to:

take the version (already part of the test config) and pass it as a variable to the run rust test task
update the run rust test script to export this variable so the test generator in mongosql has access to this
have tests specify if there is a minimum or maximum server version the test is valid for, and if this option is specified, verify the underlying server version is in bounds.
I think this was the simplest way since we are round tripping from mongosql -> common test infra -> mongosql.

PR will fail until shared test infra is merged as it is a module. 

Here is a patch with the tests passing: https://spruce.mongodb.com/version/68dc46486dddaf000743eaf3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2389,9 +2389,9 @@ dependencies = [
 
 [[package]]
 name = "mongodb"
-version = "3.2.2"
+version = "3.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e5f2b7791f13490c99dc2638265006b08e8675945671e29bd0d763da09fdc61"
+checksum = "d0f8c69f13acf07eae386a2974f48ffd9187ea2aba8defbea9aa34e7e272c5f3"
 dependencies = [
  "async-trait",
  "base64 0.13.1",
@@ -2421,7 +2421,7 @@ dependencies = [
  "serde",
  "serde_bytes",
  "serde_with",
- "sha-1",
+ "sha1",
  "sha2",
  "socket2",
  "stringprep",
@@ -2438,9 +2438,9 @@ dependencies = [
 
 [[package]]
 name = "mongodb-internal-macros"
-version = "3.2.2"
+version = "3.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d848d178417aab67f0ef7196082fa4e95ce42ed9d2a1cfd2f074de76855d0075"
+checksum = "b9202de265a3a8bbb43f9fe56db27c93137d4f9fb04c093f47e9c7de0c61ac7d"
 dependencies = [
  "macro_magic",
  "proc-macro2",
@@ -3773,10 +3773,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "sha-1"
-version = "0.10.1"
+name = "sha1"
+version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5058ada175748e33390e40e872bd0fe59a19f265d0158daa551c5a88a76009c"
+checksum = "e3bf829a2d51ab4a5ddf1352d8470c140cadc8301b2ae1789db023f01cedd6ba"
 dependencies = [
  "cfg-if",
  "cpufeatures",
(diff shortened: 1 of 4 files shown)


tests/spec_tests/query_tests/type_conversions.yml
```diff
@@ -981,6 +984,24 @@ tests:
       - {'': {'type': 'minKey', 'asString': null}}
       - {'': {'type': 'maxKey', 'asString': null}}
 
+    # in mongodb 8.3.0, many of the previously unsupported casts to string became valid. This test
```
jchemburkar, line 987: I considered breaking the invalid types dataset up entirely into the individual types and adding two tests for each -- one with a max version of 8.2 and one with a min version of 8.3.

For now, I decided it would be easier to just leave this as is, but happy to bikeshed on this if anyone feels strongly.

test-utils/src/lib.rs
```diff
@@ -165,3 +165,37 @@ pub fn drop_catalog_data<T: Into<String>>(
...
+    min_server_version: &Option<String>,
+    max_server_version: &Option<String>,
+    server_version: &Option<String>,
+) -> bool {
+    // max server version implies there is some version above which the test should not run; hence
+    // if the server version is "latest" we skip the test.
+    if server_version == &Some("latest".to_string()) {
+        return !max_server_version.is_some();
+    }
+    // rapid can be a moving target, so for now, we will skip tests that have version constraints on rapid
```
jchemburkar, line 180: as mentioned in this comment, rapid right now is 8.2.1; this means that for now, it disagrees with "latest" when it comes to how these specific tests behave. And in general, these bleeding edge versions probably will waffle back and forth and require a lot of small changes as rapid catches up with latest. So for now, I just skip tests for `rapid` if a min or max server version is specified.

test-utils/src/lib.rs
```diff
@@ -165,3 +165,37 @@ pub fn drop_catalog_data<T: Into<String>>(
...
+    if server_version == &Some("rapid".to_string()) {
+        return max_server_version.is_some() && min_server_version.is_some();
```
bucaojit, line 182: Can you clarify this logic.  It should return `true` if these are both not specified?
Wouldn't it then be:
```
    return !min_server_version.is_some() && !max_server_version.is_some();
```

or use `is_none()` instead

test-utils/src/lib.rs
```diff
@@ -165,3 +165,37 @@ pub fn drop_catalog_data<T: Into<String>>(
...
+    }
+    // if the min server version is specified, check that the server version is >= min
+    if let Some(min_version) = min_server_version.as_ref() {
+        if let Some(server_version) = server_version.as_ref() {
+            if server_version < min_version {
```
bucaojit, line 187: I'm not sure if this comparison is enough to compare the versions if they're coming in as strings.  Try testing with "8.10" and "8.2".  I think for these values it would say 8.10 is less than 8.2.

bucaojit (changes requested):

test-utils/src/lib.rs
jchemburkar: unless it changed in another version of rust I think the lexicographic order should apply here. eg if you use the repl [here](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024) with the following snippet:
```
#![allow(unused)]
fn main() {
    let x = "8.10" < "8.2";
    println!("{}", x);
}
```
it should correctly have 8.2 as greater. 

The reason I didnt initially just float cast these is that I thought it might lead to _more_ complexity -- either if we have the string values (rapid, latest) or if in the future we allow some sort of prerelease values (8.1.0-alpha vs 8.2.0, eg). Just doing string comparison isn't perfect but should be roughly right for this. But since we control the values being passed/set via the mongodb_version variable maybe it doesnt matter. wdyt?

jchemburkar:
updated run with float version comparison: https://spruce.mongodb.com/version/68e009cd368fcc00072a35b3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

EthanHardyMongo (approved):

PR #132, opened 2026-03-31
SQL-3163: move schema builder library to mongosql

jchemburkar:
This pr moves the schema-builder-library folder from schema-manager-rs to mongosql, while preserving git history. It also moves relevant evergreen tests over.

a successful evergreen run can be found [here](https://spruce.corp.mongodb.com/version/69cb15b9250b990007546413/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

this pr will need to be merged before the corresponding schema-manager-rs [pull request](https://github.com/10gen/schema-manager-rs/pull/989), which will then pull the schema builder library in as a dependency.


EthanHardyMongo (commented):
LGTM! Just two things: 

1. Lets add the part of the readme in `schema-manager-rs` that explains how to run the integration tests to this readme. Its the [Integration testing for schema-builder-library](https://github.com/10gen/schema-manager-rs/blob/main/README.md#integration-testing-for-schema-builder-library) section.
2. I was checking your patch, and I noticed that the `ubuntu2204` variant didn't run. Then I checked our evergreen project settings and noticed that we were trying to run `ubuntu2004` instead of `ubuntu2204`. I guess we updated this at some point but forgot to update the evergreen settings. Anyways, I fixed it in our settings, so you may want to push an empty commit just to make sure the integration tests are working properly (because they're only in the `ubuntu2204` variant).

EthanHardyMongo (approved):
LGTM! Although it looks like merging is blocked due to some unverified commits.
> Commits must have verified signatures.

This happened to me once. To fix it, I had to rebase and resign all the unverified commits. You'll have to force push which will change the commit hash and committer date, but all the important git history will be preserved.

nicholascioli (approved):
LGTM!

PR #135, opened 2026-04-06
SQL-3083: Fix inefficiency in match_null_filtering optimization

bucaojit:
Fix to allow `found_impure` flag to be set.  

Confirmed by running the schema manager using the test in the JIRA ticket, that the NullableSetter visitor is not entered. 

Added unit test. 

This looked to cover the issue with `Coalesce` so i did not add a separate case in `visit_expression()`.

### DIFF
--- a/mongosql/src/mir/optimizer/match_null_filtering/test.rs
+++ b/mongosql/src/mir/optimizer/match_null_filtering/test.rs
@@ -790,6 +790,65 @@ mod all_fields_always_nullable {
     );
 }
 
+mod impure_and_pure_field_access {
+    use super::*;
+    
+    test_match_null_filtering!(
+        impure_field_access_with_pure_nullable_field_does_not_flip_semantics,
+        expected = Stage::Filter(Filter {
+            source: Box::new(Stage::Filter(Filter {
+                source: mir_collection("db", "foo"),
+                condition: field_existence_expr("foo", vec!["b"], 0u16, vec![true]),
+                cache: SchemaCache::new(),
+            })),
+            condition: Expression::ScalarFunction(ScalarFunctionApplication {
+                function: ScalarFunction::Gt,
+                args: vec![
+                    Expression::FieldAccess(FieldAccess {
+                        expr: Box::new(Expression::ScalarFunction(ScalarFunctionApplication {
+                            function: ScalarFunction::Coalesce,
+                            args: vec![
+                                field_access_expr("foo", vec!["o1"], 0u16, vec![true]),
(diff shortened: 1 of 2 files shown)


mongosql/src/mir/optimizer/match_null_filtering/test.rs
```diff
@@ -790,6 +790,65 @@ mod all_fields_always_nullable {
...
+                        expr: Box::new(Expression::ScalarFunction(ScalarFunctionApplication {
+                            function: ScalarFunction::Coalesce,
+                            args: vec![
+                                field_access_expr("foo", vec!["o1"], 0u16, vec![true]),
+                                field_access_expr("foo", vec!["o2"], 0u16, vec![true]),
+                            ],
+                            is_nullable: true,
+                        })),
+                        field: "a".to_string(),
+                        is_nullable: false,
```
mattChiaravalloti, line 846: Can you add a comment here explaining why we have this `FieldAccess` marked as `is_nullable: false` in the input?

By saying `COALESCE(o1, o2).a` is not nullable in the `input`, it becomes confusing to see the `expected` result. If the expression is `COALESCE(o1, o2).a > b` and we know that `a` is not nullable before match-null-filtering is applied and that `b` is not nullable after match-null-filtering is applied, why wouldn't we expect the `Gt`'s `is_nullable` to flip to false? Well the answer is that we actually wish it could flip but right now we can't precisely determine nullability of impure `FieldAccess` expressions! We should have a comment explaining that.

I think something like this could be helpful: "We artificially set `is_nullable` to `false` for this `FieldAccess` even though that is not a realistic value for it. In a real query, this `FieldAccess` would be marked as `is_nullable: true`. However, this test aims to demonstrate that the mere presence of impure field paths results in semantics not being flipped for filtered expressions. Here, we add a null-filter for `b`, but still do not flip the `is_nullable` flag for the `Gt` since `COALESCE(o1, o2).a` is impure."

Feel free to reword that however you want!

mattChiaravalloti (approved):
Looks good to me! Great idea for the unit test! 💯 I think it will benefit long-term from a comment explaining it, which I suggested below.

EthanHardyMongo (approved):
LGTM!

mongosql/src/mir/optimizer/match_null_filtering/test.rs
bucaojit: Yes great point! I added the comment you suggested, it explains things perfectly thanks.

PR #136, opened 2026-04-09
SQL-3164: create basic DataService trait with wasm and rust implementations

jchemburkar:
implemented a basic data service trait with wasm and rust default implementations 

I only did basic deserialization tests because when we integrate the library to using the data service trait, we will get the benfit of the existing integration test suite.

### DIFF
--- a/schema-builder-library/Cargo.toml
+++ b/schema-builder-library/Cargo.toml
@@ -10,7 +10,6 @@ name = "schema_builder_library"
 path = "src/lib.rs"
 crate-type = ["cdylib", "rlib"]
 
-# Common dependencies for all targets
 [dependencies]
 futures = { workspace = true }
 serde = { workspace = true }
@@ -24,28 +23,39 @@ tracing = { workspace = true }
 bson.workspace = true
 async-trait = { workspace = true }
 
+# Optional dependencies enabled by features
+tokio = { version = "1", default-features = false, optional = true }
+
+# native-client dependencies
+mongodb = { workspace = true, optional = true }
+test-utils = { path = "../test-utils", optional = true }
+
+# wasm dependencies
+wasm-bindgen = { version = "0.2", optional = true }
+wasm-bindgen-futures = { version = "0.4", optional = true }
+serde-wasm-bindgen = { version = "0.6", optional = true }
+js-sys = { version = "0.3", optional = true }
+console_error_panic_hook = { version = "0.1", optional = true }
+
 [dev-dependencies]
 tracing-test = "0.2"
-
-# Native-only dependencies (excluded from WASM)
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-mongodb = { workspace = true }
-tokio = { workspace = true, features = ["full"] }
-test-utils = { path = "../test-utils" }
-
-# WASM-only dependencies
-[target.'cfg(target_arch = "wasm32")'.dependencies]
-wasm-bindgen = "0.2"
-wasm-bindgen-futures = "0.4"
-serde-wasm-bindgen = "0.6"
-js-sys = "0.3"
-tokio = { version = "1", default-features = false, features = ["sync", "macros", "rt"] }
-console_error_panic_hook = "0.1"
 
 [features]
 integration = []
 sync = []
-default = []
+default = ["native-client"]
+native-client = ["dep:mongodb", "dep:test-utils", "dep:tokio", "tokio/full"]
+wasm = [
+    "dep:console_error_panic_hook",
+    "dep:js-sys",
+    "dep:serde-wasm-bindgen",
+    "dep:wasm-bindgen",
+    "dep:wasm-bindgen-futures",
+    "dep:tokio",
+    "tokio/sync",
+    "tokio/macros",
+    "tokio/rt",
+]
 
 [lints.clippy]
 unwrap_used = "deny"
--- a/schema-builder-library/Cargo.toml
+++ b/schema-builder-library/Cargo.toml
@@ -32,9 +32,7 @@ test-utils = { path = "../test-utils", optional = true }
 
 # wasm dependencies
 wasm-bindgen = { version = "0.2", optional = true }
-wasm-bindgen-futures = { version = "0.4", optional = true }
 serde-wasm-bindgen = { version = "0.6", optional = true }
-js-sys = { version = "0.3", optional = true }
 console_error_panic_hook = { version = "0.1", optional = true }
 
 [dev-dependencies]
@@ -46,10 +44,8 @@ default = ["native-client"]
 native-client = ["dep:mongodb", "dep:test-utils", "dep:tokio", "tokio/full"]
 wasm = [
     "dep:console_error_panic_hook",
-    "dep:js-sys",
     "dep:serde-wasm-bindgen",
     "dep:wasm-bindgen",
-    "dep:wasm-bindgen-futures",
     "dep:tokio",
     "tokio/sync",
     "tokio/macros",
--- a/schema-builder-library/Cargo.toml
+++ b/schema-builder-library/Cargo.toml
@@ -40,7 +40,7 @@ wasm-bindgen-futures = "0.4"
 serde-wasm-bindgen = "0.6"
 js-sys = "0.3"
 tokio = { version = "1", default-features = false, features = ["sync", "macros", "rt"] }
-console_error_panic_hook = { version = "0.1", optional = true }
+console_error_panic_hook = "0.1"
 
 [features]
 integration = []
--- a/schema-builder-library/Cargo.toml
+++ b/schema-builder-library/Cargo.toml
@@ -8,19 +8,18 @@ edition = "2024"
 [lib]
 name = "schema_builder_library"
 path = "src/lib.rs"
+crate-type = ["cdylib", "rlib"]
 
+# Common dependencies for all targets
 [dependencies]
 futures = { workspace = true }
-mongodb = { workspace = true }
 serde = { workspace = true }
-tokio = { workspace = true, features = ["full"] }
 mongosql = { workspace = true }
 agg-ast = { workspace = true }
 schema_derivation = { workspace = true }
 thiserror = { workspace = true }
 lazy_static = { workspace = true }
 glob = { workspace = true }
-test-utils = { path = "../test-utils" }
 tracing = { workspace = true }
 bson.workspace = true
 async-trait = { workspace = true }
@@ -28,8 +27,25 @@ async-trait = { workspace = true }
 [dev-dependencies]
 tracing-test = "0.2"
 
+# Native-only dependencies (excluded from WASM)
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+mongodb = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+test-utils = { path = "../test-utils" }
+
+# WASM-only dependencies
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2"
+wasm-bindgen-futures = "0.4"
+serde-wasm-bindgen = "0.6"
+js-sys = "0.3"
+tokio = { version = "1", default-features = false, features = ["sync", "macros", "rt"] }
+console_error_panic_hook = { version = "0.1", optional = true }
+
 [features]
 integration = []
+sync = []
+default = []
 
 [lints.clippy]
 unwrap_used = "deny"
--- a/schema-builder-library/Cargo.toml
+++ b/schema-builder-library/Cargo.toml
@@ -42,7 +42,6 @@ tracing-test = "0.2"
 
 [features]
 integration = []
-sync = []
 default = ["native-client"]
 native-client = ["dep:mongodb", "dep:test-utils", "dep:tokio", "tokio/full"]
 wasm = [
--- a/schema-builder-library/src/collection/mod.rs
+++ b/schema-builder-library/src/collection/mod.rs
@@ -36,10 +36,10 @@ static EXCLUDE_DUNDERSCORE_PATTERN: LazyLock<glob::Pattern> = LazyLock::new(|| {
 
 static INCLUDE_LIST_IN_DB_AND_COLL_PAIRS: OnceLock<Vec<(String, String)>> = OnceLock::new();
 
-/// CollectionInfo is responsible for extracting the collections and views
-/// and preparing them for processing.
+/// DatabaseCollections holds all collections for a single database, categorized by type
+/// (views, regular collections, and timeseries), and prepared for schema processing.
 #[derive(Debug, Default)]
-pub(crate) struct CollectionInfo {
+pub(crate) struct DatabaseCollections {
     pub views: Vec<CollectionDoc>,
     pub collections: Vec<CollectionDoc>,
     pub timeseries: Vec<CollectionDoc>,
@@ -86,8 +86,8 @@ pub(crate) async fn query_for_initial_schemas(
     Ok(initial_collection_schemas)
 }
 
-impl CollectionInfo {
-    /// Create a new CollectionInfo instance. Collections and Views within a database
+impl DatabaseCollections {
+    /// Create a new DatabaseCollections instance. Collections and Views within a database
     /// will be enumerated, checked for inclusion/exclusion, and prepared for
     /// processing. The caller must actually process the collections/views by calling
     /// their JoinHandle.
@@ -108,7 +108,7 @@ impl CollectionInfo {
             )
             .await?;
 
-        CollectionInfo::separate_collection_types(
+        DatabaseCollections::separate_collection_types(
             db_name,
             &include_list,
             &exclude_list,
--- a/schema-builder-library/src/collection/mod.rs
+++ b/schema-builder-library/src/collection/mod.rs
@@ -35,8 +35,8 @@ static EXCLUDE_DUNDERSCORE_PATTERN: LazyLock<glob::Pattern> = LazyLock::new(|| {
 
 static INCLUDE_LIST_IN_DB_AND_COLL_PAIRS: OnceLock<Vec<(String, String)>> = OnceLock::new();
 
-/// DatabaseCollections holds all collections for a single database, categorized by type
-/// (views, regular collections, and timeseries), and prepared for schema processing.
+/// DatabaseCollections is responsible for extracting the collections and views
+/// and preparing them for processing.
 #[derive(Debug, Default)]
 pub(crate) struct DatabaseCollections {
     pub views: Vec<CollectionInfo>,
--- a/schema-builder-library/src/collection/mod.rs
+++ b/schema-builder-library/src/collection/mod.rs
@@ -327,14 +327,8 @@ impl DatabaseCollections {
                     if let Ok(pipeline) = view_doc.options
                         .pipeline
                         .iter()
-                        .map(|doc| bson::from_document(doc.clone()))
-                        .try_fold(Vec::<agg_ast::definitions::Stage>::new(), |mut acc, doc| match doc {
-                            Ok(stage) => {
-                                acc.push(stage);
-                                Ok(acc)
-                            }
-                            Err(e) => Err(e),
-                        }) {
+                        .map(|doc| bson::from_document::<agg_ast::definitions::Stage>(doc.clone()))
+                        .collect::<core::result::Result<Vec<_>, _>>() {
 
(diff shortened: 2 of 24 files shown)


schema-builder-library/src/collection/mod.rs
```diff
@@ -4,8 +4,9 @@
  */
 use crate::{
     Error, NamespaceInfo, NamespaceInfoWithSchema, NamespaceType, Result, client_util::DatabaseExt,
-    derive_schema_for_partitions, derive_schema_for_view, get_partitions,
-    result_set::ShareableResultSet, schema::initial_schema::InitialSchema,
+    data_service::CollectionInfo as CollectionDoc, derive_schema_for_partitions,
```
nicholascioli: Nit: Is there any reason for this alias vs just using `CollectionInfo`?

schema-builder-library/src/collection/mod.rs
nicholascioli: Nit: You can just use collect here. (Note, you'll need to use the fully qualified `Result` type since it is getting shadowed by an import)

```rust
.collect::<core::result::Result<Vec<_>, _>>()
```

schema-builder-library/src/collection/patterns.rs
```diff
@@ -201,21 +202,21 @@ impl CollectionInfo {
...
                 database,
                 &collection_doc,
                 include_list,
                 exclude_list,
             )? {
-                if collection_doc.type_ == "view" {
+                if collection_doc.collection_type == "view" {
                     collection_info.views.push(collection_doc);
-                } else if collection_doc.type_ == "timeseries" {
+                } else if collection_doc.collection_type == "timeseries" {
```
nicholascioli: Nit: Since we control `CollectionInfo`, could we make the `collection_type` field be an enum? Raw strings scare me.

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+pub use wasm::{JsDataService, WasmDataService};
+
+/// Information about a single collection entry, returned by `list_collections`.
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct CollectionInfo {
+    /// The name of the collection.
+    pub name: String,
+    /// The type of the collection ("collection", "view", or "timeseries").
+    #[serde(rename = "type")]
+    pub collection_type: String,
```
nicholascioli: Nit: Could this be an enum?

nicholascioli: Other nit: you can also just name this `r#type` and not need the `#[serde(rename = "type")]` at the top.

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Additional options (primarily used for views and timeseries).
+    #[serde(default)]
+    pub options: CollectionOptions,
+}
+
+/// Options for collections. View options and timeseries options are both represented here
+/// since the `options` field in `listCollections` output is overloaded for different collection types.
+/// Consumers should check `CollectionInfo::collection_type` before accessing the relevant fields.
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct CollectionOptions {
```
nicholascioli, line 45: Nit: You can use the `rename_all` option for serde here as `camelCase`.

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// For views, the name of the source collection.
+    #[serde(rename = "viewOn", default)]
+    pub view_on: String,
+    /// For views, the aggregation pipeline.
+    #[serde(default)]
+    pub pipeline: Vec<Document>,
+    /// For timeseries collections, the timeseries options.
+    #[serde(rename = "timeseries", default)]
```
nicholascioli: Nit: Is there any difference between the rename here and the actual name of the field?

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TimeSeriesOptions {
+    pub time_field: String,
+    pub meta_field: Option<String>,
+}
+
+/// Abstraction over database operations used by the schema builder.
+///
+/// Uses namespace strings in `"database.collection"` format for collection operations.
```
nicholascioli: I'm not really sure what you mean by this, since methods in this trait that take in a namespace do so by having separate arguments for the DB and collection.

schema-builder-library/Cargo.toml
```diff
@@ -8,28 +8,44 @@ edition = "2024"
...
+test-utils = { path = "../test-utils" }
+
+# WASM-only dependencies
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2"
+wasm-bindgen-futures = "0.4"
+serde-wasm-bindgen = "0.6"
+js-sys = "0.3"
+tokio = { version = "1", default-features = false, features = ["sync", "macros", "rt"] }
+console_error_panic_hook = { version = "0.1", optional = true }
```
nicholascioli: Hmm, this section might be better off gated behind a feature flag rather than an architectural one. Feature flags would allow for more flexibility in testing and other targets that might not match wasm exactly, but spiritually.

As a recommendation, it might be nice to have a feature flag called `wasm` for all things wasm related and (maybe) a different feature flag called `native-client` for anything relating to default implementations using the native MongoDB client. The `native-client` feature could be enabled by default.

Rust also allows you to specify compile errors if you try to enable features that don't work with each other by doing something like the following:

```rust
#[cfg(all(feature = "native-client", feature = "wasm"))]
compile_error!("the native client does not work in WASM");
```

Possible configuration for `Cargo.toml` below:
```toml
[features]
default = ["native-client"]
native-client = [
  "dep:mongodb",
  "dep:test-utils",
  
  "tokio/full"
]
wasm = [
  "dep:console_error_panic_hook",
  "dep:js-sys",
  "dep:serde-wasm-bindgen",
  "dep:wasm-bindgen",
  "dep:wasm-bindgen-features",

  "tokio/macros", "tokio/rt", "tokio/sync"
]

tokio = { workspace = true, default-features = false, features = ["full"], optional = true }

# Native-client dependencies
mongodb = { workspace = true, optional = true }
test-utils = { path = "../test-utils", optional = true }

# WASM-only dependencies
wasm-bindgen = { version = "0.2", optional = true }
wasm-bindgen-futures = { version = "0.4", optional = true }
serde-wasm-bindgen = { version = "0.6", optional = true }
js-sys = { version = "0.3", optional = true }
console_error_panic_hook = { version = "0.1", optional = true }
```

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
+use bson::Document;
+use serde::{Deserialize, Serialize};
+
+#[cfg(test)]
+mod test;
+
+use crate::Result;
```
nicholascioli: Do we want to use the big library `Error` type here? It seems like the set of possible errors stemming from database interactions are small enough to warrant its own Error type instead.

I see that you needed to disable certain error types entirely in WASM anyway, so it might be beneficial to just make a separate type instead.

schema-builder-library/src/data_service/mongodb.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.client.list_database_names().await.map_err(Error::from)
+    }
+
+    async fn list_collections(&self, db_name: &str) -> Result<Vec<CollectionInfo>> {
+        let db = self.client.database(db_name);
+        let cursor = db
+            .run_cursor_command_with_read_preference(
+                doc! { "listCollections": 1.0, "authorizedCollections": true },
+            )
+            .await?;
```
nicholascioli: Is there a reason to use the raw command vs just using [`Database::list_collections`](https://docs.rs/mongodb/3.5.2/mongodb/struct.Database.html#method.list_collections) here?

I thought that maybe the `authorizedCollections: true` flag might be the reason, but apparently it has no effect if not used with `nameOnly: true`? (See: https://www.mongodb.com/docs/manual/reference/command/listCollections/#command-fields)

schema-builder-library/src/data_service/test.rs
```diff
@@ -0,0 +1,159 @@
...
+/// `documents` drives both `aggregate` and `find`, keyed by `"db.collection"`.
+#[allow(dead_code)]
+pub(crate) struct MockDataService {
+    pub databases: Vec<String>,
+    pub collections: HashMap<String, Vec<CollectionInfo>>,
+    pub documents: HashMap<String, Vec<Document>>,
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
```
nicholascioli: This might not be necessary since tests are almost always run on the dev machine. Do you expect tests to be run within a WASM context?

schema-builder-library/src/data_service/test.rs
```diff
@@ -0,0 +1,159 @@
...
+impl DataService for MockDataService {
+    async fn list_databases(&self) -> Result<Vec<String>> {
+        Ok(self.databases.clone())
+    }
+
+    async fn list_collections(&self, db_name: &str) -> Result<Vec<CollectionInfo>> {
+        Ok(self.collections.get(db_name).cloned().unwrap_or_default())
+    }
```
nicholascioli: Nit: Do we want listing a collection on a non-existing database to not error?

schema-builder-library/src/data_service/test.rs
```diff
@@ -0,0 +1,159 @@
...
+    };
+    let info: CollectionInfo = bson::from_document(doc).unwrap();
+    assert_eq!(info.name, "myView");
+    assert_eq!(info.collection_type, "view");
+    assert_eq!(info.options.view_on, "sourceCollection");
+    assert_eq!(
+        info.options.pipeline,
+        vec![bson::doc! { "$match": { "active": true } }]
+    );
+    assert_eq!(info.options.timeseries, None);
```
nicholascioli: Nit: (Also applies to the other tests that do this) any reason why these don't test for equality of the full `CollectionInfo` struct like in `test_plain_collection_deserialization`?

schema-builder-library/src/data_service/wasm.rs
```diff
@@ -0,0 +1,171 @@
+use bson::Document;
+use serde::Serialize;
+use wasm_bindgen::prelude::*;
+
+use super::{CollectionInfo, DataService};
+use crate::{Error, Result};
+
+#[wasm_bindgen(start)]
+pub fn init() {
+    #[cfg(feature = "console_error_panic_hook")]
```
nicholascioli: Nit: Does this feature exist? I didn't see it in `Cargo.toml`.

schema-builder-library/src/data_service/mod.rs
nicholascioli: Actually, it might be nice for the error type to be an associated type since I see that the WASm implementation uses a somewhat different error type. (Mostly a scoped variant of the overall error type)

nicholascioli (changes requested):
Looks really good! I left a few questions and nits.

schema-builder-library/src/data_service/wasm.rs
```diff
@@ -0,0 +1,171 @@
...
+    console_error_panic_hook::set_once();
+}
+
+// Embed TypeScript interface definitions directly in the generated .d.ts file.
+#[wasm_bindgen(typescript_custom_section)]
+const TS_INTERFACE: &str = r#"
```
nicholascioli: Nit: It might be nice to have the typescript definitions in a separate file so that syntax highlighting works as expected.


```suggestion
const TS_INTERFACE: &str = include_str!("types.d.ts");
```

nicholascioli (changes requested):
Forgot to add this other comment sorry.

schema-builder-library/src/collection/mod.rs
jchemburkar: ah we can remove the alias. on the first pass I accidentally overloaded "CollectionInfo", hence the changes I made to rename the top level one `DatabaseCollections`. I'll remove the alias now, good catch.

schema-builder-library/src/data_service/mod.rs
jchemburkar: good catch, I updated this after the fact to split rather than taking in a single string

schema-builder-library/src/data_service/mongodb.rs
jchemburkar: I actually don't know or remember the context .... we can/should jsut move to the driver version

schema-builder-library/src/data_service/test.rs
jchemburkar: hmmm fair enough (I think this one was actually added by claude). removing

schema-builder-library/src/data_service/wasm.rs
jchemburkar: I think this also just turned to spaghetti with wasm refactoring when trying to reduce duplication

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub meta_field: Option<String>,
+}
+
+/// Abstraction over database operations used by the schema builder.
+///
+/// Uses namespace strings in `"database.collection"` format for collection operations.
+/// On non-WASM targets `Send + Sync` is required; on WASM it is not.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+pub trait DataService {
```
bucaojit, line 67: In the design you have a `Send + Sync` here, did you decide not to include it?

schema-builder-library/src/data_service/test.rs
```diff
@@ -0,0 +1,159 @@
...
+fn test_missing_options_field_uses_default() {
+    let doc = bson::doc! {
+        "name": "myCollection",
+        "type": "collection"
+    };
+    let info: CollectionInfo = bson::from_document(doc).unwrap();
+    assert_eq!(info.options, CollectionOptions::default());
+}
+
+#[test]
```
bucaojit, line 142: I see there is a change in behavior with how viewOn is handled when missing.  How about a test to show that it deserializes to `""` to document the behavior.

schema-builder-library/src/data_service/mongodb.rs
bucaojit: Well I think we would want `nameOnly: true`, in the options as well, we have that in the ODBC and JDBC drivers whenever we list collections.

bucaojit (commented):
Some initial comments.

schema-builder-library/src/data_service/mod.rs
jchemburkar: Yes, talked about it a little bit with @nicholascioli during design time but the goal is to kick the send/sync part down the road to see if we can handle it in the integration of this trait throughout the library rather than requiring us to duplicate the trait itself

bucaojit (approved):
LGTM

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Additional options (primarily used for views and timeseries).
+    #[serde(default)]
+    pub options: CollectionOptions,
+}
+
+/// Options for collections. View options and timeseries options are both represented here
+/// since the `options` field in `listCollections` output is overloaded for different collection types.
+/// Consumers should check `CollectionInfo::collection_type` before accessing the relevant fields.
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
nicholascioli: Nit: You can just add default here and avoid putting it on each individual field.


```suggestion
#[serde(rename_all = "camelCase", default)]
```

schema-builder-library/src/data_service/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/// Abstraction over database operations used by the schema builder.
+///
+/// On non-WASM targets `Send + Sync` is required; on WASM it is not.
+#[cfg_attr(not(feature = "wasm"), async_trait::async_trait)]
+#[cfg_attr(feature = "wasm", async_trait::async_trait(?Send))]
+pub trait DataService {
+    /// The error type returned by this service's operations.
+    type Error: std::error::Error;
```
nicholascioli: Nit: Most associated error types typically don't have trait bounds on the `Error` type, as it is not implemented by all error types and can be inflexible. (e.g. [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) also has an associated `Error` type but enforces no bounds.

schema-builder-library/src/data_service/mongodb.rs
```diff
@@ -0,0 +1,90 @@
...
+        db_name: &str,
+    ) -> std::result::Result<Vec<CollectionInfo>, Self::Error> {
+        let db = self.client.database(db_name);
+        let cursor = db.list_collections().await?;
+        let specs = cursor.try_collect::<Vec<_>>().await?;
+
+        Ok(specs
+            .into_iter()
+            .filter_map(|spec| {
+                let collection_type = match spec.collection_type {
```
nicholascioli: Nit: you could implement [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) for `CollectionType` and make this a simple fallible call instead.

```rust
impl TryFrom<CollectionInfo> for DriverCollectionType { ... }

// in this function
let collection_type = spec.collection_type.try_into()
    .inspect_err(|_| warn!(...))
    .ok()?;
```

schema-builder-library/src/data_service/types.d.ts
```diff
@@ -0,0 +1,43 @@
...
+        /** For views, the source collection name */
+        viewOn?: string;
+        /** For views, the aggregation pipeline */
+        pipeline?: BsonDocument[];
+        /** For timeseries, the timeseries options */
+        timeseries?: {
+            timeField: string;
+            metaField?: string;
+        };
+    };
```
nicholascioli: Nit: This could probably just be a `Partial` so as to not have to have the `?` on every member.


```typescript
// Somewhere above
export interface TimeSeries {
    timeField: string;
    metaField?: string;
};

export interface CollectionInfoOptions {
    viewOn: string;
    pipeline: Array<BsonDocument>;
    timeseries: TimeSeries;
}

// In this type
export interface CollectionInfo {
    // ...
    options?: Partial<CollectionInfoOptions>;
}
```

schema-builder-library/src/data_service/wasm.rs
```diff
@@ -0,0 +1,142 @@
+use bson::Document;
+use serde::Serialize;
+use wasm_bindgen::prelude::*;
+
+use super::{CollectionInfo, DataService};
+
+/// Error type for [`WasmDataService`] operations.
+#[derive(Debug)]
+pub struct WasmDataServiceError(String);
```
nicholascioli: Nit: It seems that this error has 3 possible causes:

- Serialization
- Database query
- Deserialization

It might be useful for this to be an enum and it implement `thiserror::Error`.

schema-builder-library/src/partitioning/mod.rs
```diff
@@ -30,39 +33,40 @@ pub struct PartitionedCollection {
...
     let size_info = get_size_counts(collection).await?;
     let num_partitions = get_num_partitions(size_info.size, PARTITION_SIZE_IN_BYTES) as usize;
 
     // For timeseries collections, there is no `count` field reported, so we sample at a rate of
     // 1 per # of partitions. This is likely a much higher sample rate than for non-timeseries
     // collections, but it is our best effort for now to avoid running a full collection count.
-    let (sample_rate, partition_key, hint) = if collection_doc.type_ == "timeseries" {
-        let sample_rate = 1f64 / num_partitions as f64;
+    let (sample_rate, partition_key, hint) =
+        if collection_doc.collection_type == CollectionType::Timeseries {
```
nicholascioli: Nit: This might be better as an explicit `match` statement so that if new collection types are added down the line which also do not support counting, this doesn't cause them to silently fail.

nicholascioli (approved):
Looks great! I left a few nits, but honestly they are more stylistic so feel free to ignore.

PR #137, opened 2026-04-13
[SQL-3183]Update README to mention the tool is only for On-Prem Enterprise cust…

jpowell-mongo:
Update README to mention schema-builder CLI is for enterprise customers only.


nbagnard (approved):
LGTM! :)

PR #139, opened 2026-04-17
SQL-3168: add WASM compile task for schema-builder-library

jchemburkar:
for now, adding a skip reason to allow work integrating interface to happen.


jpowell-mongo (approved):

evergreen.yml
```diff
@@ -155,6 +155,33 @@ functions:
           cargo build --release --features "mongosql-c/test" --target-dir "test_target"
           cargo build --release
 
+  "compile schema-builder-library":
+    - command: shell.exec
+      type: test
+      retry_on_failure: true
```
nicholascioli, line 161: Is this because rustup might time out or something?

nicholascioli (approved):
This looks good! I just had a quick question, but it shouldn't be a blocker.

evergreen.yml
jchemburkar: I actually think it was just boilerplate copy/paste, though I think its probably fine to keep (unless you feel strongly). will merge as is for now.

PR #140, opened 2026-04-20
SQL-3200 SQL-3207: bump rand, tonic, rustls-webpki, and opentelemetry crates

nicholascioli:
This PR bumps versions of the `rand`, `tonic`, and `opentelemetry` crates while also disabling default features of crates that don't use it. The bulk of this PR is updating the function signatures of all of the opentelemetry primitives that changed since the last version in use.

I've also updated the generation of proto files into a build script so as to automatically compile it in when building the project, since tonic was also updated to remove references to vulnerable rand versions.


nicholascioli:
I also removed some dependencies listed as unused by `cargo-machete`.

bucaojit:
Offline you mentioned there was a Rust crate that we can use that will build `protoc` during compile time so it won't be needed to be installed as a prerequisite on the system.  Is that being used here?  The evergreen builds are failing with this patch because they can't find the `protoc` binary.

nicholascioli:
I've also updated `rust-webpki` since it was causing a failure in this PR, but let me know if it should be in its own PR instead!

bucaojit (approved):
LGTM. 
The `rust-webpki` is fine here.  That way the evergreen build will pass.

service/Cargo.toml
```diff
@@ -17,40 +19,42 @@ path = "src/main.rs"
...
+protoc-bin-vendored = "3.2.0"
+tonic-prost-build = "0.14.3"
+
 [features]
 integration = []
+
+[package.metadata.cargo-machete]
+# The following crates are used in the autogenerated proto files, which
+# confuses cargo-machete.
+ignored = ["prost", "tonic-prost"]
```
nbagnard: [nit] Can you add a new line at the end of the file.

Cargo.lock
```diff
@@ -4158,7 +3955,7 @@ name = "test-utils"
 version = "0.1.0"
 dependencies = [
  "agg-ast",
- "itertools 0.14.0",
+ "itertools 0.11.0",
```
nbagnard, line 3958: Do you know why this was downgraded?

nbagnard (approved):
LGTM!

Before merging, create a JIRA for rutls-webpki and add the JIRA id to the PR title here. This way it will be easy to know what the changes were when doing the release.

Cargo.lock
nicholascioli: Hmm, that is a good question. I'll double check and see

nicholascioli: Ah, I see. We depend on itertools version `0`, which means that we basically don't care about the patch version. One of the unused dependencies removed depended on `0.14`, which also matched our constraints, so it was at that version previously.

nicholascioli: Now the newest version of itertools in our dependency tree is at `0.11`, which means that we are also at that version.

PR #143, opened 2026-04-27
SQL-3165: Integrate DataService

nicholascioli:
This PR integrates the previously defined [`DataService`](#136) trait into all aspects of the schema manager library. With these changes in place, work can now begin on targeting WASM platforms and changing some of the existing integration tests into unit tests for the library.


schema-builder-library/src/options.rs
```diff
@@ -11,12 +13,12 @@ pub struct BuilderOptions {
     /// Whether to perform a dry run, i.e. no analysis and no writing to the database
     pub dry_run: bool,
     /// The MongoDB client
-    pub client: mongodb::Client,
+    pub service: S,
     /// Task sempahore
```
jchemburkar:
```suggestion
    /// Task sempaphore
```

schema-builder-library/src/errors.rs
```diff
@@ -18,24 +18,16 @@ pub enum Error {
...
+    #[error("Error in DataService: {0}")]
+    DataServiceError(ServiceError),
     #[error("Schema Error {0}")]
     SchemaError(mongosql::schema::Error),
     #[error("NoCollection {0}")]
     NoCollection(String),
-    #[cfg(feature = "native-client")]
-    #[error("Execution Error {0}")]
-    TokioError(tokio::task::JoinError),
     #[error("Inital schema for {0} is not valid")]
```
jchemburkar:
```suggestion
    #[error("Initial schema for {0} is not valid")]
```

schema-builder-library/src/options.rs
```diff
@@ -11,12 +13,12 @@ pub struct BuilderOptions {
     /// Whether to perform a dry run, i.e. no analysis and no writing to the database
     pub dry_run: bool,
     /// The MongoDB client
```
jchemburkar:
```suggestion
    /// The backend data service used to connect to MongoDB
```

something like this maybe?

schema-builder-library/src/lib.rs
```diff
@@ -189,7 +192,7 @@ pub async fn build_schema(options: BuilderOptions) -> Result<ResultSet> {
 
     // To start computing the schema for all databases, we need to wait for the
     // list_database_names method to finish.
```
jchemburkar: Just noting here and two lines below func name has been updated

jchemburkar (commented):
I did a quick scan (hence all the comment nitpicks lol) and then looked through mod.rs, I'll pick up the rest tomorrow morning, but so far looks great, and awesome that it looks like tests are still passing!

I also wanted to note two followup tickets that we can consider for some of the work we discussed earlier:
- [add single threaded feature support](https://jira.mongodb.org/browse/SQL-3167): I think we can use this one for the express task of working on any additional changes needed to target wasm and get it to build properly (haven't verified myself yet)
- [verify library still works with poc](https://jira.mongodb.org/browse/SQL-3170): I think this can be an investigation / cover any additional work to get the non-tokio futures working with wasm

basically just wanted to confirm I think this ticket is well scoped after thinking more about our convo earlier. Nice work from what I've seen so far 😁

schema-builder-library/src/context.rs
```diff
@@ -0,0 +1,26 @@
+use std::sync::Arc;
+
+use crate::DataService;
+
+/// Handle to a context
+pub type ContextHandle<S> = Arc<Context<S>>;
+
+/// Execution Context
+///
+/// This struct contains all needed handles for executing queries against
```
jchemburkar, line 10: very neat

schema-builder-library/src/lib.rs
```diff
@@ -352,72 +343,17 @@ async fn fetch_initial_schema(db: Database, collection: &str) -> Result<Catalog>
...
+        fetch_initial_schema(service, db, collection)
+            .await
+            .map(|catalog| (db.clone(), catalog))
     }))
     .await
     .into_iter()
     .collect()
 }
-
-async fn process_collection_tasks(
```
jchemburkar, line 374: kinda nice that we can get rid of these wrappers

jchemburkar (approved):
lgtm!

schema-builder-library/src/collection/mod.rs
```diff
@@ -39,50 +38,39 @@ static INCLUDE_LIST_IN_DB_AND_COLL_PAIRS: OnceLock<Vec<(String, String)>> = Once
...
-) -> Result<HashMap<String, (Schema, bool)>> {
-    let mut initial_collection_schemas: HashMap<String, (Schema, bool)> = HashMap::new();
-
-    let coll = database.collection::<Document>(schema_collection);
-    let mut cursor = coll.find(doc! {}).await?;
+pub(crate) async fn query_for_initial_schemas<S: DataService>(
+    service: &S,
+    db: &str,
+    collection: &str,
+) -> Result<HashMap<String, (Schema, bool)>, S::Error> {
```
jpowell-mongo: Suggestion (non-blocking): Could we consider creating a type for the (Schema, bool) instead of returning a tuple? 

Not particular about the naming, but possibly: 
```rust
struct QueryForInitialScehamResult {
    schema: Schema
    is_schema_unstable: bool
}
```

schema-builder-library/src/lib.rs
```diff
@@ -187,32 +190,29 @@ pub async fn build_schema(options: BuilderOptions) -> Result<ResultSet> {
...
-            // so we have to manually filter out unwanted databases here.
-            let valid_databases = databases
-                .into_iter()
-                .filter(|db| !DISALLOWED_DB_NAMES.contains(&db.as_str()))
-                .collect();
+    // To start computing the schema for all databases, we need to wait for the list of
+    // available databases.
+    //
+    // Note: The built-in mongodb `filter` option doesn't work for Atlas Free and
+    // Shared tier clusters, so we have to manyally filter out unwanted databases here.
```
jpowell-mongo: nit (non-blocking): spelling
```suggestion
    // Shared tier clusters, so we have to manually filter out unwanted databases here.
```

schema-builder-library/src/lib.rs
nicholascioli: Yeah honestly I'm all for removing duplicative code when doing new work

schema-builder-library/src/collection/mod.rs
```diff
@@ -39,50 +38,39 @@ static INCLUDE_LIST_IN_DB_AND_COLL_PAIRS: OnceLock<Vec<(String, String)>> = Once
...
-    database: &Database,
-) -> Result<HashMap<String, (Schema, bool)>> {
-    let mut initial_collection_schemas: HashMap<String, (Schema, bool)> = HashMap::new();
-
-    let coll = database.collection::<Document>(schema_collection);
-    let mut cursor = coll.find(doc! {}).await?;
+pub(crate) async fn query_for_initial_schemas<S: DataService>(
+    service: &S,
+    db: &str,
+    collection: &str,
```
nbagnard:
```suggestion
    schema_collection: &str,
```
(non-blocking )I think it is still good to keep the variable name as is to clearly convey that we are getting the info from the collection used to store schema.

schema-builder-library/src/collection/patterns.rs
```diff
@@ -1,13 +1,20 @@
...
 use crate::data_service::{CollectionInfo, CollectionType};
-use crate::{Error, Result, consts::DISALLOWED_COLLECTION_NAMES};
-use futures::TryStreamExt;
-use mongodb::{
-    Cursor,
-    bson::{self, Document},
-};
 use tracing::instrument;
 
+/// An error in a database pattern
```
nbagnard:
```suggestion
/// An error in a namespace pattern
```

schema-builder-library/src/collection/mod.rs
nicholascioli: I actually looked at this a bit more and realized it should have been contained in the initial schema. This simplified the signature.

schema-builder-library/src/collection/mod.rs
```diff
@@ -39,50 +38,32 @@ static INCLUDE_LIST_IN_DB_AND_COLL_PAIRS: OnceLock<Vec<(String, String)>> = Once
...
+) -> Result<HashMap<String, Schema>, S::Error> {
+    let docs = service
+        .find(db, collection, doc! {})
```
nbagnard: issue:
I am a bit concerned about this. If a database contains many collections, the user ran the schema manager for all of them at some point, the schemas are big, then we are going to increase the memory consumption notably by storing both `docs` and `InitialSchema` until the end of the function.
Before, only one `doc` at a time was hold thanks to the cursor. 
The node.js mongo driver seems to be supporting returning a cursor for streaming : https://www.mongodb.com/docs/drivers/node/current/crud/query/retrieve/
I think all drivers do.
It would be safer to go back to the service design and incorporate cursor everywhere possible.

schema-builder-library/src/partitioning/mod.rs
```diff
@@ -150,41 +146,49 @@ pub(crate) struct CollectionSizes {
...
-            _ => {
-                return Err(Error::BsonFailure);
-            }
-        };
-        return if size == 0 || count.is_some_and(|c| c == 0) {
-            Err(Error::EmptyCollection(collection.name().to_string()))
-        } else {
-            Ok(CollectionSizes { size, count })
-        };
+        .map_err(|_| Error::NoCollectionStats(collection.to_string()))?;
```
nbagnard: issue:
| think this hides the underlying issue (Error::BsonFailure) that was popping up before when an error happened.

Maybe we can add the root cause as a warning? Or we could fold it into `NoCollectionStats` if that makes more sense.
```suggestion
         .inspect_err(|e| warn!("collStats aggregate failed: {e}"))
        .map_err(|_| Error::NoCollectionStats(collection.to_string()))?;
```

schema-builder-library/src/data_service/mod.rs
```diff
@@ -83,6 +83,7 @@ pub trait DataService {
         db_name: &str,
         coll_name: &str,
         pipeline: Vec<Document>,
+        key_hint: Option<Document>,
```
nbagnard: issue:
Rather than starting with only exposing one option, we should expose all options from the beginning.
This will allow the API to be more robust and not break if we ever need to expose/use other aggregation options.

question:
Is it better to only support `hint` to start with and silently discard whatever else is passed in or should we allow all options to be supported? I'm thinking that maybe implementer of the Data Service might want to be able to support `max_time` or `allow_disk_use` as well for example.

schema-builder-library/src/schema/mod.rs
```diff
@@ -217,22 +202,22 @@ pub(crate) async fn derive_schema_for_partition(
...
 ) -> Option<Schema> {
     let pipeline = vec![doc! { "$sample": { "size": VIEW_SAMPLE_SIZE } }]
         .into_iter()
         .chain(view.options.pipeline.clone().into_iter())
         .collect::<Vec<Document>>();
 
-    let Ok(mut cursor) = database
-        .collection::<Document>(&view.options.view_on)
-        .aggregate(pipeline)
+    let Ok(cursor) = service
```
nbagnard:
```suggestion
    let Ok(sampledDocs) = service
```

Let's not call it cursor to avoid hidden the fact that the full resultset is returned back as a vector.

schema-builder-library/src/schema/mod.rs
```diff
@@ -242,22 +227,20 @@ pub(crate) async fn derive_schema_for_view(
...
-    while let Some(Ok(doc)) = cursor.try_next().await.transpose() {
+    for (iterations, doc) in cursor.iter().enumerate() {
         // Notify every 100 iterations, so it isn't too spammy
         if iterations.is_multiple_of(100) {
             info!(
-                db = database.name(),
+                db,
                 view_name = view.name,
                 iteration = iterations,
                 "Sampling view"
```
nbagnard: nitpick:
```suggestion
                "Calculating schema for sampled view"
```

schema-builder-library/src/client_util.rs
```diff
@@ -76,35 +84,41 @@ pub async fn get_opts(uri: &str, resolver: Option<ResolverConfig>) -> Result<Cli
...
 
     async fn run_cursor_command_with_read_preference(
         &self,
         command: Document,
-    ) -> Result<Cursor<Document>>;
+    ) -> Result<Cursor<Document>, <MongoDbDataService as DataService>::Error>;
 }
 
 #[async_trait]
 impl DatabaseExt for Database {
```
nbagnard, line 94: issue:
Right now, this is dead code.
I tracked it down, and I think that the issue stems from the implementation of `list_collections` in the `mongodb.rs`. The implementation needs to be reverted to use `run_command` and `run_command` need to be the extended version Ethan crafted otherwise the read preference is not applied.

We switched to use the Rust function "listCollection" and we will need to roll that back.
We discovered the issue when implementing the ODBC driver:  `listCollection` does not filter authorized collections only.
The command itself lets us  control exactly what is executed rather than rely on the Rust driver "magic".
Right now, `listCollection` can be chained with `authorized_collections`, but the documentation says "When used with list_collections this option has no effect." <--  And here you get funny surprises.

schema-builder-library/src/data_service/mod.rs
```diff
@@ -66,7 +66,7 @@ pub struct TimeSeriesOptions {
 #[cfg_attr(feature = "wasm", async_trait::async_trait(?Send))]
 pub trait DataService {
     /// The error type returned by this service's operations.
-    type Error;
+    type Error: core::error::Error;
 
     /// List all database names.
     async fn list_databases(&self) -> std::result::Result<Vec<String>, Self::Error>;
```
nbagnard: thought:
I would love the service methods name to map more closely to the client libraries methods name.
```suggestion
    async fn list_database_names(&self) -> std::result::Result<Vec<String>, Self::Error>;
```

I'm sorry this is coming out now since it is not directly related to your changes, but it might be worth doing now rather than later.

nbagnard (changes requested):
Use the DataService is neatly incorporated, but it surfaced a few things in the DataService or overall design that we should look into.

schema-builder-library/src/data_service/types.d.ts
```diff
@@ -52,3 +52,14 @@ export interface SqlDataService {
...
+/**
+ * SqlCursor interface for streaming database results
+ *
+ * Implement this interface in your JavaScript/TypeScript code to provide a
+ * way to stream over database results
+ */
+export interface SqlCursor {
+    /** Get the next element in this cursor. Return undefined to signal the end of the stream */
+    next(): Promise<BsonDocument | undefined>;
+}
```
nbagnard: todo: (non-blocking)
Add missing new line at the en of the file

nbagnard (changes requested):
The cursor change looks good. I'm happy to see that it was pretty straightforward :)

But, please address at the very minimum comments tagged `issue:`. If something is deferred to a follow-up ticket, reply to the comment with the ticket ID.

For `thought:` and `nitpick:`, they are non-blocking by nature, though they are quick win and worth incorporating while making changes.

schema-builder-library/src/collection/mod.rs
nicholascioli: This was resolved by switching to streaming / using a cursor.

schema-builder-library/src/partitioning/mod.rs
nicholascioli: I decided to match the behaviour in other fallible paths and return a `DataServiceError` instead

schema-builder-library/src/data_service/mod.rs
nicholascioli: Hmm, this is a good question. We only ever use the `key_hint` in the entirety of our codebase, so I figured that was a good one to have passed in. Clients shouldn't be able to affect the options passed in, so it's closed in that sense. What I can do for now is make this take in an `Options` struct with just the key hint for now so that we can add more options as we see the need.

schema-builder-library/src/schema/mod.rs
nicholascioli: We swapped back to cursor, so I've kept it as is to match the previous behaviour.

schema-builder-library/src/data_service/types.d.ts
```diff
@@ -36,6 +36,14 @@ export interface CollectionInfo {
     options?: Partial<CollectionInfoOptions>;
 }
 
+/**
+ * Options for aggregate queries
+ */
+export interface AggregateOptions {
+    /** A hint for which key to use for indexing */
+    key_hint: BsonDocument,
```
nbagnard: (non-blocking)
Mirrors the Rust `Option<Document>` field
```suggestion
    key_hint?: BsonDocument,
```

schema-builder-library/src/data_service/types.d.ts
```diff
@@ -48,7 +56,18 @@ export interface SqlDataService {
     /** List all collections in a database */
     listCollections(dbName: string): Promise<CollectionInfo[]>;
     /** Execute an aggregation pipeline on a collection */
-    aggregate(dbName: string, collName: string, pipeline: BsonDocument[]): Promise<BsonDocument[]>;
+    aggregate(dbName: string, collName: string, pipeline: BsonDocument[], options: Partial<AggregateOptions>): Promise<BsonDocument[]>;
```
nbagnard: todo:
Stale definition.

```suggestion
    aggregate(dbName: string, collName: string, pipeline: BsonDocument[], options: Partial<AggregateOptions>): Promise<SqlCursor>;
```

schema-builder-library/src/data_service/types.d.ts
```diff
@@ -48,7 +56,18 @@ export interface SqlDataService {
...
     /** Execute a find query on a collection */
     find(dbName: string, collName: string, filter: BsonDocument): Promise<BsonDocument[]>;
```
nbagnard: todo:
Stale definition.
```suggestion
    find(dbName: string, collName: string, filter: BsonDocument): Promise<SqlCursor>;
```

schema-builder-library/src/data_service/mongodb.rs
```diff
@@ -32,39 +37,33 @@ impl MongoDbDataService {
...
-                    }),
-                };
-                Some(CollectionInfo {
-                    name: spec.name,
-                    collection_type,
-                    options,
-                })
+            .filter_map(|doc| {
+                bson::from_document(doc)
+                    .map_err(|e| warn!("Skipping malformed listCollections entry: {e}"))
```
nbagnard: suggestion: (non-blocking)
Capture the collection name (if present) before deserialization so that the skip warning identifies the malformed collection and makes debugging easier.

```suggestion
               let name = doc.get_str("name").unwrap_or("<unknown>").to_string();
                bson::from_document(doc)
                    .map_err(|e| warn!("Skipping malformed listCollections entry '{name}': {e}"))
```

schema-builder-library/src/partitioning/mod.rs
```diff
@@ -88,32 +87,38 @@ pub(crate) async fn get_partitions(
...
+    let Ok(cursor) = service
+        .aggregate(
+            db,
+            &collection_info.name,
+            sample_pipeline,
+            AggregateOptions {
+                key_hint: hint.clone(),
+            },
+        )
+        .await
```
nbagnard, line 107: todo:
Log the underlying error before falling through to the single-partition path.
Otherwise the failure is dropped silently and there is no signal that partitioning failed and we had to fallback.
```suggestion
        .await
        .inspect_err(|e| {
            tracing::warn!(
                db,
                collection = %collection_info.name,
                "partition sampling failed; falling back to a single partition: {e}"
            )
        })

```

schema-builder-library/src/schema/mod.rs
```diff
@@ -217,46 +211,51 @@ pub(crate) async fn derive_schema_for_partition(
...
         })
-    else {
-        return None;
-    };
+        .ok()?;
 
     let mut schema = None;
     let mut iterations = 0u64;
+    let mut cursor = Box::pin(cursor);
     while let Some(Ok(doc)) = cursor.try_next().await.transpose() {
```
nbagnard: todo:
Let's log that there was an issue retrieving sampling results and we exited the loop early.
```suggestion
    while let Some(Ok(doc)) = cursor.try_next().await.inspect_err(|e| {
          warn!(db, view_name = view.name, iteration = iterations,                                                                                                                                                                                                                                                                                             
                "view sampling encountered an error: {e}")
      }) transpose() {
```

schema-builder-library/src/data_service/mod.rs
```diff
@@ -59,37 +60,42 @@ pub struct TimeSeriesOptions {
     pub meta_field: Option<String>,
 }
 
+/// Options for aggregate queries
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct AggregateOptions {
```
nbagnard, line 67: question:
AggregateOptions is also the name of the struct used by the Rust driver I believe, no?
Could that lead to some kind of shadowing issue?

schema-builder-library/src/data_service/mod.rs
```diff
@@ -59,37 +60,42 @@ pub struct TimeSeriesOptions {
     pub meta_field: Option<String>,
 }
 
+/// Options for aggregate queries
+#[derive(Debug, Default, Serialize, Deserialize)]
```
nbagnard, line 64: todo:
- Mark it as "non_exhaustive" to prepare for extensibility.
- I think you want to add the rename_all def similar to other struct.
```suggestion
#[derive(Debug, Default, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
#[non_exhaustive]
```

schema-builder-library/src/data_service/types.d.ts
nbagnard: todo:
Keep field name casing consistent with the other options and following Node.ja naming convention. 
```suggestion
    keyHint: BsonDocument,
```

schema-builder-library/src/data_service/wasm.rs
```diff
@@ -102,40 +107,68 @@ impl DataService for WasmDataService {
...
+            js_cursor,
+            |cursor| async move {
+                let next = cursor
+                    .next()
+                    .await
+                    .map_err(|e| WasmDataServiceError::Query(format!("{e:?}")))?;
+                let deserialized = serde_wasm_bindgen::from_value(next)
+                    .map_err(|e| WasmDataServiceError::Deserialization(e.to_string()))?;
+
+                Ok(Some((deserialized, cursor)))
```
nbagnard: todo:
JS returns `undefined` to signal end-of-stream, serde_wasm_bindgen cannot deserialize undefined into a Document. It returns a deserialization error. The ? propagates that error, and the stream terminates with Err(Deserialization(...)) instead of Ok(None). The stream never ends cleanly.


```suggestion
                let deserialized: Option<Document> = serde_wasm_bindgen::from_value(next)
                    .map_err(|e| WasmDataServiceError::Deserialization(e.to_string()))?;

                Ok(deserialized.map(|doc| (doc, cursor)))
```

nbagnard (changes requested):
This is a solid! 
I think once you address the small todo: we should be good to go 🚀 

A few small updates were missed when switching to cursor for aggregate(). The end of stream handling in particular.
Otherwise, using a struct Option seems good to me.
I noticed a few places were logging and/or error hanlding could be improved too and flag them too.

schema-builder-library/src/data_service/types.d.ts
nicholascioli: Oh, I made this non optional because the signature for the aggregate function takes in a `Partial<AggregateOptions>`. I also configured the Rust wasm side of things to remove anything that is `None` to match this behaviour.

schema-builder-library/src/data_service/mod.rs
nicholascioli: Only if we import both in the same context. I can rename it, if you prefer!

schema-builder-library/src/data_service/types.d.ts
nbagnard: suggestion:
(non-blocking but worth thinking about)
I would prefer the type definition itself to reflect the fact that fields are not mandatory.
You can move partial on the export `export type AggregateOptions = Partial<{ .. }` and remove the partial wrapper from the function signature.
Anyway, it looks like because in wasm.rs you have `serialize_missing_as_null(true)` which will always include { keyHint: null }, the Partial flag isn't really making a difference right now. 
If we want a clear type definition, we can either update the Serde side and include `skip_serializing_if = "Option::is_none` on `key_hint` or remove `Partial` and change `KeyHint:BsonDocument` to `KeyHint:BsonDocument|null`.

schema-builder-library/src/data_service/mod.rs
nbagnard: I think that's fine, it's probably not likely to happen soon and easy to work-around.

nbagnard (approved):
LGTM!
For AggregationOption type definition, I think that even though things work right now, it would help to have a clear contract so that people are not left guessing (maybe incorrectly) what exactly we send.

PR #144, opened 2026-04-28
[SQL-2712] Update IN translation to use $in

jpowell-mongo:
This change updates mongosql to preserve the IN operator in SQL queries rather than translate them into a disjunction of ORs. Index Utilization is preserved.

Before, our rewriters would translate the query to a disjunction of ORs:
```sql
-- original query
SELECT a IN (b, c, d)
-- query after rewrites
SELECT a = b OR a = c OR a = d
```

Now, the code will preserve the IN operator, and later translate the code to the corresponding mql operator:
```sql
-- original query
SELECT a IN (b, c, d)
-- query after rewrites
SELECT a IN (b, c, d)
```

Along with these changes, I've made changes throughout the rest of the code, to allow IN / NOT IN to be translated to $in and { "$not": { "$in": [...] } } expressions in match stage aggregation pipelines, and use $in and $nin in $expr expressions. 

```bash
./target/debug/mongosql-cli "SELECT * FROM nullable_fields WHERE a IN (100, 250, 322)"
```

Will produce this aggregation pipeline:
```js
[
    { "$match": { "a": { "$in": [100, 250, 322] } } },
    { "$project": { "nullable_fields": "$$ROOT", "_id": 0 } },
]
```

## Changes
- Introduce `In` and `NotIn` enums to mir, air, ast, and codegen codepaths
- Introduce a `Tuple` enum type throughut mir, air, and codegen. It operates the same as the Array type but I felt like it made it clearer to understand the cases. 
- Remove the InTupleRewriteVisitorPass, update the SingleTupleRewriter with a special case to not unwrap single element IN expressions (ex: SELECT * FROM orders where status IN ('pending'))
- General threading of the new In, NotIn, and Tuple logic through the codebase

## Testing

- Unit Tests added at each step of the compiler, mir, air, codegen.
- Index Utilization test to validate index usage still works.
- Validated that the SQL from related Looker ticket works as expected.
- Manual testing to validate $in expressions are used.


mongosql/src/mir/definitions.rs
```diff
@@ -1084,6 +1108,21 @@ pub enum MatchLanguageComparisonOp {
...
 }
 
+#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+pub enum MatchLanguageInOp {
+    In,
+    NotIn,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct MatchLanguageIn {
```
jpowell-mongo, line 1099: TODO: Add comments about what each represents.

$in expression has a field_path + a list of values.

example: 

"pull_request.status": { "$in": ["draft","pending", "request_changes","merged"]}

mongosql-cli/src/main.rs
```diff
@@ -362,7 +362,7 @@ mod tests {
     #[test]
     fn empty_sql_file_returns_error() {
         let query: Option<String> = None;
-        let sql_file = Some(PathBuf::from("./test/empty_query.sql"));
+        let sql_file = Some(PathBuf::from("./test/empty_file.sql"));
```
jpowell-mongo, line 365: Note: This was a driveby fix. I realized the file name didn't match what's in the test directory

tests/index_usage_tests/filter.yml
```diff
@@ -204,3 +204,7 @@ tests:
     current_db: index_usage_filter
     query: "SELECT * FROM nullable_fields WHERE a > 100 AND b IN (SELECT b FROM non_nullable_fields WHERE a < 300)"
     expected_utilization: IX_SCAN
+  - description: in operator uses the index
```
mattChiaravalloti, line 208:
```suggestion

    - description: in operator uses the index
```
[nit] All other tests in this file have a newline between them, so I suggest adding one here as well

tests/index_usage_tests/filter.yml
mattChiaravalloti, line 1: [Questions/Thoughts] Should we also add any of the following tests?
- `NOT IN` for `nullable_fields` should use an `IX_SCAN`
- `IN` for `non_nullable_fields` _should_ use an `IX_SCAN` I think... you can check that against mongodb manually and then add the test if you observe index utilization is possible in this case (I suspect it is)
- `NOT IN` for `non_nullable_fields` _should_ use an `IX_SCAN` (again, check this and if it does then we should add the test so we don't regress on this support in the future)

mongosql/src/ast/pretty_print.rs
```diff
@@ -155,6 +157,9 @@ impl PrettyPrint for Query {
 
 impl PrettyPrint for WithQuery {
     fn pretty_print(&self) -> Result<String> {
+        if self.queries.is_empty() {
+            return Err(Error::EmptyWithQueryNamedQueries);
```
mattChiaravalloti, line 161: [Question] Wow did we finally encounter a fuzz test failure after (probably) millions of runs? That makes me so happy if so. Or did you just notice that this was a possibility on your own and fix this? Also, to be clear, this is a drive-by fix and not something directly relevant to this PR, right?

mongosql/src/ast/pretty_print_fuzz_test.rs
```diff
@@ -665,10 +670,21 @@ mod arbitrary {
 
     impl Arbitrary for BinaryExpr {
         fn arbitrary(g: &mut Gen) -> Self {
+            let op = BinaryOp::arbitrary(g);
+            // The parser always produces a Tuple on the RHS of IN/NOT IN, so arbitrary
+            // generation must match that shape to satisfy the round-trip property.
+            let right = if op == BinaryOp::In || op == BinaryOp::NotIn {
```
mattChiaravalloti, line 676: [Praise] Smart!

mongosql/src/ast/rewrites/mod.rs
```diff
@@ -31,7 +31,6 @@ pub use with_query::WithQueryRewritePass;
 
 #[cfg(test)]
 mod test;
-
```
mattChiaravalloti, line 34: [Question] Is there a reason you got rid of this newline?

mongosql/src/ast/rewrites/test.rs
```diff
@@ -778,6 +778,35 @@ mod single_tuple {
         expected = Ok("SELECT * FROM (SELECT (a, b)) AS z"),
         input = "SELECT * FROM (SELECT (a, b)) AS z",
     );
+    // The RHS of IN/NOT IN is semantically an array and must not be collapsed, even when
+    // it contains a single element.  Collapsing (true) → true would produce an invalid
+    // second argument for $in at codegen time.
```
mattChiaravalloti, line 692: [Thought] I wonder if we should consider rewriting these cases to `=` and `<>`. We could investigate the performance differences between `{$match: {a: {$eq: X}}}`(/`{$match: {a: {$ne: X}}}`) and `{$match: {a: {$in: [X]}}}`(/`{$match: {a: {$not: {$in: [X]}}}}`) and only commit to doing this rewrite if it's significant. Might be better as follow-up work, though, if you want to file a separate ticket to investigate this.

mongosql/src/ast/rewrites/mod.rs
```diff
@@ -76,7 +75,6 @@ pub trait Pass {
 pub fn rewrite_query(query: ast::Query) -> Result<ast::Query> {
     let passes: Vec<&dyn Pass> = vec![
         &ExtendedUnwindRewritePass,
-        &InTupleRewritePass,
```
mattChiaravalloti, line 79: [Question] Should we delete the `InTupleRewritePass` (along with its tests) if we no longer use it?

mongosql/src/mir/definitions.rs
```diff
@@ -262,6 +262,7 @@ pub enum Expression {
     Subquery(SubqueryExpr),
     SubqueryComparison(SubqueryComparison),
     TypeAssertion(TypeAssertionExpr),
+    Tuple(TupleExpr),
```
mattChiaravalloti: [Issue] **(blocking)** I know you mentioned in the PR summary that you preferred having a `Tuple` type separate from `Array`, but can you explain the reasoning behind this? How is `Tuple` functionally different from an `Array` in the `mir`?

I have another comment later for the `air::definitions::Expression::Tuple` type where we should **definitely** not introduce that type since `air` is a representation of aggregation language and tuples don't exist in agg.

Generally, looking through this whole PR, it appears to me that `Tuple` is an exact copy of `Array` so I don't think there is any justification for keeping it past the syntax layer (`ast`). We should be algebrizing `ast::Expression::Tuple`s into `mir::Expression::Array`s because that is what they are, abstractly, in our model.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1769,6 +1771,12 @@ impl ScalarFunction {
                 Ok(Schema::Atomic(Atomic::Date))
             }
             MergeObjects => self.schema_check_merge_objects(state, arg_schemas),
+            In | NotIn => self.propagate_variadic_null_arguments(
+                state,
+                arg_schemas,
+                Schema::Any,
+                Schema::Atomic(Atomic::Boolean),
+            ),
```
mattChiaravalloti: [Issue] **(blocking)** This is incorrect for a few reasons.

First, `In`/`NotIn` are not variadic -- they are fixed length with 2 arguments `left (NOT) IN right`. The LHS arg is expected to be `Any`, and the `RHS` _must_ be an `Array`.

Second, in addition to saying the return type is `Boolean` along with a propagated `Null` if any of the args are null, we need to assert that the operands are comparable. As in, we need to ensure that the LHS arg is comparable with each element of the RHS (which must be an array).

You could create a method similar to (or possibly modify/reuse) the `get_comparison_schema` method (we use this for `Lt`/`Lte`/`Gt`/`Gte`/`Eq`/`Neq` a few lines above this). We could possibly modify `get_comparison_schema` to accept the pair of `required_schemas` (for the binary comparison operators it is still `(Any, Any)`, but for `In`/`NotIn`, it should be `(Any, Array)`). In addition to that modification, we'd also need to compare the LHS with the elements of the RHS. It's this part that makes me think we should create a new helper method as opposed to modifying the existing `get_comparison_schema`. Continue reading for my preferred idea.

You can create something specific such as `fn get_in_comparison_schema(&self, state: &SchemaInferenceState, arg_schema: &[Schema]) -> Result<Schema, Error>` which would do the steps I described at the beginning. I'll summarize again here to make it clear:
1. Assert that there are 2 arguments
2. Assert that the left is Any and the right is Array<Any>
3. If the left is nullable _or_ the inner value of right's Array is nullable (e.g. `Array<AnyOf(Int, Null)>`), then be sure to propagate `Null` in the final output
4. Assert that the left argument is comparable with the inner value of right's Array schema (e.g., for `a::!LONG IN (1, 2, 3)` we'd see that the left arg is an `LONG` and the right is an `Array<Int>` and we'd want to assert that `LONG` is comparable with `INT` (using `state.check_comparable_with`)
5. If all of that is okay, return the expected schema (possibly nullable Boolean)
6. Otherwise, return whatever error we encounter along the way via `?` or manually return an `InvalidComparison` error if `check_comparable_with` returns false

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -3476,3 +3476,61 @@ mod neg {
...
+                        Expression::Literal(LiteralValue::Integer(2)),
+                        Expression::Literal(LiteralValue::Integer(3)),
+                    ],
+                }),
+            ],
+            is_nullable: true,
+        }),
+        schema_env = map! {
+            ("foo", 1u16).into() => Schema::Document(Document {
+                keys: map! { "a".into() => Schema::Atomic(Atomic::String) },
```
mattChiaravalloti:
```suggestion
                keys: map! { "a".into() => Schema::Atomic(Atomic::Integer) },
```
[Issue] **(blocking)** As written, this test is technically invalid since we cannot compare a `String` to an `Integer`. This passes for what is currently implemented but will fail for the corrected implementation. I advise updating these test.

We should also have a test showing it is invalid to do this comparison you have here, so you could actually copy this test and change the name/expectation to document we expect it to error (InvalidComparison).

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -3476,3 +3476,61 @@ mod neg {
...
+                required: set! {},
+                additional_properties: false,
+                ..Default::default()
+            })
+        },
+    );
+
+    test_schema!(
+        basic_in_operator_schema_is_boolean,
```
mattChiaravalloti: [nit] Name does not match the test -- seems like this should be `not_inoperator_schema_is_boolean` (or similar)

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -3476,3 +3476,61 @@ mod neg {
...
+                    array: vec![
+                        Expression::Literal(LiteralValue::Integer(1)),
+                        Expression::Literal(LiteralValue::Integer(2)),
+                        Expression::Literal(LiteralValue::Integer(3)),
+                    ],
+                }),
+            ],
+            is_nullable: true,
+        }),
+        schema_env = map! { ("array", 0u16).into() => ANY_ARRAY.clone() },
```
mattChiaravalloti: Is a schema env even necessary for this test?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1714,8 +1718,10 @@ impl<'a> Algebrizer<'a> {
 
             Comparison(_) => self.algebrize_binary_comparison_operands(*b.left, *b.right)?,
 
-            // In and NotIn should have been rewritten during ast rewriting.
-            In | NotIn => panic!("'{}' cannot be algebrized", b.op.as_str()),
+            In | NotIn => (
+                self.algebrize_expression(*b.left, false)?,
+                self.algebrize_expression(*b.right, false)?,
```
mattChiaravalloti: [Issue] **(blocking)** We'll need to consider the `implicit_type_conversion` value here -- it should not just be `false` by default.

Consider this example: `SELECT * FROM foo WHERE d IN ('{"$date": "..."}', '{"$date": "..."}', '{"$date": "..."}')`, where `d` is a `BSON_DATE`. Here, we must assume that the user intened implicit type conversion for the `$date`-formatted string values, since this is a comparison operation and one operand contains `String`s while the other is non-string. The current implementation does not do this.

Unfortunately, since this is a comparison between a single expression and an array of expressions to compare against, it isn't so straightforward and we can't reuse existing functions following the exact patterns that are already established. We'll have to write some new code to support this:
1. Our ultimate goal is to match how `algebrize_binary_comparison_operands` works (used just a couple lines above) but in a way that supports comparing the LHS of `In`/`NotIn` to each element of the RHS.
2. We should create a helper method like `algebrize_in_comparison_operands` or similar, or we could write the relevant code in-line here in this method -- your call
3. That method should check if both sides (the LHS and every _element_ of the RHS array) are strings
    1. If they all are, then we proceed with algebrization with the flag set to `false`
    2. If only one side is string (for example, the LHS is a `BSON_DATE` but the RHS elements are all string, _or_ the LHS is a string but the RHS are all `Int`, then we should algebrize the `String` expr(s) with the flag set to `true` (and the other expr(s) with it set to `false`)
    3. If neither are strings, then we proceed with algebrization with the flag set to `false`
4. We will also need to update the `algebrize_expression` case for `ast::Tuple`. Currently, `algebrize_expression` assumes we should always algebrize `Tuple` elements with `in_implicit_type_conversion_context` set to `false`. We could/should note in a code-comment above the match case for algebrizing `Tuple` that it is unlike `Array` since it supports implicit type conversion of its elements since `Tuple` exprs explicitly exist in contexts of comparison whereas `Array` exprs can exist anywhere. You'll want to update that from using a hardcoded `false` to using the argument `is_in_implicit_type_conversion_context`.

mongosql/src/algebrizer/test/expressions/scalar_function.rs
```diff
@@ -1247,6 +1247,68 @@ test_algebrize!(
     }),
 );
 
+test_algebrize!(
```
mattChiaravalloti: [Question] Is this the same test as the one from `binary.rs`? Does it need to live in both places? I _think_ the test file names correspond to input types, not output types, so this test is better placed in `binary.rs`. We can likely remove this here.

mongosql/src/mir/optimizer/match_null_filtering/mod.rs
```diff
@@ -395,7 +395,9 @@ impl ScalarFunction {
             | ScalarFunction::DayOfYear
             | ScalarFunction::IsoWeek
             | ScalarFunction::IsoWeekday
-            | ScalarFunction::MergeObjects => false,
+            | ScalarFunction::MergeObjects
+            | ScalarFunction::In
+            | ScalarFunction::NotIn => false,
```
mattChiaravalloti, line 400: [Issue] This seems incorrect. See the following tests I did in mongosh below:
```
Enterprise test> db.foo.aggregate()
[
  {
    _id: ObjectId("69fb85f1b556e39baca930f1"),
    a: null,
    c: 1,
    d: null,
    e: 1
  }
]
Enterprise test> db.foo.aggregate({$project: {"a_in_(1,2,3)": {$in: ["$a", [1, 2, 3]]}}})
[
  { _id: ObjectId("69fb85f1b556e39baca930f1"), 'a_in_(1,2,3)': false }
]
Enterprise test> db.foo.aggregate({$project: {"a_in_(1,2,3)": {$in: ["$a", [null, 1, 2, 3]]}}})
[ { _id: ObjectId("69fb85f1b556e39baca930f1"), 'a_in_(1,2,3)': true } ]
Enterprise test> db.foo.aggregate({$project: {"c_in_(1,2,3)": {$in: ["$c", [1, 2, 3]]}}})
[ { _id: ObjectId("69fb85f1b556e39baca930f1"), 'c_in_(1,2,3)': true } ]
Enterprise test> db.foo.aggregate({$project: {"c_in_(1,2,3)": {$in: ["$c", [1, 2, null, 3]]}}})
[ { _id: ObjectId("69fb85f1b556e39baca930f1"), 'c_in_(1,2,3)': true } ]
```
So it seems the semantics _do_ diverge since we say that `IN` should return true if any of the comparisons return true, otherwise if any of the comparisons return null it returns null, otherwise it returns false. Similar logic applies to `NOT IN`.

Should we change this?

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -120,6 +121,45 @@ impl MatchLanguageRewriterVisitor {
             })
     }
 
+    /// Rewrites `IN` and `NOT IN` scalar functions to native match language when the
+    /// left-hand side is a field access and every value in the list is a literal.
+    ///
+    /// Any other shape (subqueries, field refs in the list, functions) falls through
+    /// to `None` so the expression stays in `ExprLanguage`.
```
mattChiaravalloti: [Praise] Nice! Good thinking here, and the tests for this look great too.

mongosql/src/air/definitions.rs
```diff
@@ -237,6 +237,7 @@ pub enum Expression {
     SubqueryExists(SubqueryExists),
     Array(Vec<Expression>),
     Document(UniqueLinkedHashMap<String, Expression>),
+    Tuple(Vec<Expression>),
```
mattChiaravalloti: [Issue] I'm already dubious of having a `mir::Expression::Tuple`, as noted previously, but I am certain we should not introduce an `air::Expression::Tuple`. `air` is an abstract representation of aggregation pipelines, and "tuples" are not a concept that are actually represented in aggregation. This should be removed in favor of `air::Expression::Array`.

mattChiaravalloti (changes requested):
I made it through everything except `translator` and `codegen`. I'm getting nervous I'll lose these big comments at this point so I'm submitting this partial review now. I'll follow up with thoughts on `translator` and `codegen`. This is a very large change! Nice job getting to all of the areas that need to be addressed. There is more work to do to get this generally correct but we're on a good path here.

mongosql/src/translator/match_query.rs
```diff
@@ -25,9 +25,37 @@ impl MqlTranslator {
...
     }
 
+    fn translate_match_in_operator(&self, in_op: mir::MatchLanguageIn) -> Result<air::MatchQuery> {
+        let op = match in_op.op {
+            mir::MatchLanguageInOp::In => air::MatchLanguageInOp::In,
+            mir::MatchLanguageInOp::NotIn => air::MatchLanguageInOp::NotIn,
+        };
+
+        let expression = match in_op.input {
+            None => return Err(Error::InvalidMatchLanguageInputRef),
```
mattChiaravalloti: [Suggestion] Why would this ever be `None`? Can we update `mir::MatchLanguageIn.input` to be `FieldPath` instead of an `Option<FieldPath>`? Looking at the optimization that creates `mir::MatchLanguageIn`, there doesn't seem to be a need to make it an option since it is always set to `Some(field_path)`.

mongosql/src/air/definitions.rs
```diff
@@ -521,7 +525,8 @@ pub struct UnsetField {
 }
 
 #[derive(PartialEq, Debug, Clone)]
-pub struct FieldRef {
+pub struct
+    FieldRef {
```
mattChiaravalloti, line 528:
```suggestion
pub struct FieldRef {
```
[nit/suggestion/question] Why the newline?

mongosql/src/translator/utils.rs
```diff
@@ -372,6 +372,12 @@ impl From<mir::ScalarFunction> for ScalarFunctionType {
 
             // MergeObjects merges an array of objects
             MergeObjects => ScalarFunctionType::Mql(MqlOperator::MergeObjects),
+            // $in is valid in aggregation expression contexts; route through SqlSemanticOperator
+            // so both In and NotIn share a single consistent codegen path.
+            // $nin is match-language-only and rejected by the server in $project; NotIn must
+            // emit { "$not": [{ "$in": [...] }] } via SqlSemanticOperator codegen.
+            In => ScalarFunctionType::Sql(SqlOperator::In),
+            NotIn => ScalarFunctionType::Sql(SqlOperator::NotIn),
```
mattChiaravalloti, line 399: [Thought] (blocking) I'm thinking through how algrebrization, translation, and codegen will work for these operators. It seems that we may still have additional work to do in the algebrization step which will impact these other steps.

In particular, we need to determine if the algebrized expression is nullable -- right now, we're doing that the way every other binary op does it but that's wrong! Since `IN`/`NOT IN` deal with an expr and an array of exprs, we determine nullability by either the LHS being nullable or any element of the RHS being nullable, right? `nullable_a IN (1, 2, 3)` is possibly null, `non_nullable_a IN (1, 2, NULL, 3)` is possibly null, `non_nullable_a IN (1, 2, 3)` is never null, and `nullable_a IN (1, 2, NULL, 3)` is possibly null. We'll need to update algebrization to account for this. Then, when we're doing translation here, we'll follow the correct path to translate these as SQL semantic or MQL semantic operators (if the scalar function is not nullable, we can translate to the MQL version, if it is nullable we need to translate to the SQL version).

On top of that, we may need to deal with the SQL version in the desugarer. Specifically, `a IN (1, 2, NULL, 3)` should be true if `a` is `1`, `2`, or `3`, or it should be `null` if it itself is null or if it is not `1`, `2`, or `3` (because the result of the third comparison is `null`). However, as currently implemented we will just codegen `{$in: ["$a", [1, 2, null, 3]]}` and not get the expected null result when we should. The desugarer could handle this as part of the `sql_null_semantics_operators` pass.

Ultimately, the reason I'm commenting this here is because I think we'll need to update this translation util function to return `MqlOperator::In`/`MqlOperator::NotIn` in the even the function itself is not nullable.

mongosql/src/translator/test/expressions.rs
```diff
@@ -4345,3 +4345,91 @@ mod mql_intrinsic {
         },
     );
 }
+
+mod in_operator {
```
mattChiaravalloti, line 4349: [Chore] We'll need to add tests for when the functions are nullable (`is_nullable: true`) -- these tests only handle when it that value is false.

mongosql/src/translator/test/match_query.rs
```diff
@@ -389,3 +389,91 @@ mod match_path {
...
+mod match_in_operator {
+    use crate::{
+        air,
+        mapping_registry::{MqlMappingRegistry, MqlMappingRegistryValue, MqlReferenceType},
+        mir,
+        options::SqlOptions,
+        translator::{self, test::match_query::mir_field_input},
+    };
+
+    fn make_translator() -> translator::MqlTranslator {
```
mattChiaravalloti: [Question] This helper function, along with the test functions themselves, look identical to the implementation of the `test_translate_match_query` macro. Is there any difference between the two? Is there a reason you didn't use the macro?

mongosql/src/codegen/match_query.rs
```diff
@@ -33,6 +34,28 @@ impl MqlCodeGenerator {
...
+            .collect();
+
+        let op = bson!({ "$in": Bson::Array(values) });
+
+        match args.op {
+            air::MatchLanguageInOp::In => Ok(bson!({ field: op })),
+            air::MatchLanguageInOp::NotIn => Ok(bson!({
+                field: {
+                    "$not": op
+                }
```
mattChiaravalloti, line 57: [Question] It looks like `$nin` is a valid operator for `$match` query predicates ([link](https://www.mongodb.com/docs/manual/reference/operator/query/nin/#mongodb-query-op.-nin)). Do you think we can use it instead of `{$not: {$in: ...}}` for match language codegen, or is there a specific reason we can't/shouldn't?

(`$nin` is not available in general agg language, so not using it in `{$match: {$expr: ...}}` or in `$project` (generally, anywhere outside of match-language `$match`) still make sense.)

mattChiaravalloti (changes requested):
Rest of my thoughts -- this time focused on `translator` and `codegen` (and, implicitly, `air::desugarer`).

mattChiaravalloti:
One last idea is that we may want to update the `semantics.md` doc to reflect these changes. It should no longer be considered part of the spec that `X IN Y` is rewritten to `X = ANY(Y)`, etc. [This section](https://github.com/mongodb/mongosql/blob/main/semantics.md#behavior-and-grammar-for-in-any-some-all-exists) in particular covers this. We may also want to add some spec query tests to demonstrate how `IN` and `NOT IN` work.

mongosql/src/codegen/match_query.rs
jpowell-mongo: I'm going to revisit this one - I ran into issues when trying to use $nin with match language. I'll do some testing again to see if I can replicate. I remember getting some errors related to $nin in the match language during testing but unfortunately didn't take notes. I might just jump to an earlier commit

mongosql/src/ast/pretty_print.rs
jpowell-mongo: I didn't notice this naturally as a driveby fix, the fuzz test ended up failing - which prompted me to investigate. It isn't related to the changes in the PR as far as I know. I'll need to confirm again

mongosql/src/ast/rewrites/mod.rs
jpowell-mongo: Let's delete it. It'll be in the git history. 

Removed it here:
https://github.com/mongodb/mongosql/pull/144/changes/12daafce2d8116e8084e75fcb036c8a878a6d3a1

mongosql/src/mir/definitions.rs
jpowell-mongo: Discussed offline, but yes I will migrate all the changes to just use the mir::Expression::Array type instead of the duplication of the Tuple type.

jpowell-mongo: Updated the code to just use the Array type here:
https://github.com/mongodb/mongosql/pull/144/changes/d7570590923bae851fa65f04c3235e330c750fe7

mongosql/src/mir/schema/mod.rs
jpowell-mongo: Thanks for the comment, I confused the $in operator accepted an array as being variadic, rather than the actual $in operator requiring exactly two arguments - the field path, and an expression that resolves to an arry -  like you mentioned. Going to dig into this and see where I get.

jpowell-mongo: Addressed here with a new function and unit tests:
https://github.com/mongodb/mongosql/pull/144/changes/5bcaccc92999a675707bbfac09cecdcb49aad053

mongosql/src/mir/schema/test/expressions/scalar_function.rs
jpowell-mongo: Yes, you were right, I updated the code, and this did in fact fail. I added new tests and the get_in_operator_comparison_schema that you described here:
https://github.com/mongodb/mongosql/commit/5bcaccc92999a675707bbfac09cecdcb49aad053

mongosql/src/air/definitions.rs
jpowell-mongo: I removed the Tuple type across our intermediate representations in this commit:
https://github.com/mongodb/mongosql/commit/d7570590923bae851fa65f04c3235e330c750fe7

mongosql/src/codegen/match_query.rs
jpowell-mongo: Note: I'm going to explore this, but during discussion we agreed that it would okay if we preserved the not { in } vs. $nin

mongosql/src/translator/utils.rs
jpowell-mongo: I updated the algebrizer and parsing logic here:
https://github.com/mongodb/mongosql/pull/144/changes/6d16cde311dd197a75794fc290971bdbc6fad688

mongosql/src/mir/optimizer/match_null_filtering/mod.rs
jpowell-mongo: These semantics do diverge. I'll update this.

mongosql/src/translator/utils.rs
jpowell-mongo: I'll make the desugarer change in another PR

mongosql/src/algebrizer/test/expressions/scalar_function.rs
mattChiaravalloti: Question still stands. My suggestion is that we remove this test; this PR should not need to make changes to this file.

mongosql/src/algebrizer/test/expressions/binary.rs
```diff
@@ -715,3 +715,403 @@ test_algebrize!(
...
+            right: Box::new(ast::Expression::Tuple(vec![
+                ast::Expression::Literal(ast::Literal::Integer(1)),
+                ast::Expression::Literal(ast::Literal::Integer(2)),
+                ast::Expression::Literal(ast::Literal::Integer(3)),
+            ])),
+        }),
+        env = map! {
+            ("foo", 1u16).into() => Schema::Document( Document {
+                keys: map! {
+                    "a".into() => Schema::Atomic(Atomic::String),
```
mattChiaravalloti:
```suggestion
                    "a".into() => Schema::Atomic(Atomic::Integer),
```
We still need to update this or else it will ultimately fail when schema checking is implemented correctly.

mongosql/src/algebrizer/test/expressions/binary.rs
```diff
@@ -715,3 +715,403 @@ test_algebrize!(
...
```
mattChiaravalloti:
```suggestion
                    "a".into() => Schema::Atomic(Atomic::Integer),
```
Same issue applies here.

mongosql/src/algebrizer/test/expressions/binary.rs
```diff
@@ -715,3 +715,403 @@ test_algebrize!(
...
+            left: Box::new(ast::Expression::Identifier("a".to_string())),
+            op: ast::BinaryOp::In,
+            right: Box::new(ast::Expression::Tuple(vec![ast::Expression::Literal(
+                ast::Literal::Boolean(true)
+            )])),
+        }),
+        env = map! {
+            ("foo", 1u16).into() => Schema::Document( Document {
+                keys: map! {
+                    "a".into() => Schema::Atomic(Atomic::String),
```
mattChiaravalloti:
```suggestion
                    "a".into() => Schema::Atomic(Atomic::Boolean),
```
Same issue applies here as well.

mongosql/src/algebrizer/test/expressions/binary.rs
```diff
@@ -715,3 +715,403 @@ test_algebrize!(
...
+        env = map! {
+            ("foo", 1u16).into() => Schema::Document( Document {
+                keys: map! {
+                    "a".into() => Schema::Atomic(Atomic::Integer),
+                },
+                required: set!{},
+                additional_properties: false,
+                ..Default::default()
+            }),
+        },
```
mattChiaravalloti, line 890:
```suggestion
```
Since this test uses all `Literal`s, we don't even need to include an `env`.

mongosql/src/algebrizer/test/expressions/binary.rs
```diff
@@ -715,3 +715,403 @@ test_algebrize!(
...
+    );
+
+    test_algebrize!(
+        in_operator_enables_conversion_when_lhs_is_extended_json_and_rhs_is_tuple,
```
mattChiaravalloti: Is this test a duplicate of the one above? Is this the one we discussed offline that you said you'd remove? If so, feel free to ignore this comment since you already plan to remove it. If not, I think this is a duplicate and we should remove it!

mongosql/src/algebrizer/test/expressions/binary.rs
mattChiaravalloti, line 1: [Suggestion] Discussed-offline already, but leaving this comment for tracking purposes: let's make sure to add a test where only some of the array elements are `StringConstructor`s when the LHS is not a String type -- that way we confirm we still convert the `StringConstructor` elements.

An example is something like `int_field IN ('{"$numberInt": "1"}', other_int_field, 2)`. The first element of the tuple should be implicit-type-converted from a string into an integer.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -314,7 +313,35 @@ impl<'a> Algebrizer<'a> {
         // some functions can always be nullable regardless of argument nullablity,
         // we check those first. If this function is not one of those, we set nullablity
         // based off the arguments.
```
mattChiaravalloti: [Suggestion] I think this comment belongs inside the `else` block since it is associated specifically with that code.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1567,60 @@ impl<'a> Algebrizer<'a> {
         }
     }
 
+    /// Algebrizes the operands of an `IN`/`NOT IN` expression with ITC awareness.
+    ///
+    /// Mirrors [`Self::algebrize_binary_comparison_operands`] but handles the fact that the RHS
+    /// is an [`ast::Expression::Tuple`] containing multiple elements rather than a single
+    /// expression. If exactly one side consists entirely of [`ast::Expression::StringConstructor`]
```
mattChiaravalloti: [Suggestion] **(blocking)** We'll need to update this comment to reflect the correct semantics: "if exactly one side contains _any_ `StringConstructor` nodes..."

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1567,60 @@ impl<'a> Algebrizer<'a> {
...
+    fn algebrize_in_operands(
+        &self,
+        left: ast::Expression,
+        right: ast::Expression,
+    ) -> Result<(mir::Expression, mir::Expression)> {
+        // 1. Check if all the elements of the right expression are StringConstructors.
+        let are_array_elements_all_string_constructors = match &right {
+            ast::Expression::Tuple(arr) => arr
+                .iter()
+                .all(|e| matches!(e, ast::Expression::StringConstructor(_))),
```
mattChiaravalloti: Discussed offline: change to `any`

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1567,60 @@ impl<'a> Algebrizer<'a> {
...
+                self.algebrize_expression(left, false)?,
+                self.algebrize_expression(right, false)?,
+            )),
+            // LHS is a StringConstructor; RHS Tuple contains at least one non-string element.
+            // Algebrize the RHS to derive its schema, then ITC-convert the LHS accordingly.
+            (true, false) => {
+                let (literal, non_literal) =
+                    self.algebrize_itc_eligible_binary_comparison_operands(left, right)?;
+                Ok((literal, non_literal))
+            }
```
mattChiaravalloti: Note we still can't just directly use `algebrize_itc_eligible_binary_comparison_operands` in this case since `right` is an array. The array will never report that it is `STRING_OR_NULLISH` after it algebrizes unless we update how either `algebrize_itc_eligible_binary_comparison_operands` (plural) or `algebrize_itc_eligible_binary_comparison_operand` (singluar) work.

We should either update one of those methods to deal with Arrays of Strings (not advised) or create a new helper method (or in-line code block) for this case specifically. In this case, we want to algebrize the RHS (the `non_literal`) with `itc: false` and then check its resulting the schema. We expect the schema to be `Array<something>`. What we want to do is check if that `something` satisfies `STRING_OR_NULLISH`. If so, don't attempt to itc the LHS (`literal`) -- as in, call `self.algebrize_expression(literal, false)`; if not, do attempt to idc the LHS (`literal`) -- as in, call `self.algebrize_expression(literal, true)`.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1567,60 @@ impl<'a> Algebrizer<'a> {
...
+            // LHS is not a string, RHS has elements that are all strings
```
mattChiaravalloti:
```suggestion
            // LHS is not a string, RHS has at least one element that is a StringConstructor
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1567,60 @@ impl<'a> Algebrizer<'a> {
...
+            // Do the reverse. We algebrize the non-literal LHS and use that to algebrize the RHS.
+            (false, true) => {
+                let (literal, non_literal) =
+                    self.algebrize_itc_eligible_binary_comparison_operands(right, left)?;
```
mattChiaravalloti: For context, this side is fine since the result of `algebrize_itc_eligible_binary_comparison_operand` (singular) for `literal` in this case is valid since it is truly a singular value and not an array.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1771,6 +1859,7 @@ impl<'a> Algebrizer<'a> {
 
         let args = vec![left, right];
         let function = mir::ScalarFunction::try_from(b.op)?;
+        // TODO: if the operator is an IN or NOT IN expression we determine nullabillity if LHS or any element in RHS is nullable
```
mattChiaravalloti:
```suggestion
```
Looks like you addressed this!

mongosql/src/algebrizer/definitions.rs
```diff
@@ -2415,6 +2504,105 @@ impl<'a> Algebrizer<'a> {
     }
 }
 
+mod in_operator_nullability {
```
mattChiaravalloti, line 2523:
```suggestion
#[cfg(test)]
mod in_operator_nullability {
```
[Suggestion] this can be guarded with `#[cfg(test)]` since it is just a test module!

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
mattChiaravalloti: [Suggestion] We should update this comment a bit to reflect that this now supports `IN`/`NOT IN` as well.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -3476,3 +3476,100 @@ mod neg {
         )),
     );
 }
+
+mod in_operator {
+    use super::*;
+
+    test_schema!(
+        in_operator_requires_two_args,
```
mattChiaravalloti:
```suggestion
        not_in_operator_requires_two_args,
```
[Question/Suggestion] Should we name these for `in` and `not_in` distinctly?

mattChiaravalloti: This question applies throughout this file but I'll only comment it once.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
mattChiaravalloti: Question still stands, thought there's a typo in my suggestion. Should be `not_in_operator_schema_is_boolean`.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1489,6 +1492,86 @@ trait SqlFunction {
...
+    ///   not satisfy `Schema::Array(Any)`.
+    /// - [`Error::InvalidComparison`] — the LHS type is not comparable to the element type of
+    ///   the RHS array.
+    ///
+    /// # Panics
+    ///
+    /// Contains an `unreachable!()` branch for the non-array RHS match arm. This branch is
+    /// structurally unreachable: the preceding guard already returns
+    /// [`Error::SchemaChecking`] for any RHS that does not satisfy `Schema::Array(Any)`.
+    fn get_in_operator_comparison_schema(
```
mattChiaravalloti, line 1516: [Praise] The implementation of the method is fantastic!

mongosql/src/translator/test/expressions.rs
```diff
@@ -4345,3 +4345,91 @@ mod mql_intrinsic {
...
+    // Test conversion
+    use crate::mapping_registry::MqlReferenceType;
+    use crate::mir::{
+        ArrayExpr, Expression, LiteralValue, ScalarFunction, ScalarFunctionApplication,
+    };
+    use crate::util::mir_field_access;
+    use crate::{air, mapping_registry::MqlMappingRegistryValue};
+
+    test_translate_expression!(
+        it_converts_to_mql_when_lhs_is_nullable,
```
mattChiaravalloti: [Question] **(blocking)** Shouldn't it be that if either side is nullable it translates to SQL? And when both are not nullable it translates to MQL?

mattChiaravalloti: It's really just the test name that is misleading, and the input that is invalid. If either operand is marked with `is_nullable: true`, the entire expression should also be marked with `is_nullable: true`. This test sets `is_nullable: false` for the overall `In` expression, but marks one of its operands, the LHS, with `is_nullable: true`.

Can you fix this test to be valid to demonstrate that when none of the args is nullable, the translation is MQL. And then a separate test that shows when either arg is nullable (and therefore the expression is nullable) it translates to SQL?

mattChiaravalloti (approved):
Marking as approved even though I am leaving a ton of comments. I just don't want to block this from merging while I'm gone.

There is still work to be done even though I marked as approved:
- all outstanding open comment threads should either result in code changes or have answers to them that explain why not (I resolved the ones that are addressed in one way or another)
    - A lot of the threads are still just some bookkeeping changes like fixing test names and/or inputs, addressing odd newlines, etc.
- still need to add the spec/query tests as discussed offline
- still need to update the desugarer before this can merge since it won't behave correctly until we do (we'll see the spec/query tests fail without this work done)

Feel free to tag someone else in in addition to Nick to give this another thorough pass if you want!

mongosql/src/translator/utils.rs
jpowell-mongo: The Desugaring Changes are implemented here:
https://github.com/mongodb/mongosql/pull/144/changes/771ae2d123f93f932c8aff72fe183e4f0469f713

tests/index_usage_tests/filter.yml
jpowell-mongo: Index Utilization tests pass. Added here:
https://github.com/mongodb/mongosql/pull/144/changes/5c6b0f6fd4f6a874571f324de5db4ceec704c2d5

mongosql/src/translator/match_query.rs
jpowell-mongo: Updated here: 
https://github.com/mongodb/mongosql/pull/144/changes/5c6b0f6fd4f6a874571f324de5db4ceec704c2d5

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
```diff
@@ -48,6 +49,192 @@ impl SqlNullSemanticsOperatorsDesugarerVisitor {
...
+
+        assert_eq!(
+            sql_operator.args.len(),
+            2,
+            "desugar_sql_in: expected exactly 2 args (lhs, rhs array), got {}",
+            sql_operator.args.len()
+        );
+        let mut args = sql_operator.args.into_iter();
+        let lhs = args.next().unwrap();
+        let rhs = args.next().unwrap();
```
nicholascioli: Nit: you can use take for collecting all at once


```suggestion
        let [lhs, rhs] = sql_operator.args[..2].try_into().unwrap();
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -46,9 +47,8 @@ impl TryFrom<ast::BinaryOp> for mir::ScalarFunction {
             ast::BinaryOp::Mul => mir::ScalarFunction::Mul,
             ast::BinaryOp::Or => mir::ScalarFunction::Or,
             ast::BinaryOp::Sub => mir::ScalarFunction::Sub,
-            ast::BinaryOp::In | ast::BinaryOp::NotIn => {
-                panic!("{0} cannot be algebrized", op.as_str())
```
nicholascioli, line 50: Praise: Love to see a panic go away

mongosql/src/algebrizer/definitions.rs
```diff
@@ -311,10 +311,38 @@ impl<'a> Algebrizer<'a> {
...
+    /// An IN / NOT IN is nullable if the left-hand side is nullable, or if any elements
+    /// of the right-hand side are nullable.
+    pub fn determine_in_expression_nullability(
+        func: mir::ScalarFunction,
+        args: &[mir::Expression],
+    ) -> bool {
+        assert!(
+            matches!(func, mir::ScalarFunction::In | mir::ScalarFunction::NotIn),
+            "In Expression nullability is only for In and NotIn Operator. Given {func:?}"
+        );
```
nicholascioli: Thought: I'm not sure about how `asserts!` are used throughout this crate, but know that they are only enabled for debug builds. Considering that `func` is never used in this method, is it worth just not taking in a `func` in this method?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -311,10 +311,38 @@ impl<'a> Algebrizer<'a> {
...
+
+        let lhs = args.first();
+        let is_lhs_is_nullable = lhs.is_some_and(|e| e.is_nullable());
+
+        let rhs = args.get(1);
+        let is_rhs_nullable = rhs.is_some_and(|e| match e {
+            mir::Expression::Array(arr) => arr.array.iter().any(|elem| elem.is_nullable()),
+            _ => e.is_nullable(), // For other types of expressions, we check their nullability directly.
```
nicholascioli, line 334: Question (non-blocking): Is there a case where the `rhs` is not an array? Comments in other files seem to imply that this should never happen.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1532,6 +1568,83 @@ impl<'a> Algebrizer<'a> {
...
+                            let algebrized_element = match element {
+                                ast::Expression::StringConstructor(_) => {
+                                    self.algebrize_expression(element, true)?
+                                }
+                                _ => self.algebrize_expression(element, false)?,
+                            };
+
+                            algebrized_elements.push(algebrized_element);
+                        }
+                        algebrized_elements
```
nicholascioli, line 1625: Nit: Might be nice to just `collect` here


```suggestion
                        elements.into_iter()
                            .map(|e| match e {
                                ast::Expression::StringConstructor(_) => {
                                    self.algebrize_expression(element, true)
                                }
                                _ => self.algebrize_expression(element, false),
                            })
                            .collect::<Result<Vec<_>, _>>()?
```

mongosql/src/mir/schema/mod.rs
```diff
@@ -26,8 +26,10 @@ use std::{
 
 mod errors;
 pub use errors::Error;
-#[cfg(test)]
+#[allow(unused_imports)]
```
nicholascioli: Question (non-blocking): Why swap the test cfg for a clippy warning disable?

mongosql/src/mir/schema/mod.rs
```diff
@@ -1489,6 +1492,92 @@ trait SqlFunction {
...
+            return Err(Error::IncorrectArgumentCount {
+                name: self.as_str(),
+                required: 2,
+                found: arg_schema.len(),
+            });
+        }
+
+        // 2. In Operator LHS must be ANY and RHS must be an Array
+        let in_operator_lhs = &arg_schema[0];
+        let in_operator_rhs = &arg_schema[1];
```
nicholascioli: Nit: You can technically do both of these at once:


```suggestion
        let [in_operator_lhs, in_operator_rhs] = arg_schema.try_into()
            .map_err(|_| Error::IncorrectArgumentCount {
                name: self.as_str(),
                required: 2,
                found: arg_schema.len(),
            })?;
```

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
nicholascioli, line 82: Question (non-blocking): Is it OK to panic here versus return an error?

Nit: Also, You can do all of this at once.


```suggestion
        let [lhs, rhs] = sql_operator.args.try_into().unwrap_or_else(|args: Vec<_>| {
            panic!(
                "desugar_sql_in: expected exactly 2 args (lhs, rhs array), got {}",
                args.len()
            )
        });
```

nicholascioli (commented):
Sorry this took so long to review! It looks absolutely fantastic and goes a bit over my head in terms of how it fits into everything else. I left a few nits and non-blocking questions, but well done.

mongosql/src/algebrizer/definitions.rs
jpowell-mongo: There aren't any cases where rhs is not an array - I've added checks upstream, so it's unlikely, but I felt that because these are still passed down as enum variants it was worthwhile to have that validation. 

You cannot construct a SQL query that would make it to the algebrizer that is not an array. I think that would be prevented even at the AST level.

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
jpowell-mongo: Since we do the validation before grabbing the elements, I would not expect the unwrap to panic. But you're write - we should avoid the unwraps. I'll update the code accordingly.

mongosql/src/algebrizer/definitions.rs
jpowell-mongo: That's a good suggestion, I was split on this, but given the method is literally called `determine_in_expression_nullability` I think callers (if this was ever used elsewhere) would know that it needs to be an IN / NOT IN expression. I'll remove the func argument.

jpowell-mongo: https://github.com/mongodb/mongosql/pull/144/changes/18440593ce695fa718c802cc15b0774fada1e9ba

mongosql/src/mir/schema/mod.rs
jpowell-mongo: I didn't even realize this is what I dropped. Thanks for catching this! I didn't realize that was causing the issue.

PR #148, opened 2026-05-22
SQL-3225: seperate cli and schema management logic from library

jchemburkar:
During the process of creating WASM vs Multi-threaded schema manager, we discovered that it was difficult to support different forms of orchestration without requiring the duplication of code. Namely, we could easily do things like getting partitions, deriving schema for a partition, etc atomically, but for broader, parallelizable tasks like creating tasks for each collection to construct a schema, we started needing multiple code paths.

After some discussion, it was determined a cleaner, more sustainable approach would be to focus on "separation of concerns" -- that is, we will have core schema generation logic in the schema-builder-library, while deferring orchestration (how do we get schemas for every collection in a database? how do we get schemas for every collection in a cluster? and of course, creating a cli to manage this) to downstream consumers. A mitigation for this for other downstream consumers (wasm) is that we can add higher level, synchronous functions for deriving a schema for a collection in the bindings.

This PR removes the bulk of the logic that is more of a "schema manager" concern, while making certain functions public for the schema manager to pull in. The companion pr on schema manager can be found [here](https://github.com/10gen/schema-manager-rs/pull/999). The major followup to this will be [SQL-3167](https://jira.mongodb.org/browse/SQL-3167), which will cover adding anything needed to the wasm bindings and ensuring it still compiles.

There are likely ergonomic benefits to be gained for callers of the library that are not wasm _or_ schema manager; that is, things we could do to make using this api easier. Some ideas:
- we can add constructors/default values that limit the lines needed to do "get partitions" and "derive schema for partitions"
- we could write a thin wrapper around derive_schema_for_collection -- I have a suspicion parallelizing the partition schema derivation is actually not gaining us more parallelism, so we can remove one level of `tokio::spawn` from schema manager and move that function here
- if ^ is feasible, we could consider adding initial schema querying _into_ the derive_schema_for_collection wrapper and doing it inline for every coll

the api as it stands is entirely bare bones, enough for schema manager, wasm, and an internal user (with our guidance). The above proposals (which can go in the followup schema manager improvements epic) would help users to not want/need our guidance to implement


schema-builder-library/src/lib.rs
```diff
@@ -30,320 +18,16 @@ pub use data_service::{JsDataService, WasmDataService};
...
 pub mod client_util;
 
-mod consts;
-pub(crate) mod context;
-use consts::{DISALLOWED_DB_NAMES, VIEW_SAMPLE_SIZE};
-mod collection;
-use collection::{DatabaseCollections, query_for_initial_schemas};
 mod partitioning;
-use partitioning::get_partitions;
+pub use partitioning::{PartitionedCollection, get_partitions};
```
jchemburkar, line 22: note: pub exports, the core change of this pr.

schema-builder-library/src/data_service.rs
```diff
@@ -70,11 +70,8 @@ pub struct AggregateOptions {
 }
 
 /// Abstraction over database operations used by the schema builder.
-///
-/// On non-WASM targets `Send + Sync` is required; on WASM it is not.
-#[cfg_attr(not(feature = "wasm"), async_trait::async_trait)]
-#[cfg_attr(feature = "wasm", async_trait::async_trait(?Send))]
-pub trait DataService {
+#[trait_variant::make(DataService: Send)]
+pub trait LocalDataService {
```
nicholascioli, line 74: Just wanted to point out the the switch from `async_trait` to `trait_variant` was necessitated by a long-standing rust bug (found [here](https://github.com/rust-lang/rust/issues/100013)) causing issues when using this trait with tokio. When async in traits was stabilized, the Rust blog brought up this exact problem and solution, as shown [here](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/#async-fn-in-public-traits).

This new procedural macro does come with the added benefit that downstream consumers can decide whether or not they care about the added `Send` bounds by using either `LocalDataService` or `DataService`, which is shown in the changes made to the mongodb and wasm backed clients.

schema-builder-library/src/errors.rs
```diff
@@ -26,12 +26,6 @@ pub enum Error<ServiceError: core::error::Error> {
...
     #[error("Initial schema for {0} is not valid")]
     InitialSchemaError(String),
-    #[error("Error when processing database pattern: {0}")]
-    PatternError(#[from] crate::collection::patterns::Error),
-    #[error(
-        "The `{0}` contains the following invalid pattern: `{1}`. All patterns must be in `<database_pattern>.<collection_pattern>` format"
-    )]
-    IncludeOrExcludeListContainsInvalidPatterns(String, String),
     #[error("{0}")]
     ChannelClosed(String),
```
nicholascioli: Nit: This is also dead code, right?

schema-builder-library/src/schema.rs
```diff
@@ -26,100 +22,15 @@ pub struct SinglePartition {
 
 pub const PARTITION_DOCS_PER_ITERATION: i64 = 20;
```
nicholascioli, line 23: Nit: This probably doesn't need to be `pub` anymore.

schema-builder-library/Cargo.toml
```diff
@@ -61,4 +59,4 @@ expect_used = "deny"
 [package.metadata.cargo-machete]
 ignored = [
     "wasm-bindgen-futures" # Note, this should be removed when the following is resolved: https://github.com/wasm-bindgen/wasm-bindgen/issues/5104#issuecomment-4223185700
-]
\ No newline at end of file
+]
```
nicholascioli, line 62: Nit: We might want to add the newline at the end here

Cargo.toml
nicholascioli, line 1: Praise: I love to see dependencies get removed

nicholascioli (commented):
Just left a few nits, but LGTM. I can't approve the PR since I wrote some of it, but thanks for putting this together!

schema-builder-library/src/errors.rs
nicholascioli: This was resolved with https://github.com/nicholascioli/mongosql/pull/2

bucaojit (approved):
The changes look good and I like the idea of separation of concerns.  That does clean things up and will help your upcoming work.

jpowell-mongo (approved):

PR #150, opened 2026-05-22
SQL-3167: wasm process collection

nicholascioli:
This PR adds a utility method for deriving a collection's schema linearly over its partitions. This should allow for single-threaded consumers to easily derive a schema for a collection without worrying about how to best split up the partitions or any other internal details.

If a consumer wishes to be more performant in how it processes the collection, it can always use the lower-level `derive_schema_for_partition` to do so.

Also, this PR is dependent on #148, so please review only the commits starting at "SQL-3167" and beyond. This PR will be rebased once the other PR is merged.


nicholascioli:
The PR has been rebased and is ready for review.

mongosql/Cargo.toml
```diff
@@ -45,3 +43,7 @@ harness = false
 [build-dependencies]
 lalrpop = "0.20.2"
 regex = { workspace = true }
+
+[dev-dependencies]
+quickcheck = { version = "1", default-features = false }
+rand = { workspace = true }
```
bucaojit: nit: Add a newline here.  
A few other files in this PR need this as well.

schema-builder-library/tests/wasm/tests/common.ts
```diff
@@ -0,0 +1,129 @@
+import type { SqlDataService, SqlCursor, BsonDocument, AggregateOptions } from "schema-builder-library";
```
bucaojit: It looks like this test isn't being run in evergreen, is it meant to be run?
If it's a manual test, you can add a small section in the README with a command on how to run it.  Although, I'd suggest having it run in evergreen since you have it here already.

bucaojit (approved):
Looks good, I had one comment regarding the integration test to consider.

schema-builder-library/tests/wasm/tests/common.ts
jchemburkar: I agree with the suggestion that we should run it on evergreen

schema-builder-library/src/data_service/wasm.rs
```diff
@@ -23,6 +26,20 @@ pub fn init() {
     console_error_panic_hook::set_once();
 }
 
+#[wasm_bindgen]
+pub async fn process_collection(
```
jchemburkar, line 34: I think we should add process view if we are going to add this

schema-builder-library/tests/wasm/src/index.ts
```diff
@@ -0,0 +1,129 @@
+import type { SqlDataService, SqlCursor, BsonDocument, AggregateOptions } from "schema-builder-library";
+
+import { EJSON, type Document } from "bson";
+import { MongoClient } from "mongodb";
+import { init, process_collection } from "schema-builder-library";
+
+const URI = "mongodb://localhost:27017";
```
jchemburkar: blocking: we should get this from the environment variable which will be set locally, or by evergreen, like other tests

schema-builder-library/tests/wasm/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (next == null) {
+            return null;
+        }
+
+        return toEJSON(next);
+    }
+
+}
+
+async function main() {
```
jchemburkar: blocking: I think we should aim to have a couple test functions rather than one main -- I'm no test expert, but we could even make a func for each test we care about and run them all in main.

I would like to check a few cases:
1. lets make sure the numeric case works (long vs int)
2. lets run one that hits various types and make sure they all work. 

we can probably reuse the test data and data loader for existing integraiton / e2e testing, take a look at the `testdata` folder, we can probably jsut pick a couple files from there to use for this.

schema-builder-library/tests/wasm/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    console.log("Connecting to client...");
+    let client = await new MongoClient(URI).connect();
+    console.log("> Connected");
+
+    let service = new MongoDataService(client);
+    try {
+        console.log(`Deriving schema for ${DB}.${COLLECTION}...`);
+
+        let start = new Date().getTime();
+        let schema = await process_collection(service, DB, COLLECTION);
```
jchemburkar: blocking: in addition to testing that schema derivation completes, at least for one of the subtests proposed ^ (if not all), lets check that the schema actually comes out correctly -- we want to ensure we're not producing `any` eg

schema-builder-library/tests/wasm/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+        let end = new Date().getTime();
+
+        console.log(`> Schema (took ${(end - start) / 1000}s):`, EJSON.stringify(schema, undefined, 2));
```
jchemburkar: question: In practice, we will never hit this because our existing integration tests are too small. I'm wondering if we should have some way of asserting we didn't just scan every single document / run into the issue we had before where it was generating a schema that didn't match the actual documents? 

alternatively maybe if we show we successfully generate a schema for every type, we know it will be user error / something with how they serialize/deserialize and we call that out of scope. if we choose to go this path, I vote we remove timing checks (and just rely on evergreen timeouts). 

wdyt?

jchemburkar (changes requested):
Great job with getting this to compile + adding the ts testing scaffolding! I just have some thoughts + questions on what exactly we should be testing

nicholascioli:
I've rebased against https://github.com/jchemburkar/mongosql/pull/5 to pull in changes needed to build. Once that gets merged, I'll rebase again.

schema-builder-library/tests/wasm/tests/process_collection.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        let aggregateSpy = vi.spyOn(SERVICE, "aggregate");
+        expect(aggregateSpy).not.toHaveBeenCalled();
+
+        // Verify that we only ran the aggregation pipeline exactly 4 times:
+        // - Get the size of the collection
+        // - Get the starting index
+        // - Get the end index
+        // - Get the first few elements
+        // - No longer get any other elements since they'll be filtered out by the schema
+        await process_collection(SERVICE, "uniform", "small");
```
jchemburkar, line 43: tiny nit but could we add just one test that does `process_view` -- we could even have the test create the view itself by running an agg pipeline. Just want to make sure that codepath gets hit

schema-builder-library/tests/wasm/tests/process_collection.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    test(`generates a valid schema on small collection`, async () => {
+        await expect(toEJSON(await process_collection(SERVICE, "uniform", "small")))
+            .toMatchFileSnapshot(`../snaps/uniform.small.schema.json`);
+    });
+
+    test(`does not sample more than it needs to`, async () => {
+        let aggregateSpy = vi.spyOn(SERVICE, "aggregate");
+        expect(aggregateSpy).not.toHaveBeenCalled();
+
+        // Verify that we only ran the aggregation pipeline exactly 4 times:
```
jchemburkar:
```suggestion
        // Verify that we only ran the aggregation pipeline exactly 5 times:
```

jchemburkar (approved):
lgtm. have one more test case I think we could add. might be worth having claude do one final dead code / style pass for the JS code to make sure there's no loose ends. but no need to retag once that's all set 😁 !

schema-builder-library/tests/wasm/tests/process_collection.test.ts
nicholascioli: Turns out that the sample dataset includes a view called `test_view`, so I just used that.

PR #152, opened 2026-05-27
SQL-3229: SBL integration test compile fix

jchemburkar:


jchemburkar:
rhel 89 added to evergreen checks, and succeeding: https://spruce.corp.mongodb.com/version/6a1737308330e30007dff23e/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Erhel89%24

schema-builder-library/src/internal_integration_tests/mod.rs
```diff
@@ -25,8 +25,6 @@ async fn create_mdb_client() -> mongodb::Client {
 
 #[cfg(feature = "integration")]
 #[cfg(test)]
-async fn get_mdb_collection(db: &str, coll: &str) -> mongodb::Collection<mongodb::bson::Document> {
-    let client = create_mdb_client().await;
-    let db = client.database(db);
-    db.collection::<mongodb::bson::Document>(coll)
+async fn create_mdb_service() -> crate::data_service::MongoDbDataService {
```
jchemburkar, line 28: core change -- the api of schema builder library operates more on data service now than the client collection object. Thus, we change the helper to get the data service, and then use it throughout the tests to get the collection.

schema-builder-library/src/internal_integration_tests/get_partitions.rs
```diff
@@ -96,15 +106,20 @@ async fn one_doc_collection() {
...
 
-    let coll = get_mdb_collection(UNIFORM_DB_NAME, "empty").await;
+    let service = create_mdb_service().await;
 
-    let actual_res = get_partitions(&coll, DEFAULT_COLLECTION_INFO.clone()).await;
+    let actual_res =
+        get_partitions(&service, UNIFORM_DB_NAME, DEFAULT_COLLECTION_INFO.clone()).await;
     match actual_res {
-        Err(Error::NoCollectionStats(_)) => {} // expect the NoBounds errors
+        Err(Error::DataServiceError(e)) => match *e.kind {
```
jchemburkar, line 119: Other important change, which we discussed during the original pr, was the changing of these error. These are correct by my recollection, but this is the other place that I would appreciate confirmation (perhaps @nicholascioli remembers)

Cargo.lock
nbagnard, line 1: question:
Why did you need a Cargo.lock update for this PR?

nbagnard (approved):
Changes looks good, I just don't understand why the Cargo.lock update.

Cargo.lock
jchemburkar: I got these changes without making any changes to the toml, and without calling `cargo update ...` on anything. It just updated when I ran `cargo build`. It appears that it comes from https://github.com/mongodb/mongosql/pull/149

what I see when asking augment is:
```
The PR bumped mongodb from "3" to "3.7" in  Cargo.toml (plus same bump in  test-utils/Cargo.toml) and refreshed  Cargo.lock by 71 lines. That refresh was incomplete:

- The bot likely ran cargo update -p hickory-proto --precise 0.26.1 (or equivalent) on a single platform (probably Linux), which only updates the edges visible from that host's resolver graph.
- Apple/Android-gated transitive deps of hickory-net 0.26.1 weren't visible from Linux, so they never made it into the lockfile.

Any later cargo invocation on macOS (or one that walks all targets, like building docs or running cargo update) sees the gap and tries to add the missing entries.
```

nbagnard: oh, ok, I think the bot is also not so helpful at dependency updates unfortunately.
Look at https://github.com/10gen/schema-manager-rs/pull/1000.
I mean it technically followed the description in the ticket. We might want to add the dependency tree or something like that to avoid these kind of mismatch.

evergreen/create-tasks/build.rs
```diff
@@ -86,6 +86,12 @@ fn main() {
     for &version in VERSIONS.iter() {
         for &(topology_name, topology) in TOPOLOGIES.iter() {
             for &(test_name, test_data_path, description, flags) in TEST_TYPES.iter() {
+                // SERVER-127822: the "latest" tag currently hits a $sequentialCache
```
jchemburkar, line 89: temporary fix for the failing test. Because this is in a build.rs, we need something that will avoid regenerating this test over and over. so I removed it.

could not pin a specific version, especially not an alpha version. thus found this to be the simplest solution.

jpowell-mongo (approved):

PR #160, opened 2026-06-29
[SQL-3299] Support NOT in air Match Language definitions

jpowell-mongo:
## Summary

This PR updates the AIR directory to support the $not operator in Match Stages. 

## Changes

- Introduce a MatchQuery NOT to existing enums
- Wire MatchQuery::Not into a match arm in codegen module
- Unit tests

## Testing

Unit tests to validate that $not is generated from the MatchQuery::Not expression

## Notes

This is a partial implementation. Follow up PRs will wire together the rest of Mongosql so that we can translate $not from SQL all the way to match language. Expect followups that connect the MIR -> AIR stage.

The $not support is based on the documentation for the $not _expression_ operator rather than the query predicate. Documentation can be found here:
https://www.mongodb.com/docs/manual/reference/operator/aggregation/not/

### DIFF
--- a/mongosql/src/air/definitions.rs
+++ b/mongosql/src/air/definitions.rs
@@ -720,6 +720,7 @@ pub struct Reduce {
 pub enum MatchQuery {
     Or(Vec<MatchQuery>),
     And(Vec<MatchQuery>),
+    Not(Box<MatchQuery>),
     Type(MatchLanguageType),
     Regex(MatchLanguageRegex),
     ElemMatch(ElemMatch),
--- a/mongosql/src/codegen/match_query.rs
+++ b/mongosql/src/codegen/match_query.rs
@@ -63,7 +63,7 @@ impl MqlCodeGenerator {
     /// Codegens `$not` scoped to a field's operator expression (mirroring how `NotIn` wraps
     /// `$in`), since `$not` — unlike `$and`/`$or`/`$nor` — is not a valid top-level `$match`
     /// boolean operator.
-    /// For
+    /// $and and $or are translated to use $nor instead, since $not is not allowed in top level expressions.
     fn codegen_match_not(&self, inner: air::MatchQuery) -> Result<Bson> {
         match inner {
             // $not is not allowed in top level expressions, so we use the $nor operator instead for $and, and $or.
@@ -83,7 +83,10 @@ impl MqlCodeGenerator {
                 Ok(match inner_codegen {
                     // Field-scoped operator, e.g. {"age": {"$gt": 10}} -> {"age": {"$not": {"$gt": 10}}}
                     Bson::Document(doc) if doc.len() == 1 => {
-                        let (key, value) = doc.into_iter().next().expect("checked len == 1");
+                        let (key, value) = doc
+                            .into_iter()
+                            .next()
+                            .expect("Match document expected to have 1 key but had 0.");
                         match value {
                             Bson::Document(_) => bson!({ key: { "$not": value } }),
                             // Fieldless operator (e.g. a comparison nested inside $elemMatch, which has
--- a/mongosql/src/codegen/match_query.rs
+++ b/mongosql/src/codegen/match_query.rs
@@ -26,6 +26,7 @@ impl MqlCodeGenerator {
         match q {
             Or(v) => self.codegen_match_logical_operator("$or", v),
             And(v) => self.codegen_match_logical_operator("$and", v),
+            Not(inner) => self.codegen_match_logical_operator("$not", vec![*inner]),
             Type(t) => self.codegen_match_type(t),
             Regex(r) => self.codegen_match_regex(r),
             ElemMatch(em) => self.codegen_match_elem_match(em),
--- a/mongosql/src/codegen/match_query.rs
+++ b/mongosql/src/codegen/match_query.rs
@@ -26,7 +26,7 @@ impl MqlCodeGenerator {
         match q {
             Or(v) => self.codegen_match_logical_operator("$or", v),
             And(v) => self.codegen_match_logical_operator("$and", v),
-            Not(inner) => self.codegen_match_logical_operator("$not", vec![*inner]),
+            Not(inner) => self.codegen_match_not(*inner),
             Type(t) => self.codegen_match_type(t),
             Regex(r) => self.codegen_match_regex(r),
             ElemMatch(em) => self.codegen_match_elem_match(em),
@@ -58,6 +58,36 @@ impl MqlCodeGenerator {
                 }
             })),
         }
+    }
+
+    /// Codegens `$not` scoped to a field's operator expression (mirroring how `NotIn` wraps
+    /// `$in`), since `$not` — unlike `$and`/`$or`/`$nor` — is not a valid top-level `$match`
+    /// boolean operator.
+    fn codegen_match_not(&self, inner: air::MatchQuery) -> Result<Bson> {
+        let inner = self.codegen_match_query(inner)?;
+        Ok(match inner {
+            // Field-scoped operator, e.g. {"age": {"$gt": 10}} -> {"age": {"$not": {"$gt": 10}}}
+            Bson::Document(doc) if doc.len() == 1 => {
+                let (key, value) = doc.into_iter().next().expect("checked len == 1");
+                match value {
+                    Bson::Document(_) => bson!({ key: { "$not": value } }),
+                    // Fieldless operator (e.g. a comparison nested inside $elemMatch, which has
+                    // no field ref of its own), e.g. {"$gt": 10} -> {"$not": {"$gt": 10}}
+                    _ => bson!({ "$not": { key: value } }),
+                }
+            }
+            // Anything else. This covers two different situations:
+            // - A fieldless *multi*-key operator document, e.g. a $regex nested inside
+            //   $elemMatch (which has no field ref of its own) codegens to {"$regex": ..,
+            //   "$options": ..}; wrapping the whole thing correctly produces {"$not": {"$regex":
+            //   .., "$options": ..}} (see `fieldless_regex`).
+            // - A query that isn't a single field-scoped operator expression at all: $and/$or
+            //   (whose value is an array of conditions, see `not_and`/`not_or`) or the `False`
+            //   sentinel's {"_id": ..., "$expr": false} (see `constant_false`). These don't have
+            //   a single, fully-specified valid $match shape when negated, so wrapping the whole
+            //   document here is a safe placeholder.
+            bson => bson!({ "$not": bson }),
+        })
     }
 
     fn codegen_match_logical_operator(
--- a/mongosql/src/codegen/match_query.rs
+++ b/mongosql/src/codegen/match_query.rs
@@ -76,16 +76,7 @@ impl MqlCodeGenerator {
                     _ => bson!({ "$not": { key: value } }),
                 }
             }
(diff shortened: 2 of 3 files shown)


mongosql/src/codegen/match_query.rs
```diff
@@ -26,6 +26,7 @@ impl MqlCodeGenerator {
         match q {
             Or(v) => self.codegen_match_logical_operator("$or", v),
             And(v) => self.codegen_match_logical_operator("$and", v),
+            Not(inner) => self.codegen_match_logical_operator("$not", vec![*inner]),
```
jchemburkar: The design mentions "Implementation wise it should look similar to the codegen NotIn codegen." It looks like that outputs the form of `{ field: { "$not": { "$in": [...] } } }` ([code ref](https://github.com/mongodb/mongosql/blob/25850304b10f90d3b8cfc07197df7704d648b31f/mongosql/src/codegen/match_query.rs#L54)). I think we want to produce agg that looks more like: `{ field: {"$not": { inner }}}`? Though if inner is a list, we might need to make this slightly more complicated.

a verification of what I'm thinking:
```
test> db.jptest.insertOne({_id: 0, age: 5})
{ acknowledged: true, insertedId: 0 }
test> db.jptest.insertOne({_id: 1, age: 15})
{ acknowledged: true, insertedId: 1 }
test> db.jptest.aggregate([{$not: [{"age": {"$gt": 10}]}])
Uncaught:
SyntaxError: Unexpected token, expected "," (1:48)

> 1 | db.jptest.aggregate([{$not: [{"age": {"$gt": 10}]}])
    |                                                 ^
  2 |

test> db.jptest.aggregate([{$not: [{"age": {"$gt": 10}}]}])
MongoServerError[Location40324]: Unrecognized pipeline stage name: '$not'
test> db.jptest.aggregate([{$match: {$not: [{"age": {"$gt": 10}}]}}])
MongoServerError[BadValue]: unknown top level operator: $not
test> db.jptest.aggregate([{$match: {"age": [{"$not": {"$gt": 10}}]}}])

test> db.jptest.aggregate([{$match: {"age": {"$not": {"$gt": 10}}}}])
[ { _id: 0, age: 5 } ]
```

jchemburkar (changes requested):
I think we may want to wrap the $not slightly differently -- let me know if this makes sense, happy to brainstorm if you think I'm on the right track!

mongosql/src/codegen/match_query.rs
mattChiaravalloti: Agreed! Importantly, as the error messages demonstrate, `$not` is _not_ a valid top-level `$match` operator like `$and`, `$or`, and `$nor` so we should be codegenning this as Jon is proposing here -- nesting it around the inner operation on a field ref. You can actually do something like this as well:
```
Enterprise test> db.car.aggregate([{$match: {a: {$not: {$exists: true, $nin: [null, false, 0]}}}}])
[
  ...
]
```
Where multiple expressions can be negated at once by providing multiple arguments to `$not`.

jpowell-mongo: Yeah, let me revisit this. Thanks for raising. I was trying to reconcile the docs for the $not query predicate vs. $not expression operator.

mongosql/src/codegen/test/match_query.rs
```diff
@@ -214,6 +234,181 @@ mod match_constant_false {
...
+            air::MatchLanguageComparison {
+                function: air::MatchLanguageComparisonOp::Gt,
+                input: None,
+                arg: air::LiteralValue::Integer(10),
+            }
+        )))
+    );
+
+    // Negating an $and/$or that spans multiple different fields
+    test_codegen_match_query!(
```
jchemburkar, line 267: I think for field scoped operators the changes made look good! However for this test case I think we may need to work around it (this is what I was alluding to with "if inner is a list").

This codegen is invalid, I think:
```
db.jptest.insertMany([
  { _id: 1, age: 5 },
  { _id: 2, age: 15 },
  { _id: 3, age: 25 }
])

db.jptest.aggregate([
      {
        $match: {
          $not: {
            $and: [
              { age: { $gt: 10 } },
              { name: { $eq: "Alice" } }
            ]
          }
        }
      }
    ]).toArray()
```

I think for $not where the inner is $and, instead of `{ $match: { $not: { $and: [...] } } }` we want to generate `{ $match: { $nor: [ { $and: [...] } ] } }`. For $or, we can simply use nor directly on the list of $or's inner.

I looked at the neighboring enum types and think we might be safe there, but worth validating. I couldn't figure out a valid query that would lead to a type / comparison / regex where the input is None (though I was suspicious given it is an option) and I think ElemMatch is actually fine to be negated / field scoped? So it might be limited to these two.

Let me know if this makes sense / you agree!

jchemburkar (changes requested):
Looking pretty good! I had two last cases I think may be worth further investigation, let me know if you have any questions.

mongosql/src/codegen/test/match_query.rs
jpowell-mongo: This is a good point, and I was split during implementation on if I should consider code-genning to $nor.

You're right in that $nor returns TRUE is ALL of the elements of it's inner expression are false. So it should _think_ that this code gen is reasonable - but let me do some testing. 

I wasn't sure if $nor is considered a more general version of $not if you just have a single expression rather than an array. 

What I mean it:

NOT (x AND Y) being equal to NOR (x AND Y).

jpowell-mongo: You're right here. I'm going to rewrite the handling for $and and $or.

The definition in the docs:
"[$nor](https://www.mongodb.com/docs/manual/reference/operator/query/nor/#mongodb-query-op.-nor) performs a logical NOR operation on an array of one or more query predicates and selects the documents that fail all the query predicates in the array." makes it seem like $nor is the logical opposite of $and. 

I think nesting is probably the best option. I want to do some tests to validate my understanding of the operator.

jpowell-mongo: Updated here:
https://github.com/mongodb/mongosql/pull/160/changes/e4791b90b62783fba9c3bbb71a2d870072b031de

I'm still not 100% convinced that $nor is the _exact_ negation of the $or operator, I just need to test stuff out in mongosh, but the translations do seem to work in mongosh

mongosql/src/air/definitions.rs
```diff
@@ -720,6 +720,7 @@ pub struct Reduce {
 pub enum MatchQuery {
```
mattChiaravalloti, line 720: **[thought]** A more invasive way to address the challenge we face here is to change this type tree. It seems the only valid top-level keys in match language are fields, `$and`, `$or`, and `$nor`. All other match-language operators can only appear as descendants from those (or other nested match-language operators). Our type tree here is, therefore, invalid.

Reworking this type hierarchy may be out of scope for this ticket, so we don't need to address it in this PR (or even this project, since you've implemented a workaround here that _should_ be sufficient). Maybe we can file a follow-up ticket so we can isolate this pretty big shift to its own unit of work later. A tech-debt ticket, maybe.

My ultimate goal is to do away with the work-around of inspecting the actual generated bson in favor of relying on the types representing valid agg pipeline shapes. We set ourselves up incorrectly here a long time ago, it seems!

mongosql/src/codegen/test/match_query.rs
```diff
@@ -214,6 +234,181 @@ mod match_constant_false {
...
+            air::MatchLanguageComparison {
+                function: air::MatchLanguageComparisonOp::Gt,
+                input: Some("age".to_string().into()),
+                arg: air::LiteralValue::Integer(10),
+            }
+        )))
+    );
+
+    test_codegen_match_query!(
+        binary_comparison,
```
mattChiaravalloti: **[question]** What does this name mean?

mongosql/src/codegen/match_query.rs
```diff
@@ -59,6 +60,44 @@ impl MqlCodeGenerator {
         }
     }
 
+    /// Codegens `$not` scoped to a field's operator expression (mirroring how `NotIn` wraps
+    /// `$in`), since `$not` — unlike `$and`/`$or`/`$nor` — is not a valid top-level `$match`
+    /// boolean operator.
+    /// For
```
mattChiaravalloti:
```suggestion
```
[suggestion] Did you mean to remove this line or finish this sentence?

mongosql/src/codegen/match_query.rs
```diff
@@ -59,6 +60,44 @@ impl MqlCodeGenerator {
...
+            air::MatchQuery::Or(v) => self.codegen_match_logical_operator("$nor", v),
+            air::MatchQuery::And(v) => {
+                self.codegen_match_logical_operator("$nor", vec![air::MatchQuery::And(v)])
+            }
+            _ => {
+                let inner_codegen = self.codegen_match_query(inner)?;
+                Ok(match inner_codegen {
+                    // Field-scoped operator, e.g. {"age": {"$gt": 10}} -> {"age": {"$not": {"$gt": 10}}}
+                    Bson::Document(doc) if doc.len() == 1 => {
+                        let (key, value) = doc.into_iter().next().expect("checked len == 1");
```
mattChiaravalloti: **[suggestion]** Can we use a more informative panic message here? Something like "match document expected to have 1 key but had 0" or similar?

mattChiaravalloti (approved):
Looks good! I have a few small nits to address before this merges but they are very minor and ultimately non-blocking if you choose not to address them. I have one larger idea that we could pursue as a follow-up tech-debt ticket. For now, this implementation should be sufficient for our purposes!

jchemburkar (approved):
lgtm pending matt's comments!

mongosql/src/codegen/test/match_query.rs
jpowell-mongo: This needs to be renamed to binary operator. I'll update the test to be x > 10 rather than just the > 10. Thanks for the call out

mongosql/src/codegen/match_query.rs
jpowell-mongo: Will remove.

mongosql/src/codegen/match_query.rs
jpowell-mongo: Will update.

mongosql/src/air/definitions.rs
jpowell-mongo: Created this investigation ticket based on this comment:
https://jira.mongodb.org/browse/SQL-3323

mongosql/src/codegen/match_query.rs
jpowell-mongo: Addressed here:
https://github.com/mongodb/mongosql/pull/160/changes/02473c43fd92489f40113137889265cdf87a2cd3

mongosql/src/codegen/match_query.rs
jpowell-mongo: Updated this comment here:
https://github.com/mongodb/mongosql/pull/160/changes/02473c43fd92489f40113137889265cdf87a2cd3

mongosql/src/codegen/test/match_query.rs
jpowell-mongo: Added an updated test and named it "negate_binary_comparison" here:
https://github.com/mongodb/mongosql/pull/160/changes/02473c43fd92489f40113137889265cdf87a2cd3

jpowell-mongo:
evergreen retry

PR #161, opened 2026-06-29
SQL-3286: Fix accumulator functions to return correct schemas

mattChiaravalloti:
This PR addresses the underlying bug that affected the query mentioned in the associated ticket. The ticket proposed a seemingly viable solution that proved to address a symptom rather than the cause. See the ticket for more details. Be sure to read the comments.

The issue is that several accumulator (aka `AggregateFunction`) expressions returned their `arg_schema` as their own schema. This was invalid for `FIRST`, `LAST`, `MIN`, and `MAX` since it allowed `Missing` to propagate through as a result of those accumulations. Those four always upconvert missing to null, so the schema code was updated accordingly. Additionally, I noticed that `MERGE_DOCUMENTS` was doing the same thing -- returning its `arg_schema` -- so I updated it to return the correct schema. In this case, that means it always returns a document (it never returns missing or even null, nullish arguments are simply ignored).

The scalar function `MERGE_OBJECTS` was already implemented correctly, so I refactored the code to reuse `MERGE_OBJECTS`'s schema implementation for `MERGE_DOCUMENTS`. As a drive-by, I also fixed it to allow for nullish arguments (this was required since I wanted to use the type-constraint spec tests to ensure the changes were valid but those tests were blocked on supporting nullish args). The nullish-args fix was trivial, as you'll see in the code and tests.

I elected to not add an e2e test for the query in the ticket, instead relying on unit tests to demonstrate this works as expected. I manually confirmed the query in the ticket now works. I can probably be convinced to add an e2e test if someone can come up with good motivation for why, but I personally could not justify one at this time.

### DIFF
--- a/mongosql/src/mir/schema/mod.rs
+++ b/mongosql/src/mir/schema/mod.rs
@@ -1616,10 +1616,7 @@ trait SqlFunction {
                     .unwrap_or_else(|| EMPTY_DOCUMENT.clone()),
                 Schema::Document(d) => Schema::Document(d),
                 Schema::Any => Schema::Document(Document::any()),
-                Schema::Unsat
-                | Schema::Missing
-                | Schema::Atomic(_)
-                | Schema::Array(_) => {
+                Schema::Unsat | Schema::Missing | Schema::Atomic(_) | Schema::Array(_) => {
                     unreachable!()
                 }
             })
--- a/mongosql/src/mir/schema/mod.rs
+++ b/mongosql/src/mir/schema/mod.rs
@@ -1607,7 +1607,7 @@ trait SqlFunction {
         // union all AnyOf arg_schemas into union Document schemata
         arg_schemas
             .iter()
-            .filter(|s| state.check_satisfies(s, &ANY_DOCUMENT.clone()))
+            .filter(|s| state.check_satisfies(s, &ANY_DOCUMENT))
             .cloned()
             .map(|s| match s {
                 Schema::AnyOf(ao) => ao
--- a/mongosql/src/mir/schema/mod.rs
+++ b/mongosql/src/mir/schema/mod.rs
@@ -8,9 +8,9 @@ use crate::{
     },
     schema::{
         Atomic, Document, ResultSet, Satisfaction, Schema, SchemaEnvironment, ANY_ARRAY,
-        ANY_ARRAY_OR_NULLISH, ANY_DOCUMENT, BOOLEAN_OR_NULLISH, DATE_OR_NULLISH, EMPTY_DOCUMENT,
-        INTEGER_LONG_OR_NULLISH, INTEGER_OR_NULLISH, NULLISH, NUMERIC, NUMERIC_OR_NULLISH,
-        STRING_OR_NULLISH,
+        ANY_ARRAY_OR_NULLISH, ANY_DOCUMENT, ANY_DOCUMENT_OR_NULLISH, BOOLEAN_OR_NULLISH,
+        DATE_OR_NULLISH, EMPTY_DOCUMENT, INTEGER_LONG_OR_NULLISH, INTEGER_OR_NULLISH, NULLISH,
+        NUMERIC, NUMERIC_OR_NULLISH, STRING_OR_NULLISH,
     },
     set,
     util::unique_linked_hash_map::UniqueLinkedHashMap,
@@ -934,14 +934,7 @@ impl AggregationFunction {
                 }
                 arg_schema.upconvert_missing_to_null()
             }
-            MergeDocuments => {
-                self.schema_check_fixed_args(
-                    state,
-                    from_ref(&arg_schema),
-                    from_ref(&ANY_DOCUMENT),
-                )?;
-                arg_schema
-            }
+            MergeDocuments => self.schema_check_merge_objects(state, from_ref(&arg_schema))?,
             Sum => self.get_arithmetic_schema(state, &[arg_schema])?,
         })
     }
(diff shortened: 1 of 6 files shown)


mongosql/src/mir/schema/mod.rs
```diff
@@ -924,24 +924,17 @@ impl AggregationFunction {
                 Schema::Atomic(Atomic::Integer),
                 Schema::Atomic(Atomic::Long)
             ]),
-            First | Last => arg_schema,
+            First | Last => arg_schema.upconvert_missing_to_null(),
```
mattChiaravalloti, line 927: This is the core of the fix for `FIRST`/`LAST`/`MIN`/`MAX`. `upconvert_missing_to_null` replaces `Missing` in a schema with `Atomic::Null`. This is valid because the operators `$first`, `$last`, `$min`, and `$max` will return null even if the argument to them is missing on every invocation. Example:
```
Enterprise test> db.foo.aggregate()
[
  { _id: 0, a: 1, b: 10 },
  { _id: 1, a: 1, b: 50 },
  { _id: 2, a: 2, b: 1 },
  { _id: 3, a: 2, b: 5 }
]
Enterprise test> db.foo.aggregate([{$group: {_id: "$a", first_missing: {$first: "$c"}, last_missing: {$last: "$c"}, min_missing: {$min: "$c"}, max_missing: {$max: "$c"}}}])
[
  {
    _id: 1,
    first_missing: null,
    last_missing: null,
    min_missing: null,
    max_missing: null
  },
  {
    _id: 2,
    first_missing: null,
    last_missing: null,
    min_missing: null,
    max_missing: null
  }
]
```

mongosql/src/mir/schema/mod.rs
```diff
@@ -1599,6 +1592,51 @@ trait SqlFunction {
         }
     }
 
+    /// Schema checks the arguments to a $mergeObjects-backed function (the
+    /// scalar MergeObjects and the aggregation MergeDocuments) and returns the
+    /// resulting Document schema. Since $mergeObjects ignores nullish operands
+    /// and returns the empty document when all operands are nullish, the result
+    /// is always a Document.
+    #[allow(clippy::manual_try_fold)]
+    fn schema_check_merge_objects(
```
mattChiaravalloti, line 1601: This function is _almost_ entirely just refactored from the `ScalarFunction` impl (you'll notice its deletion lower in the file). The only difference is it is updated to supported nullish arguments while ignoring those arguments in the output. See the 3 comments below for the only changes I made to this function.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1599,6 +1592,51 @@ trait SqlFunction {
...
+        &self,
+        state: &SchemaInferenceState,
+        arg_schemas: &[Schema],
+    ) -> Result<Schema, Error> {
+        self.schema_check_variadic_args(state, arg_schemas, ANY_DOCUMENT_OR_NULLISH.clone())?;
```
mattChiaravalloti, line 1606: Change 1: `ANY_DOCUMENT` -> `ANY_DOCUMENT_OR_NULLISH`. This allows nullish arguments to `MERGE_OBJECTS`/`MERGE_DOCUMENTS`.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1599,6 +1592,51 @@ trait SqlFunction {
...
+        // union all AnyOf arg_schemas into union Document schemata
+        arg_schemas
+            .iter()
+            .filter(|s| state.check_satisfies(s, &ANY_DOCUMENT.clone()))
```
mattChiaravalloti: Change 2: Filter out any non-document schemas since they are irrelevant to the output schema.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1599,6 +1592,51 @@ trait SqlFunction {
...
+            .cloned()
+            .map(|s| match s {
+                Schema::AnyOf(ao) => ao
+                    .into_iter()
+                    .reduce(Schema::document_union)
+                    .unwrap_or_else(|| EMPTY_DOCUMENT.clone()),
+                Schema::Document(d) => Schema::Document(d),
+                Schema::Any => Schema::Document(Document::any()),
```
mattChiaravalloti, line 1618: Change 3: (drive-by bug fix for relaxed schema checking mode) Map `Any` to any document since this is reachable code.

mongosql/src/mir/schema/test/expressions/aggregate.rs
```diff
@@ -426,6 +519,67 @@ mod merge_documents {
...
+            function: AggregationFunction::MergeDocuments,
+            arg: Box::new(Expression::Reference(("bar", 0u16).into())),
+            distinct: false,
+            arg_is_possibly_doc: Satisfaction::Not,
+        }),
+        schema_env = map! {("bar", 0u16).into() => Schema::Any},
+        schema_checking_mode = SchemaCheckingMode::Relaxed,
+    );
+
+    // In relaxed mode, it is acceptable for the argument's schema can to only
```
jchemburkar:
```suggestion
    // In relaxed mode, it is acceptable for the argument's schema to only
```

mongosql/src/mir/schema/mod.rs
jchemburkar:
```suggestion
            .filter(|s| state.check_satisfies(s, &ANY_DOCUMENT))
```
nit (I think this works)

jchemburkar (approved):
LGTM! I concur with not _needing_ to add the e2e test

mongosql/src/mir/schema/mod.rs
bucaojit: Thanks for the example, it made the change much clearer 👍

bucaojit (approved):
LGTM

mongosql/src/mir/schema/mod.rs
mattChiaravalloti: great catch, you're right

PR #162, opened 2026-06-30
SQL-3288: Add spec tests for higher order functions and array functions

mattChiaravalloti:
This PR introduces the new `spec/query`, `spec/rewrite`, and `spec/type_constraint` tests for the new higher order functions and array functions, as described by the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.7k998qpudnu) for [SQL-3076](https://jira.mongodb.org/browse/SQL-3076).


mongosql/src/internal_spec_test/mod.rs
```diff
@@ -151,12 +151,12 @@ pub fn load_file_paths(dir: PathBuf) -> Result<Vec<String>, Error> {
...
 pub fn parse_rewrite_yaml(path: &str) -> Result<RewriteYamlTest, Error> {
-    let mut f = fs::File::open(path).map_err(|e| Error::InvalidFile(format!("{e:?}")))?;
+    let mut f = fs::File::open(path).map_err(|e| Error::InvalidFile(format!("{path}: {e:?}")))?;
     let mut contents = String::new();
     f.read_to_string(&mut contents)
-        .map_err(|e| Error::CannotReadFileToString(format!("{e:?}")))?;
+        .map_err(|e| Error::CannotReadFileToString(format!("{path}: {e:?}")))?;
     let yaml: RewriteYamlTest = serde_yaml::from_str(&contents)
-        .map_err(|e| Error::CannotDeserializeYaml(format!("{e:?}")))?;
+        .map_err(|e| Error::CannotDeserializeYaml(format!("{path}: {e:?}")))?;
```
mattChiaravalloti, line 159: I had a typo in one of the test files and could not figure out which, so I updated these errors to include the file name. Very helpful for now and for the future 👍

testdata/spec_tests/query_tests/array_functions.yml
```diff
@@ -0,0 +1,120 @@
+dataset:
+  - db: "spec_query_array_functions"
+    collection:
+      name: "array_default_numeric_data"
```
mattChiaravalloti, line 4: I tried to minimize the number of datasets so I reused this one multiple times, as well as `array_default_boolean_data`.

testdata/spec_tests/query_tests/higher_order_functions.yml
```diff
@@ -0,0 +1,118 @@
...
+                    - bsonType: !!str "null"
+              - bsonType: !!str "null"
+          i:
+            anyOf:
+              - bsonType: "int"
+              - bsonType: !!str "null"
+
+  - db: "spec_query_higher_order_functions"
+    collection:
+      name: "unambiguous"
```
mattChiaravalloti, line 89: In the same spirit of minimizing the number of datasets, I used a single dataset for the disambiguation tests for all three higher order functions even though only `REDUCE` uses the `value` field. This leads to 2 extra rows in the `result` sets for `FILTER` and `MAP`, but I think that's a fine trade-off.

tests/spec_tests/query_tests/array_functions.yml
```diff
@@ -0,0 +1,162 @@
...
+    result:
+      - { '': { "a": [], "v": [] } }
+      - { '': { "a": [1], "v": [1] } }
+      - { '': { "a": [1, 2, 3], "v": [1, 2, 3] } }
+      - { '': { "a": [1, null, 3], "v": [1, 3] } }
+      - { '': { "a": null, "v": null } }
+      - { '': { "v": null } }
+
+  # ARRAY_REMOVE removes all instances of the specified element from the input array
+  # Note that the removal predicate uses SQL = semantics so NULL elements are removed since NULL = x is NULL.
```
mattChiaravalloti: What do we think of these semantics? Its certainly most straightforward and SQL-like to do it this way, but I can see a reasonable person arguing for retaining `NULL` values as well, so I could be convinced the other way if that's what the team thinks is best.

testdata/spec_tests/query_tests/array_functions.yml
```diff
@@ -0,0 +1,120 @@
...
+        additionalProperties: false
+        properties:
+          _id:
+            bsonType: "int"
+          a:
+            anyOf:
+              - bsonType: "array"
+                items:
+                  anyOf:
+                    - bsonType: "int"
```
mattChiaravalloti: Whoops, copy-paste mistake. I'll update this schema to match the expected schema `{x: {y: int/null/missing}}/null`.

tests/spec_tests/type_constraint_tests/higher_order_functions.yml
mattChiaravalloti, line 1: These tests deviate very slightly from the design doc. I said in the design that the spec/type_constraint tests would impose the restrictions on `this` and `value`, but now at implementation time I decided against that. The test runner is not well suited for that type of testing right now. To extend it to support that type of testing would be very non-trivial work. I lean towards just sticking with the (already planned) unit testing for the described constraints on `this` and `value`. These tests, as written, impose the actual expected top-level type constraints for the arguments to these function, which is in line with how these types tests are meant to be written.

tests/spec_tests/query_tests/array_functions.yml
bucaojit: I checked Snowflake and Postgres, they seem to retain the `NULL` in `ARRAY_REMOVE()`.   They also have it so `NULL` can be the specified element that is removed.  Not sure why they went this route but maybe we can consider matching their behavior.

bucaojit (approved):
Tests cover what was in the design.  I had one response regarding `ARRAY_REMOVE`.

tests/spec_tests/query_tests/array_functions.yml
mattChiaravalloti: Great find. I guess if even Postgres is "breaking" standard SQL semantics for this, it's safe for us to do as well. I wonder if it would be better to do a complex rewrite that captures these semantics, or if we should capture these semantics internally by having a null-safe flag added to ast structs. We'll have to brainstorm this a bit. For now for this PR, I'll update this spec test to demonstrate what we want this result set to be and I'll leave the existing rewrite as-is. We'll revisit this during later tickets.

Thanks for the shout out. Adding a comment to later tickets to track this.

jpowell-mongo (approved):

PR #166, opened 2026-07-06
SQL-3268: Migrate and update (NOT) IN tests for implicit type conversion of extended JSON strings

mattChiaravalloti:
This PR migrates the spec/query tests for `IN` and `NOT IN` that intended to cover implicit type conversion of extended JSON encoded strings into the e2e test suite, and updates those tests to cover the intended behavior on both sides of the operators. See the ticket for more info.

As part of this change, I realized that we were unconditionally codegenning `air::MqlOperator::NotIn` to `$nin`, which was invalid. I updated the `air::MqlOperator::NotIn` codepath to mimic what `air::SqlOperator::NotIn` does. That is, it now codegens as `{$not: {$in: ... }}` instead of `{$nin: ...}`. That was a quick drive-by fix! I considered filing a separate bug ticket, but I figured why block these new e2e tests on a bug ticket when it's just essentially a one-liner change.

This is a tech-debt ticket.

### DIFF
--- a/tests/spec_tests/query_tests/scalar_functions.yml
+++ b/tests/spec_tests/query_tests/scalar_functions.yml
@@ -660,26 +660,12 @@ tests:
       - {'': {'a': 1, 's': 'hello'}}
       - {'': {'a': 2, 's': 'world'}}
 
-  - description: NOT IN operator with $numberInt encoded field correctness test
-    current_db: spec_query_scalar_functions
-    query: "SELECT _id, n FROM in_expressions WHERE n NOT IN (1, 2, 4, null)"
-    result:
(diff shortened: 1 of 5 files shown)


tests/spec_tests/query_tests/scalar_functions.yml
mattChiaravalloti, line 1: Note that neither of these tests actually included coverage for implicit type conversion of extended JSON encoded strings since neither query contained such string values! That's why new tests in the e2e file look slightly different!

jpowell-mongo (approved):
Thank you for catching this! Apologies for the miss, I should have also checked the mql codegen path along with the sql codegen path.

nicholascioli (approved):
LGTM!

PR #167, opened 2026-07-07
SQL-3343: Rework FLATTEN field-access to avoid repeated work

jameseh96:
Previously, algebrize_flattened_field_path was recursive.

For a path a.b.c, the call chain would look like:

```
algebrize_flattened_field_path(key, [a,b,c])
> construct_field_access_expr( algebrize_flattened_field_path(key,[a,b]), "c" )
  > construct_field_access_expr( algebrize_flattened_field_path(key,[a]), "b" )
      > construct_field_access_expr( algebrize_flattened_field_path(key,[]), "a" )
          > construct_field_access_expr( Reference(key), "a" )
```

Each `construct_field_access_expr` call evaluated

```
FieldAccess{...}.schema(state)
```

to decide nullability.

FieldAccess::schema is itself recursive: it calls self.expr.schema(), which will call `fa.schema(state)` on its contents.

The base case Expression::Reference does `env.get(key).cloned()`, cloning the root schema.

So each layer of `construct_field_access_expr` would eventually clone the root schema, which could be very large. Further, it would then repeat work done by a lower call:

1. clone root -> a
2. clone root -> a -> b
3. clone root -> a -> b -> c

This could be quite costly.

Change to an iterative approach, tracking the current schema as the call descends through the path.

For the sample_analytics FLATTEN benchmark this cuts the project field-access build from ~11.4s to ~3.3s, -71%.

### DIFF
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -898,22 +898,11 @@ impl<'a> Algebrizer<'a> {
                 }
 
                 let mut project_expression = UniqueLinkedHashMap::new();
-                let mut sub_schema_env = SchemaEnvironment::new();
-                sub_schema_env.insert(key.clone(), schema);
-                let path_algebrizer = Algebrizer::with_schema_env(
-                    self.current_db,
-                    sub_schema_env,
-                    self.catalog,
-                    self.scope_level,
-                    self.schema_checking_mode,
-                    self.allow_order_by_missing_columns,
-                    *self.clause_type.borrow(),
-                );
                 project_expression
                     .insert_many(field_paths_copy.into_iter().map(|path| {
                         (
                             path.join(separator),
-                            path_algebrizer.algebrize_flattened_field_path(key.clone(), path),
+                            Self::build_flattened_field_access(key.clone(), &path, &schema),
                         )
                     }))
                     .map_err(|e| Error::DuplicateDocumentKey(e.get_key_name()))?;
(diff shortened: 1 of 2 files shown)


jameseh96:
`cargo test` passes locally, but I'm not familiar with the Atlas SQL workflow.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -998,17 +987,30 @@ impl<'a> Algebrizer<'a> {
...
-                    path.last().unwrap().to_string(),
-                )
-                .unwrap(),
+    // Construct the nested FieldAccess expression for a single flattened field path in
+    // O(path_depth) time. It tracks the current schema while traversing the path, computing
+    // is_nullable at each level directly from the sub-schema.
+    fn build_flattened_field_access(
+        key: Key,
+        path: &[String],
+        root_schema: &crate::schema::Schema,
```
mattChiaravalloti:
```suggestion
        root_schema: &schema::Schema,
```
[nit] We already import `crate::schema::self` at the top of the file, so we should not need to prefix this with `crate::` here.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -998,17 +987,30 @@ impl<'a> Algebrizer<'a> {
...
+    ) -> mir::Expression {
+        let mut expr = mir::Expression::Reference(mir::ReferenceExpr { key });
+        // Avoid cloning the (potentially huge) root schema: narrow from a borrow for the first
+        // field, then from the owned narrowed sub-schema for subsequent fields.
+        let mut current_schema: Option<crate::schema::Schema> = None;
```
mattChiaravalloti:
```suggestion
        let mut current_schema: Option<schema::Schema> = None;
```

mattChiaravalloti (approved):
This is a great idea and a useful change. Everything is passing on evergreen so we should be confident this is as correct as the previous solution. I'm marking this as approved but we'll hold off on merging this until after the team triages the associated ticket later on today.

mattChiaravalloti:
@jameseh96  The team triaged this and decided to move forward with it. I tagged another team member to give it a second look. When you get 2 approvals and address any outstanding comments, we can get this merged.

It looks like your commits aren't signed, either. If you are able, can you sign the commits and re-push? Feel free to force push the signed commits to this PR when you sign them.

bucaojit (approved):
Nice work @jameseh96 . 
LGTM

jameseh96:
Comments addressed, commits signed, re-pushed!

mongosql/src/algebrizer/definitions.rs
jameseh96: Done

mongosql/src/algebrizer/definitions.rs
jameseh96: Done

PR #168, opened 2026-07-07
[SQL-3300] Introduce NOT and Field Access to match language optimizer

jpowell-mongo:
## Summary

This pull request extends the match_language_rewriter optimizer to support translating NOT and FieldAccesses into match language.

This means that field accesses can appear in match language:
```rust
./target/debug/mongosql-cli "SELECT * FROM foo WHERE is_active AND is_verified"

/// Generated Pipeline
[
    { "$match": { "is_active": { "$eq": true } } },
    { "$match": { "is_verified": { "$eq": true } } },
    { "$project": { "foo": "$$ROOT", "_id": 0 } },
]
```

And the NOT operator is also translated to match language.
```rust
 ./target/debug/mongosql-cli "SELECT * FROM foo WHERE NOT (is_active AND is_verified)"

/// Generated Pipeline
[
    { "$match": { "$nor": [{ "$and": [{ "is_active": { "$eq": true } }, { "is_verified": { "$eq": true } }] }] } },
    { "$project": { "foo": "$$ROOT", "_id": 0 } },
]
```

## Implementation

Implementation is mainly extending the match arms in the match language optimizer.

## Testing
Introduced new unit tests in the match language rewriter, and a spec query tests to validate that field accesses can be used in queries.

### DIFF
--- a/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
+++ b/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
@@ -519,7 +519,7 @@ test_rewrite_to_match_language!(
         op: MatchLanguageLogicalOp::Not,
         args: vec![MatchQuery::Comparison(MatchLanguageComparison {
             function: MatchLanguageComparisonOp::Eq,
-            input: Some(mir_field_path("foo", vec!["int"])),
+            input: Some(mir_field_path("foo", vec!["bool"])),
             arg: LiteralValue::Boolean(true),
             cache: SchemaCache::new(),
         })],
--- a/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
+++ b/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
@@ -172,7 +172,7 @@ fn comp_expr() -> Expression {
         ScalarFunction::Lt,
         vec![
             *mir_field_access("foo", "int", true),
-            Expression::Literal(LiteralValue::Integer(10)),
+            *mir_field_access("bar", "int", true),
         ],
     ))
 }
--- a/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
+++ b/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
@@ -598,21 +598,6 @@ test_rewrite_to_match_language!(
     input = filter_stage(*mir_field_access_multi_part("foo", vec!["a", "b"], true))
 );
 
-// A field access whose base is a scalar function (not a Reference) cannot be
-// converted to a FieldPath, so it stays in expr language.
-test_rewrite_to_match_language_no_op!(
-    field_access_not_convertible_to_field_path_is_noop,
-    filter_stage(Expression::FieldAccess(FieldAccess {
-        expr: Box::new(Expression::ScalarFunction(ScalarFunctionApplication {
-            function: ScalarFunction::Upper,
-            args: vec![*mir_field_access("foo", "str", true)],
-            is_nullable: true,
-        })),
-        field: "x".to_string(),
-        is_nullable: true,
-    }))
-);
-
 // A boolean field access composes inside a logical operation.
 test_rewrite_to_match_language!(
     rewrite_boolean_field_access_in_conjunction,
--- a/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
+++ b/mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
@@ -10,7 +10,7 @@ use crate::{
     },
     schema::{Atomic, Document, Schema, SchemaEnvironment},
     set, unchecked_unique_linked_hash_map,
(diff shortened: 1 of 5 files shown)


tests/spec_tests/query_tests/where.yml
```diff
@@ -40,3 +40,14 @@ tests:
     result:
       - {'items': {'_id': 0, 'foo_bar': 2}}
       - {'items': {'_id': 4, 'foo_bar': 2}}
+  - description: WHERE with BOOLEAN field access
+    query: "SELECT * FROM [{'is_active': true, 'id': 1}, {'is_active': true, 'id': 2}, {'is_active': false, 'id': 3}] AS arr WHERE is_active"
+    current_db: test
+    result:
+      - {'arr': {'is_active': true, 'id': 1}}
+      - {'arr': {'is_active': true, 'id': 2}}
+  - description: WHERE with NEGATED BOOLEAN field access
```
mattChiaravalloti, line 51:
```suggestion

  - description: WHERE with BOOLEAN field access
    query: "SELECT * FROM [{'is_active': true, 'id': 1}, {'is_active': true, 'id': 2}, {'is_active': false, 'id': 3}] AS arr WHERE is_active"
    current_db: test
    result:
      - {'arr': {'is_active': true, 'id': 1}}
      - {'arr': {'is_active': true, 'id': 2}}

  - description: WHERE with NEGATED BOOLEAN field access
```
[nit] For consistency with the rest of the file, and readability, can we put blank lines between tests?

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
```diff
@@ -498,3 +498,144 @@ test_rewrite_to_match_language_no_op!(
...
+        ))],
+    )))
+);
+
+// The inner comparison is not rewritable, so the whole NOT stays in expr language.
+test_rewrite_to_match_language_no_op!(
+    cannot_rewrite_not_over_comparison,
+    filter_stage(Expression::ScalarFunction(ScalarFunctionApplication::new(
+        ScalarFunction::Not,
+        vec![comp_expr()],
```
mattChiaravalloti, line 584: Not sure this is the best one to go with for now since we're soon going to support exactly the shape of expression `comp_expr` returns! Maybe let's update `comp_expr` to compare two fields (not rewritable) instead of a field and a literal (soon-to-be rewritable).

mattChiaravalloti (commented):
Looks great! One nit in the yaml file, and one minor suggestion for the longevity of one of the new unit tests. Beyond that, nothing much to change here!

Thanks for sharing the new generated pipeline. It highlights yet another thing we should consider: `merge_neighboring_matches` ([link](https://github.com/mongodb/mongosql/blob/3a4d06271a8d06954f0bd56b86d8a6d363f431ff/mongosql/src/mir/optimizer/mod.rs#L72))! Given the description written in the `merge_neighboring_matches` optimization, we should probably file a ticket to update that optimization to be capable of merging neighboring match-language matches into a single stage. That's out of scope for this ticket, hence me suggesting filing a follow-up. I'd suggest filing that follow-up as part of this epic since it is likely worth doing before considering this "done" since it may have performance implications.

mattChiaravalloti (approved):
Whoops -- meant to approve! The only two actionable comments are optional and trivial, so there's no need to not approve now!

jpowell-mongo:
> Looks great! One nit in the yaml file, and one minor suggestion for the longevity of one of the new unit tests. Beyond that, nothing much to change here!
> 
> Thanks for sharing the new generated pipeline. It highlights yet another thing we should consider: `merge_neighboring_matches` ([link](https://github.com/mongodb/mongosql/blob/3a4d06271a8d06954f0bd56b86d8a6d363f431ff/mongosql/src/mir/optimizer/mod.rs#L72))! Given the description written in the `merge_neighboring_matches` optimization, we should probably file a ticket to update that optimization to be capable of merging neighboring match-language matches into a single stage. That's out of scope for this ticket, hence me suggesting filing a follow-up. I'd suggest filing that follow-up as part of this epic since it is likely worth doing before considering this "done" since it may have performance implications.

Yeah this is an interesting point, because it seems like the match splitting optimizer is braking these apart! I was curious on why I kept getting two match stages instead of a single $and with the two clauses. I think this is a really solid followup. I'll log a ticket for this improvement.

mattChiaravalloti:
> Yeah this is an interesting point, because it seems like the match splitting optimizer is braking these apart! I was curious on why I kept getting two match stages instead of a single $and with the two clauses. I think this is a really solid followup. I'll log a ticket for this improvement.

It is beneficial to have match splitting break compound filters apart because it is possible that we can push part of a filter earlier in a pipeline while another part cannot. For example, consider a compound filter on a join datasource that references both sides of the join independently. Half of that filter can be pushed to the left side, and half to the right! Thanks for filing the filing the follow-up 🙏

bucaojit (approved):
LGTM

PR #171, opened 2026-07-07
SQL-3294: Update schema checking for variadic operators to require minimum argument counts

mattChiaravalloti:
This PR updates schema checking for variadic operators to require minimum argument counts, as noted in the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#bookmark=id.wzg1xfbnma1p). This set of operators includes `Add`, `And`, `Concat`, `Mul`, and `Or`. This is the set of "variadic operators" we support in the `flatten_variadics` optimization, and they also correspond to the set of functions in schema checking for which we did not include an arg-count assertion already (either directly or implicitly).

As part of this change, I updated the `IncorrectArgumentCount` error to include info on precision. Now, it reads either "required _exactly_ X, found Y" or "required _at least_ X, found Y", depending on how the error is constructed. This update to the error variant is why there are so many changes. The core change, really, is the introduction of `ensure_minimum_arg_count` and its three uses.

### DIFF
--- a/mongosql/src/mir/schema/mod.rs
+++ b/mongosql/src/mir/schema/mod.rs
@@ -27,9 +27,9 @@ use std::{
 mod errors;
 pub use errors::Error;
 
+use crate::mir::schema::errors::IncorrectArgCountPrecision;
 #[cfg(test)]
 pub(crate) use errors::ANY_SCHEMA_ADDENDUM;
-use crate::mir::schema::errors::IncorrectArgCountPrecision;
 
 mod util;
 
@@ -1737,7 +1737,7 @@ impl ScalarFunction {
                     STRING_OR_NULLISH.clone(),
                     Schema::Atomic(Atomic::String),
                 )
-            },
+            }
             // Unary arithmetic operators.
             Pos | Neg | Abs | Ceil | Floor => {
                 self.ensure_arg_count(arg_schemas.len(), 1)?;
@@ -1749,7 +1749,7 @@ impl ScalarFunction {
                 // arguments.
                 self.ensure_minimum_arg_count(arg_schemas.len(), 2)?;
                 self.get_arithmetic_schema(state, arg_schemas)
-            },
+            }
             // Arithmetic operators with fixed (two) arguments.
             Sub | Div => {
                 self.ensure_arg_count(arg_schemas.len(), 2)?;
@@ -1834,7 +1834,7 @@ impl ScalarFunction {
                     BOOLEAN_OR_NULLISH.clone(),
                     Schema::Atomic(Atomic::Boolean),
                 )
-            },
+            }
             // Computed Field Access operator when the field is not known until runtime.
             ComputedFieldAccess => {
                 self.schema_check_fixed_args(
--- a/mongosql/src/mir/schema/mod.rs
+++ b/mongosql/src/mir/schema/mod.rs
@@ -29,6 +29,7 @@ pub use errors::Error;
 
 #[cfg(test)]
 pub(crate) use errors::ANY_SCHEMA_ADDENDUM;
+use crate::mir::schema::errors::IncorrectArgCountPrecision;
 
 mod util;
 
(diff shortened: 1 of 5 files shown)


mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -125,7 +125,7 @@ mod and {
     use super::*;
 
     test_schema!(
-        and_first_arg_is_not_bool_is_error,
+        first_arg_is_not_bool_is_error,
```
mattChiaravalloti, line 128: drive-by clean up: the module is already called `and`, so there's no need to prefix each test with `and_` since that ultimately displays the tests as `mir::schema::test::expressions::scalar_functions::and::and_<test description>`. By removing the `and_` prefixes (and the `or_` and `concat_` prefixes for their modules), the output will now read more naturally: `mir::schema::test::expressions::scalar_functions::and::<test description>`.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -217,13 +217,46 @@ mod and {
         )),
         schema_env = map! {("bar", 0u16).into() => Schema::Atomic(Atomic::Null)},
     );
+
+    test_schema!(
+        more_than_two_args,
```
mattChiaravalloti, line 222: Since we allow (and even construct, as part of the `flatten_variadics` optimization) variadic `And` functions. This same concept applies to `Or` and `Concat`. I did not include new tests for `Add` or `Mul` since this "more than 2 args" behavior is already implicitly covered by a number of our unit tests for those functions.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -975,24 +1074,6 @@ mod hour {
 mod arithmetic {
     use super::*;
 
-    test_schema!(
```
mattChiaravalloti, line 978: These tests should never have been added. It seems more a consequence of implementation rather than a planned decision to support this feature. It is syntactically invalid to create an `Add` or `Mul` expression without at least 2 operands, and we should also reflect that semantically. (Note: unary addition does exist as the `Pos` operator, but that is determined syntactically and is represented distinctly in the `mir` so there is no risk here of unary addition failing this test).

tests/errors/schema.yml
```diff
@@ -1,9 +1,16 @@
...
-  - description: Error 1001 IncorrectArgumentCount
+  - description: Error 1001 IncorrectArgumentCount with Exact requirement
     query: "SELECT ROUND(a,1,5) FROM foo"
     current_db: errors_schema
     should_compile: false
-    algebrize_error: 'Error 1001: incorrect argument count for Round: required 2, found 3'
+    algebrize_error: 'Error 1001: incorrect argument count for Round: required exactly 2, found 3'
+
+  - description: Error 1001 IncorrectArgumentCount with Minimum requirement
+    query: "SELECT MAP(a, +) FROM foo"
```
mattChiaravalloti, line 9: This will ultimately be rewritten to `MAP(a, this +)` and will be rejected at semantic validation time since the arg count is too low. This is actually the motivation behind this work, as noted in the tech design!

mongosql/src/mir/schema/mod.rs
```diff
@@ -27,6 +27,7 @@ use std::{
 mod errors;
 pub use errors::Error;
 
+use crate::mir::schema::errors::IncorrectArgCountPrecision;
```
mattChiaravalloti: Moving this since it is awkwardly placed!

mongosql/src/mir/schema/mod.rs
```diff
@@ -1426,6 +1438,9 @@ trait SqlFunction {
     ///
     /// Since the argument count can vary, the required schema is a single
     /// value that's compared against each of the arguments.
+    ///
+    /// The minimum_args parameter specifies the minimum number of arguments
+    /// that are required for this variadic function.
```
mattChiaravalloti: Removing this outdated comment!

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -2955,7 +3062,7 @@ mod coalesce {
         expected_error_code = 1001,
         expected = Err(mir_error::IncorrectArgumentCount {
             name: "Coalesce",
-            required: 1,
+            required: IncorrectArgCountPrecision::Exact(1),
```
jchemburkar: question: shouldn't Coalesce be `IncorrectArgCountPrecision::Minimum(1)` ?

jchemburkar (approved):
lgtm! Just one question on correctness

mongosql/src/mir/schema/test/expressions/scalar_function.rs
mattChiaravalloti: Oh great catch, yes. I'll update this and the implementation.

EthanHardyMongo (approved):
LGTM!

PR #172, opened 2026-07-08
[SQL-3301] Support Binary Comparisons in Match Language

jpowell-mongo:
## Summary                                                                                                                                                                           
                                                                                                                                                                                    
 Extends the rewrite_to_match_language MIR optimizer to rewrite SQL comparison operators (<, <=, <>, =, >, >=) into native MQL match language (find syntax) instead of leaving them in $expr. These changes encourage index usage for more statements and will likely improve index utilization for customers.
                                                                                                                                                                                    
## Key Changes

### Rewrite To Match Language Optimizer
- Match Language Rewriter to rewrite comparison operators to match language

### Stage Movement
- Introduce logic for handling correlated subqueries when rewriting to match language. A MatchFilter using datasources from both sides of a join is now pinned above the join.

### Use Def Analysis
- Mirror existing logic for counting expr language field uses and apply it to counting field uses for match language
- A new field `noop_on_missing_key` to allow MatchFilter stages to bubble up above Unwind stages, which has an empty theta (substitution map)

### Tests
- Update Index Utilization Tests
- Unit Tests


mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
```diff
@@ -63,6 +63,63 @@ impl Visitor for PrefilterUnwindsVisitor {
     fn visit_stage(&mut self, node: Stage) -> Stage {
         let node = node.walk(self);
         match node {
+            Stage::MqlIntrinsic(MqlStage::MatchFilter(f)) => {
```
jpowell-mongo, line 66: Note:

Although this looks like net new code, this logic is a near copy of the `Stage::Filter(f) => match *f.source` match arm, but with small adjustments to work for `MqlStage::MatchFilter` instead.

test-utils/src/index/mod.rs
```diff
@@ -259,13 +259,20 @@ impl InputStage {
 }
 
 /// as_index_utilization converts an ExecutionStage.stage type into an
-/// IndexUtilization value. Only COLLSCAN and IXSCAN are valid.
+/// IndexUtilization value.
+///
+/// `IXSCAN`, `IDHACK`, and the `EXPRESS_*` family all read through an index, so
+/// they map to `IxScan`. `IDHACK` is the server's fast path for a sole `_id`
+/// equality predicate, and `EXPRESS_IXSCAN` (MongoDB 8.0+) is the equivalent
+/// fast path for other single-field index lookups; both use an index.
```
mattChiaravalloti, line 267: [praise] excellent find, thank you for expanding and documenting

mongosql/src/mir/optimizer/use_def_analysis/mod.rs
```diff
@@ -412,6 +412,14 @@ struct SubstituteVisitor {
     theta: HashMap<Key, Expression>,
     // Substitution can fail when trying to substitute a FieldPath
     failed: bool,
+    // When true, a FieldPath whose key is absent from theta is left unchanged (a no-op success)
+    // rather than failing. This mirrors how `visit_expression` treats an unbound `Reference`, and
+    // lets a MatchFilter bubble past stages with an empty theta (Unwind, Derived) the way an
+    // `$expr` Filter already does. It is left false for Sort, whose substitution must keep failing
+    // on a missing key: that failure is what prevents a Sort from reordering past a Filter or
+    // MatchFilter (which would otherwise loop forever, since `is_filter_filter_only_change` does
+    // not exempt Sort swaps).
```
mattChiaravalloti, line 421: [praise] excellent comment

[question] "... since `is_filter_filter_only_change` does not exempt Sort swaps)." _Should_ we update how that is handled to prevent `Sort` stages from being swapped above `Filter` stages (of any variety) so that we don't need to depend on this?

mongosql/src/mir/optimizer/stage_movement/test.rs
```diff
@@ -1872,63 +1926,46 @@ test_move_stage!(
...
+// MatchFilter must stay above the join, where every datasource is a plain field reference. (A
+// regular $expr Filter *can* move into the subquery, since correlated variables are valid in
+// $expr; see move_filter_into_lateral_inner_join_subquery_if_both_datasources_are_used above.)
+//
+// This models the following query, whose WHERE conjunct references both `nation` aliases (`n1` on
+// the LHS, `n2` on the RHS of the join) and is rewritten to native match language because every
+// leaf is a `<field> = <literal>` comparison:
+//
+//     SELECT ...
+//     FROM lineitem, nation n1, nation n2
```
mattChiaravalloti:
```suggestion
//     FROM nation n1, nation n2
```
It looks like `lineitem` is not actually used in this test, so having it here in the comment is a bit confusing. Did you mean to include it in the test, or should we remove it from the comment?

mongosql/src/mir/optimizer/stage_movement/test.rs
```diff
@@ -1872,63 +1926,46 @@ test_move_stage!(
...
+            source: mir_collection("tpch", "n1"),
+            subquery: mir_collection("tpch", "n2"),
+            cache: SchemaCache::new(),
+        }))),
+        condition: nation_names_match_condition(),
         cache: SchemaCache::new(),
-    })),
-    expected_changed = true,
+    }))),
+    expected_changed = false,
```
mattChiaravalloti: [issue] This may lead to performance degradation in some cases. I see what the problem is and where you're coming from, but I don't think it is preferable to do this in all cases. We may want to flag this change as something we should reconsider after running tpch benchmarks.

The issue is that running the filter _after_ the `$lookup` guarantees 2 things:
1. the `$lookup` cannot ever utilize an index scan, even if server implements future changes/optimizations to enable such a scan
2. the filter will run after the desugared `$lookup`/`$unwind` meaning we cannot even leverage the "prefilter_unwinds" optimization to help alleviate the number of documents filtered

I cannot guarantee that either of those points will definitely cause performance degradation, but they seem less desirable than pushing the filter into the join and switching to expr language. Since we're not pushing the match-language filter to the beginning of any pipeline (the top-level pipeline or the `$lookup` pipeline) we aren't really getting any win out of keeping it written as match language. We might as well switch it back to expr language and push it to the beginning of the join pipeline.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -61,10 +62,48 @@ struct MatchLanguageRewriterVisitor {
...
+    /// ```text
+    /// MongoServerError[BadValue]: unknown top level operator: $a.b. If you
+    /// have a field name that starts with a '$' symbol, consider using
+    /// $getField or $setField.
+    /// ```
+    ///
+    /// When this returns `false`, the expression is left to be translated to
+    /// `$expr` language so the existing `$getField`-based translation handles
+    /// the field correctly.
+    fn is_match_addressable(field_path: &FieldPath) -> bool {
```
jpowell-mongo, line 94: Note: Consider moving this to the try_into for field patch to field access. Moving it down will make this cleaner and catch bugs elsewhere.

jpowell-mongo:
Running a patch to see how TPCH performance is with these changes:
https://spruce.corp.mongodb.com/version/6a5fa338b9115f0007e24cfd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

mongosql/src/mir/optimizer/stage_movement/mod.rs
```diff
@@ -593,6 +618,26 @@ impl StageMovementVisitor<'_> {
                     }
                 }
 
+                // Native match language ($match find syntax) cannot reference correlated `let`
```
jpowell-mongo, line 626: Note: After discussion the general case of correlated subqueries suggests that we may not be able to rewrite this to match-language without some performance degradation. This isn't really a question about using the same datasource and moreso about correlated subqueries.

We may want to consider a rewrite_to_expr_lang for quereis with correlated subqueries.

jpowell-mongo: Follow up:

Based on Evergreen patches, there is no performance impact (based on our tests) of special casing correlated subqueries by attempting to keep them above joins. Before merging, I'll raise an investigation ticket for correlated subqueries as those are the queries (TPCH-19) causing the tpch dataset to fail.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
jpowell-mongo: After investigation, there wasn't an easy way to merge this into the try_into on its own because the usage of try_into doesn't perfect match the is_match_addressable usage across the codebase. I'm going to leave this here for now.

mongosql/src/mir/optimizer/prefilter_unwinds/test.rs
mattChiaravalloti, line 1: [praise/chore] The three noop tests are great. Can we also add a test for when a `MatchFilter` does successfully result in a pre-filter for an `Unwind` source?

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+    ///
+    /// # Null / missing behavior
+    ///
+    /// SQL three-valued logic makes a comparison against null or missing
+    /// UNKNOWN, and a `WHERE` clause keeps only rows that evaluate to TRUE, so
+    /// null and missing fields must never satisfy a comparison.
+    ///
+    /// MQL comparisons are *type-bracketed* — they only match values in the same
+    /// BSON type class as their argument — which gives that behavior for free
+    /// for most operators:
```
mattChiaravalloti: [nit] It is implied by this comment but we may want to state explicitly that type-bracketing exclusively applies to match language, not expr language!

```
Enterprise temp> db.foo.insertMany([{_id: 0, a: null}, {_id: 1}, {_id: 2, a: 1}, {_id: 3, a: 5}, {_id: 4, a: 2}])
{
  acknowledged: true,
  insertedIds: { '0': 0, '1': 1, '2': 2, '3': 3, '4': 4 }
}
Enterprise temp> db.foo.aggregate([{$match: {$expr: {$lt: ["$a", 5]}}}])
[ { _id: 0, a: null }, { _id: 1 }, { _id: 2, a: 1 }, { _id: 4, a: 2 } ]
Enterprise temp> db.foo.aggregate([{$match: {a: {$lt: 5}}}])
[ { _id: 2, a: 1 }, { _id: 4, a: 2 } ]
```

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+    ///
+    /// - `$lt`, `$lte`, `$gt`, `$gte` never match null or missing, so they are
+    ///   emitted as bare comparisons.
```
mattChiaravalloti: [issue] Is this true? What about this example:

```
Enterprise temp> db.foo.insertMany([{_id: 0, a: null}, {_id: 1}, {_id: 2, a: 1}, {_id: 3, a: 5}, {_id: 4, a: 2}])
{
  acknowledged: true,
  insertedIds: { '0': 0, '1': 1, '2': 2, '3': 3, '4': 4 }
}
Enterprise temp> db.foo.aggregate([{$match: {a: {$lte: null}}}])
[ { _id: 0, a: null }, { _id: 1 } ]
```

`null` is equal to `null`, and `missing` is less than `null`, so these 2 documents pass this filter. In SQL, though, we should expect an empty result set since comparing to a null value results in null which is equivalent to false. I believe this applies to `$lte` and `$gte` since they involve and equality check that can cross type boundaries.

The comment below implies that we only consider literal null values problematic for `$eq` and` $ne`.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+    ///   `$eq: null` matches both null and missing, and `$ne: null` matches every
+    ///   present non-null value, whereas SQL says both are UNKNOWN for every row.
+    ///   Such comparisons return `None` and stay in `$expr`, where the existing
+    ///   desugaring handles them.
+    ///
+    /// The guard is applied to the *post-commute* operator, so `10 <> a` is
+    /// covered as well as `a <> 10`.
+    ///
+    /// Note that this pass is the only place these semantics can be adjusted.
+    /// `MatchNullFilteringOptimizer`, which inserts `{field: {$gt: null}}`
```
mattChiaravalloti, line 306: [issue] `MatchNullFilteringOptimizer` creates an expr-language `Filter` stage that does `{$gt: [field, null]}`, which is different than a match-language `MatchFilter` stage that does `{field: {$gt: null}}`.

Example:
```
Enterprise temp> db.foo.insertMany([{_id: 0, a: null}, {_id: 1}, {_id: 2, a: 1}, {_id: 3, a: 5}, {_id: 4, a: 2}])
{
  acknowledged: true,
  insertedIds: { '0': 0, '1': 1, '2': 2, '3': 3, '4': 4 }
}
Enterprise temp> db.foo.aggregate([{$match: {a: {$gt: null}}}])

Enterprise temp> db.foo.aggregate([{$match: {a: {$ne: null}}}])
[ { _id: 2, a: 1 }, { _id: 3, a: 5 }, { _id: 4, a: 2 } ]
Enterprise temp> db.foo.aggregate([{$match: {$expr: {$gt: ["$a", null]}}}])
[ { _id: 2, a: 1 }, { _id: 3, a: 5 }, { _id: 4, a: 2 } ]
Enterprise temp> db.foo.aggregate([{$match: {$expr: {$ne: ["$a", null]}}}])
[ { _id: 1 }, { _id: 2, a: 1 }, { _id: 3, a: 5 }, { _id: 4, a: 2 } ]
```

This plays into my previous comment about how we handle null literals in this optimization.

We should update this comment to make the behavior distinction between $expr and match (find) language in `$match` stages. The implementation is correct, since it rejects any comparison operator that has a null literal operand, but even the comment above that conditional says it is only relevant for `$eq` and `$ne`.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+
+        // Native match language cannot address a field whose name contains a
+        // `.`, starts with `$`, or is empty; such fields must stay in $expr.
+        if !Self::is_match_addressable(&field_path) {
+            return None;
+        }
+
+        // MQL cannot express a SQL comparison against NULL: `$eq: null` matches
+        // null and missing, `$ne: null` matches every present non-null value,
+        // while SQL evaluates both to UNKNOWN for every row. Leave it in $expr.
```
mattChiaravalloti: [issue/follow-up] This is the in-line comment I was talking about above! This conditional is correct and we should keep it applied to all comparison operators. What we should update is this comment to mention it is relevant for all other non-`$eq`/`$ne` because of type-bracketing.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+            return Some(comparison);
+        }
+
+        // `{field: {$ne: null}}` is the existence guard in match language: it is
+        // the complement of `{$eq: null}`, which matches null and missing, so it
+        // matches exactly the present, non-null values. Note `{$gt: null}` does
+        // *not* work here — match language comparisons are type-bracketed, so
+        // `$gt: null` brackets to the null type class and matches nothing. (The
+        // `$gt: null` idiom used elsewhere in the pipeline is an `$expr`
+        // construct, where comparisons use total BSON ordering instead.)
```
mattChiaravalloti, line 383: [thought] We may want to consider updating `MatchNullFiltering` (which was introduced before we had the `MatchFilter` stage) to switch to generating `MatchFilter` stages that us `{field: {$ne: null}}` for field existence instead of `{$expr: {$gt: [field, null]}}`! This work would be a follow-up ticket, though, since it's out of scope of this PR.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,195 @@ impl MatchLanguageRewriterVisitor {
...
+            arg: LiteralValue::Null,
+            cache: SchemaCache::new(),
+        });
+
+        Some(MatchQuery::Logical(MatchLanguageLogical {
+            op: MatchLanguageLogicalOp::And,
+            args: vec![null_guard, comparison],
+            cache: SchemaCache::new(),
+        }))
+    }
```
mattChiaravalloti, line 396: [praise] the implementation of this function is beautiful!

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
```diff
@@ -639,3 +642,357 @@ test_rewrite_to_match_language_no_op!(
...
+    input = filter_stage(Expression::ScalarFunction(ScalarFunctionApplication::new(
+        ScalarFunction::Neq,
+        vec![
+            Expression::Literal(LiteralValue::Integer(10)),
+            *mir_field_access("foo", "int", true),
+        ],
+    )))
+);
+
+// A NULL literal operand has no faithful match language equivalent — `$eq: null`
```
mattChiaravalloti, line 759: [chore] We'll want to add additional tests for the other comparison operators to show they also do not get rewritten when they have a literal null operand

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
```diff
@@ -639,3 +642,357 @@ test_rewrite_to_match_language_no_op!(
...
+    let fa = match *mir_field_access("foo", "$a.b", true) {
+        Expression::FieldAccess(fa) => fa,
+        _ => unreachable!(),
+    };
+    let fp: FieldPath = fa.try_into().unwrap();
+    assert_eq!(fp, mir_field_path("foo", vec!["$a.b"]));
+    // Assert on `fields` directly: FieldPath's PartialEq ignores is_nullable
+    // and we want the component boundary to be explicit here.
+    assert_eq!(fp.fields, vec!["$a.b".to_string()]);
+}
```
mattChiaravalloti: [chore] This test may be better placed in `mir/definitions.rs` since it is testing `impl TryFrom<FieldAccess> for FieldPath` which is implemented there. This test is not strictly related to this optimization so it is out of place in this test file.

mongosql/src/mir/optimizer/stage_movement/mod.rs
```diff
@@ -593,6 +618,29 @@ impl StageMovementVisitor<'_> {
...
+                // subquery (RHS) were moved into the subquery, its LHS field references would
+                // become correlated `$$`-variables, which native match cannot express (they would
+                // fail translation with Error::InvalidMatchLanguageInputRef). Keep such a
+                // MatchFilter above the join, where every datasource is a plain field reference. A
+                // regular $expr Filter is unaffected: correlated variables are valid in $expr.
+                if matches!(node, Stage::MqlIntrinsic(MqlStage::MatchFilter(_))) {
+                    let left_schema = n
+                        .source
+                        .schema(self.schema_state)
+                        .expect("LateralJoin LHS schema must be inferable");
```
mattChiaravalloti: [chore] As of #180, `left_schema` is already computed earlier in this match branch, so you'll have access to that variable! When you rebase, you'll see it.

mongosql/src/mir/optimizer/stage_movement/test.rs
mattChiaravalloti: Comment still stands.

mattChiaravalloti (changes requested):
This looks really great. Very exciting change! I have a few change requests, but they're mostly related to comments and test coverage -- very few, if any, have to do with functionality.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
jpowell-mongo: Yes, this is correct  for comparing non-null values. If you run this aggregation pipeline against the nullable_fields collection for the index usage tests:
```javascript
db.nullable_fields.aggregate([{ $match: { a: { $ne: 10 }}}])
```

It will return:

```javascript
[
  { _id: 4, a: null, b: 6000 },
  { _id: 7, b: 5203 },
  { _id: 9, a: null, b: 6704 },
  { _id: 3, a: 42, b: 3456 },
  { _id: 8, a: 50, b: 50 },
  { _id: 1, a: 100, b: 5050 },
  { _id: 2, a: 250, b: 4132 },
  { _id: 10, a: 263, b: 1218 },
  { _id: 5, a: 321, b: 10 },
  { _id: 6, a: 322 }
]
```

If you are running a direct comparison against null, then null values will be explicitly considered. This might be a semantic difference. The main idea, though, is that comparisons against non-null values with the $ne operator will also return fields where the value is null.

jpowell-mongo: I'll update the comment to be more precise about this bit of logic.

mongosql/src/mir/optimizer/prefilter_unwinds/test.rs
jpowell-mongo: Test added in:
f33e69883bb9fab3a816f9ee0ebb7241b72a07bd

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
jpowell-mongo: Updated the comment: 
https://github.com/mongodb/mongosql/pull/172/changes/9113ee235a7282a9fee00cdcea0eb216f507a788

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
jpowell-mongo: Additional tests for comparison operators added here:
https://github.com/mongodb/mongosql/pull/172/changes/26044cfb29803d7462f9d4604f7bc5bead089667

mongosql/src/mir/optimizer/stage_movement/mod.rs
jpowell-mongo: Updated this once I rebased. Resolving.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
jpowell-mongo: Updating this comment.

mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
```diff
@@ -122,6 +179,51 @@ impl Visitor for PrefilterUnwindsVisitor {
...
+            cache: SchemaCache::new(),
+        })))
+        .into(),
+        true,
+    )
+}
+
+// Pulls (op, literal) out of a lowered match condition. Handles a bare comparison
+// directly; for an `And`, returns the first conjunct that is a comparison against
+// a non-null literal. Comparisons against `null` are skipped because they carry no
```
jchemburkar: nit: is this correct? And clearly handles null, but comparisons don't, right?

mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
jchemburkar: non-blocking: ++ is there any chance we can macro this out or find a way to simplify? up to you

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
```diff
@@ -640,3 +643,401 @@ test_rewrite_to_match_language_no_op!(
...
+
+test_rewrite_to_match_language!(
+    rewrite_lte_over_nullable_field,
+    expected = match_filter_stage(match_comparison(MatchLanguageComparisonOp::Lte)),
+    expected_changed = true,
+    input = filter_stage(comparison(ScalarFunction::Lte, true))
+);
+
+// `$ne` inverts the type bracketing and would otherwise match null and missing,
+// so over a nullable field it is wrapped in a `{$gt: null}` existence guard.
```
jchemburkar, line 728: should this be `$ne: null`? that is my understanding from the discourse in mod.rs but let me know if I'm off target

jchemburkar (commented):
just had a couple small non-blockers on top of matts comments. I'll leave this as a comment for now, feel free to retag once all feedback is addressed!

mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
jpowell-mongo: I'd prefer to leave this match arm expanded rather than a macro. I know it's a bit long, but we only use this in two places here, so I think it's easier to debug just as a match arm than as a macro. That being said, long term we could certainly break this down into smaller functions.

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
jpowell-mongo: Good catch. Updating.

mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
jpowell-mongo: Could you clarify your point about "And clearly handles null?"

mongosql/src/mir/optimizer/rewrite_to_match_language/test.rs
jpowell-mongo: I actually just ended up deleting this test, because it was mostly for my own understanding when I introduced the is_match_addressable function.

mongosql/src/mir/optimizer/stage_movement/test.rs
jpowell-mongo: Updated and removed the reference to line item

mongosql/src/mir/optimizer/prefilter_unwinds/mod.rs
jpowell-mongo: [TODO] Update comment to make behavior obvious that we check for nulls in rewrite_to_match_language

jchemburkar (approved):

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
mattChiaravalloti: I still think we should update this to be clear it creates `{$expr: {$gt: [<field>, null]}}` since that has different semantics than `{field: {$gt: null}}`.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,201 @@ impl MatchLanguageRewriterVisitor {
...
+    /// Note that this pass is the only place these semantics can be adjusted.
+    /// `MatchNullFilteringOptimizer`, which inserts `{field: {$gt: null}}`
+    /// existence guards elsewhere in the pipeline, both runs *after* this pass
+    /// and only visits `Stage::Filter`; once a `Filter`'s condition has become a
+    /// `MatchQuery` there is no longer an `Expression`/`FieldAccess` tree to
+    /// inspect, and the translator and codegen are purely syntactic.
+    ///
+    /// In Match Language the `$gt: null` compares elements against null, and
+    /// will return an empty result set, whereas the same null guard, `$gt: null` will
+    /// filter out null elements.
```
mattChiaravalloti, line 314: [suggestion] This sentence is a bit confusing. Can we consider adjusting it to something like:
```
In Match Language, `{$gt: null}` uses type-bracketing to compare elements against null. Since null is a singleton data-type (i.e., a data type with only one value) this will return an empty result set since there are no values in the type-bracket "greater than null".
```
Or, perhaps even better, can we remove this sentence since it is irrelevant to the behavior of this function? We always leave comparisons with a literal `null` operand in `$expr` language so maybe there's no need to get into the weeds explaining this.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
```diff
@@ -186,24 +228,201 @@ impl MatchLanguageRewriterVisitor {
...
+
+        // Native match language cannot address a field whose name contains a
+        // `.`, starts with `$`, or is empty; such fields must stay in $expr.
+        if !Self::is_match_addressable(&field_path) {
+            return None;
+        }
+
+        // MQL cannot express a SQL comparison against NULL: `$eq: null` matches
+        // null and missing, `$ne: null` matches every present non-null value,
+        // this is also for other comparison operators due to type bracketing.
```
mattChiaravalloti, line 356: My previous PR comment still stands: this code comment does not accurately reflect the code and does not describe the full situation. We should update this to be consistent with the function-level comment (_any_ comparison involving a null literal is left in `$expr` language), or we should remove this comment.

mattChiaravalloti (approved):
lgtm 👍 

There are still a handful of comments with outdated or misleading information which we should address before merging, but I'll leave it up to you to decide if it's worth it and how to change or delete them. No need to retag me.

mongosql/src/mir/optimizer/rewrite_to_match_language/mod.rs
jpowell-mongo: I'll remove it for clarity.

PR #173, opened 2026-07-16
SQL-3292: Implement schema checking for HigherOrderFunction and Variable expressions

mattChiaravalloti:
This PR implements schema checking for `HigherOrderFunction` and `Variable` expressions, as described by the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.ntgguxk2r0tf), in addition to all necessary changes to support this schema checking.

This ended up being a large PR mostly because it introduces new `mir::schema::Error` functionality for wrapping errors with useful context related to higher order functions and variable usage (outlined in the design doc [here](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.we7d9ihtuqwt)). As a result of the new information attached to some error variants, I added a large amount of new unit test coverage.

Break down of what was done in this PR:
- Implemented schema checking for `mir::HigherOrderFunctionApplication`
- Implemented schema checking for `mir::Variable`
- Updated schema checking for `mir::ArrayExpr` to simplify its items schema
- Updated schema checking for `mir::SearchedCaseExpr` and `mir::SimpleCaseExpr` to avoid wrapping its output schema with `AnyOf` when the output is a singleton schema
- Added a new util function to `schema/definitions.rs`, `get_array_item_schema`, for getting the array items schema from a `Schema` object
- Updated `mir::schema::errors` to include context about whether a variable caused an error to occur (and, if so, which one)
- Updated `SchemaInferenceState` to track variables
- Added extensive new unit testing for new error work and updated and newly implemented schema checking work

When reviewing, I advise prioritizing your focus on these files:
- `mir/schema/test/expressions/higher_order_functions.rs` - this is where the unit tests for higher order function schema checking are implemented. You may consider starting here to get a grip on how schema checking for these functions is expected to work.
- `mir/schema/mod.rs` - schema checking is implemented here
- `mir/schema/errors.rs` - errors are implemented here
- `schema/definitions.rs` - `get_array_item_schema` is implemented here
- `schema/definitions.rs` - `get_array_item_schema` is tested here

Apart from that, the other tests in `mir/schema/test/expressions` are worth at least a cursory glance. Once you see one `_with_var_cause` test, you've seen them all. There are over 50 new non-higher-order-function tests added here that all, effectively, look the same. Feel free to review them as closely or loosely as you want.

Finally, the changes to any files in the `algebrizer`, `mir/optimizer`, and `mir/schema/test/stages` directories are purely mechanical. They were necessary as a result of changes to the `SchemaInferenceState` and/or `mir::schema::errors::Error` types. You can safely entirely skip those (saving yourselves on hundreds of lines of review), or, at least, not waste too much time reading them deeply as no behavioral changes have been made there.

Of course, feel free to review all 3,222 additions in detail if you desire! Just wanted to provide an outline to help ease the initial shock of seeing such a large PR. Sorry it ended up this way 😓


mongosql/src/mir/schema/test/expressions/aggregate.rs
mattChiaravalloti, line 1: This is an interesting thing to consider. The current implementation ignores the potential for `Variable` expressions to pop up in stage-level contexts, and that includes `AggregationExpr`s. However, it is technically going to be possible for a user to use a subquery expression as their higher-order-function function argument. For example, `MAP(array, SELECT a FROM foo WHERE this LIMIT 1)`, or, more relevant to this file, `MAP(array, SELECT ADD_TO_SET(this) FROM foo GROUP BY NULL)`.

The primary reason I decided to ignore the `AggregateExpr` case is because it is typically nonsensical. It would be a rare case to write a meaningful subquery expression as the function argument to a higher order function that uses an aggregation function with a variable binding as its argument. Since the variable binding would be a correlated value that stays static throughout the subquery execution, it would be quite meaningless to use it in an aggregation. A secondary reason is that I simply do not anticipate users attempting to use subquery expressions for function arguments when we initially release higher order function support. Beyond that, I'm not sure anyone will ever attempt to use subquery expressions in these contexts.

That last point bleeds into why the current implementation in this PR does not support `var_cause` in stage contexts, either. They would only be relevant for subquery expressions and those seem highly unlikely. At least for now...

I could be convinced otherwise, but even then I'd defer to a follow-up ticket that is outside the epic. This is already a large PR and it has already taken twice as long as anticipated.

mongosql/src/mir/schema/test/expressions/case.rs
```diff
@@ -31,9 +33,40 @@ mod searched {
         }),
     );
 
+    test_schema!(
```
mattChiaravalloti, line 36: If you are reading the files in github-order, this is likely the first `*_with_var_cause` test you'll see. Be sure to check out `mir/schema/errors.rs` to see what this `var_cause` field is! Anyway, notice that this test's `input` uses a `Variable` as the offending `Expression`, so the `expected` is an `Err` that includes that variable as the `var_cause` ("variable cause") of the error.

This test is basically identical to all other `*_with_var_cause` tests in this PR.

Similar to how we have `*_requires_X_num_of_arg(s)` and `*_requires_X_arg_schema` tests for all expression variants and subvariants, I think it is worth including all of these `*_with_var_cause` tests for all expression variants and subvariants since the unit tests should be written agnostic to the implementation. As in, the tests should not assume any expression variants or subvariants are implemented using shared code, so all cases should be covered by unit tests. I say this to preemptively dismiss suggestions that we not include these tests since they're repetitive.

mongosql/src/mir/schema/test/expressions/field_access.rs
```diff
@@ -6,12 +6,13 @@ use crate::{
 };
 
 test_schema!(
-    field_access_accessee_cannot_be_document,
+    accessee_must_be_document,
```
mattChiaravalloti, line 9: I did _some_ drive-by name-changing, but not extensively. For things like this, the module is called `field_access`, so we don't need to prefix every test name with `field_access` as well. I only edited (some of the) files that I otherwise was already adding new tests to.

mongosql/src/mir/schema/test/expressions/literal.rs
```diff
@@ -72,25 +72,20 @@ mod array {
     use super::*;
 
     test_schema!(
-        array_literal_empty,
-        expected = Ok(Schema::Array(Box::new(Schema::AnyOf(set![])))),
+        empty,
+        expected = Ok(Schema::Array(Box::new(Schema::Unsat))),
```
mattChiaravalloti, line 76: `AnyOf(set![])` is isomorphic to `Unsat`

mongosql/src/mir/schema/test/expressions/literal.rs
```diff
@@ -72,25 +72,20 @@ mod array {
...
         input = Expression::Array(vec![].into()),
     );
 
     test_schema!(
-        array_literal_null,
-        expected = Ok(Schema::Array(Box::new(Schema::AnyOf(set![
-            Schema::Atomic(Atomic::Null)
-        ])))),
+        with_single_atomic_item_results_in_array_of_atomic,
+        expected = Ok(Schema::Array(Box::new(Schema::Atomic(Atomic::Null)))),
```
mattChiaravalloti, line 82: In general, these tests are updated to reflect the updated implementation of `Array` expression schema checking, which now simplifies its items schema.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
mattChiaravalloti, line 1: This is half the PR in this one file, and all of the changes are just new `*_with_var_cause` tests.

mongosql/src/mir/schema/test/expressions/subquery.rs
```diff
@@ -423,4 +424,44 @@ mod subquery_comparison {
             is_nullable: true,
         }),
     );
+
+    test_schema!(
+        incomparable_argument_and_output_expr_with_var_cause,
```
mattChiaravalloti, line 429: Despite what I said earlier, this one was trivial to implement so I did add it.

mongosql/src/mir/schema/test/usererror.rs
mattChiaravalloti, line 1: I should probably add unit test coverage for the new wrapper variant and the new missing-variable variant.

mongosql/src/mir/schema/errors.rs
```diff
@@ -24,17 +24,54 @@ pub enum Error {
         name: &'static str,
         required: Box<Schema>,
         found: Box<Schema>,
+        var_cause: Option<String>,
```
mattChiaravalloti, line 27: So I considered quite a few different approaches to this problem and discussed them with @jchemburkar before putting up this PR. This is ultimately the solution we landed on, which I think worked out pretty well. Here's some more context:

Goal: Wrap schema-checking errors that occur in the context of a higher order function's function argument with contextual information about the surrounding higher order function and the root cause of the problem in the function argument.

Example:
    expression: MAP(['a', 'b'], this + 1)
    current error: "schema checking failed for Add: required Numeric, found Atomic(String)"
    desired error: "Invalid function argument for `MAP`. Invalid usage of variable `this`: ... <same as current error>"

Ideas considered:
1. Inspect errors after they happen and make educated guesses about the root cause
   - For example, when schema checking `MAP(['a'], this + 1)` we could attempt to schema check `this + 1` and see that the error is a `SchemaChecking` error that "found" a `String` and we could see that the variable `this` has the schema `String` and then conclude that `this` was the the cause.
   - Critical Flaw: as implemented, expressions like `MAP(['a', 'b'], [this, string_field + 1])` could produce misleading errors. There may not be a good way around this!
   - Flaw: This requires an additional visitor pass over the Expression after schema checking is done

2. Add contextual information to errors as they occur.

2a. Update the `schema` api for Expressions to return `Result<Schema, (Expression, Error)>`
     - This would allow us to avoid an additional visitor pass, but it's a pretty big change.

2b. Update the `Error` type to always include the expression that caused the error.
     - This would also require a lot of changes to the Error type and how it's used.

2c. Update some `Error` variants to include the expression that caused the error.
     - This would have been a bit too difficult even though it's more generic. `Expression` is not `PartialEq` or `Eq`, so that would make implementation a bit more annoying to do, plus it gets a bit too memory intensive for now

2d. The solution we went with: Update some `Error` variants to include the name of the variable that caused the error, or None if the error was not caused by a variable. This is the solution you see implemented.

mongosql/src/mir/schema/errors.rs
```diff
@@ -24,17 +24,54 @@ pub enum Error {
...
     AggregationArgumentMustBeSelfComparable(String, Box<Schema>),
-    InvalidComparison(&'static str, Box<Schema>, Box<Schema>),
+    InvalidComparison {
+        name: &'static str,
+        left: Box<Schema>,
+        right: Box<Schema>,
+        var_cause: Option<String>,
+    },
     CannotMergeObjects(Box<Schema>, Box<Schema>, Satisfaction),
     AccessMissingField(String, Option<Vec<String>>),
```
mattChiaravalloti, line 38: Arguably, these two errors may also arise as a result of a variable, but for now I'm willing to leave them as generic `InvalidFunctionArgument` causes. We can see if users actually ever encounter these and amend this as needed.

mongosql/src/mir/schema/errors.rs
```diff
@@ -211,6 +253,24 @@ impl UserError for Error {
             Error::UnwindIndexNameConflict(_) => None,
             Error::CollectionNotFound(_, _) => None,
             Error::InvalidBinaryDataType => None,
+            Error::HigherOrderFunctionWrapper { name, cause, error } => {
```
mattChiaravalloti, line 256: Very much open to suggestions on phrasing here for the user-error display messages!

mongosql/src/mir/schema/test/expressions/higher_order_function.rs
mattChiaravalloti, line 1: I'll also add some tests to demonstrate how shadowing works with nested higher order functions, and to include `*_with_var_cause` for the places where that's possible! (Such as `REDUCE([1, 2, 3], [], MAP(this, [this])`, which is invalid because the first `this` you see in `MAP` is scoped to the `REDUCE` and therefore has the schema `Integer`).

mongosql/src/mir/schema/mod.rs
```diff
@@ -131,6 +136,7 @@ pub struct SchemaInferenceState<'a> {
     pub scope_level: u16,
     pub env: SchemaEnvironment,
     pub catalog: &'a Catalog,
+    pub variables: BTreeMap<&'static str, Schema>,
```
mattChiaravalloti, line 139: I went with `&'static str` for now since this map is only ever constructed with the `const`s you just saw above. Eventually we may update the variable support in MongoSQL to use generic names, but for now we can get extremely minimal performance wins by using `&'static str` as opposed to `&'a str` or, even worse, `String`.

mongosql/src/mir/schema/mod.rs
```diff
@@ -993,7 +1034,21 @@ impl SubqueryComparison {
...
+        self.get_comparison_schema(
+            state,
+            &[
+                (&self.argument, argument_schema),
+                (
+                    // The SubqueryExpr of a SubqueryComparison will never be a Variable expr. The
+                    // real SubqueryExpr may be arbitrarily large. The `&Expression` that is passed
+                    // in the arg_schemas slice is only used to check if an offending Expression is
+                    // a Variable. Since that is never the case for the subquery argument of a
+                    // SubqueryComparison, we just pass in a dummy Literal here.
```
mattChiaravalloti, line 1046: An unfortunate consequence of the new api, but I think worth accepting since it fits all other use cases so well and so effectively solves the problem of higher-order-function error context. I figure this comment and mild awkwardness in this one use case is worth that trade-off.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1134,6 +1200,7 @@ impl Expression {
                         name: "Like",
                         required: STRING_OR_NULLISH.clone().into(),
                         found: expr_schema.into(),
+                        var_cause: l.expr.as_var_cause(),
```
mattChiaravalloti, line 1203: This is a great example of how this generally works: when you encounter a schema that causes a `SchemaChecking` error, you want to say the `Expression` from which that schema was computed is the culprit. For now, we only care about `Expression::Variable`s so we use the `as_var_cause` method you saw implemented earlier in this file.

mongosql/src/mir/schema/mod.rs
```diff
@@ -1414,20 +1486,21 @@ trait SqlFunction {
...
-        for (i, arg) in arg_schemas.iter().enumerate() {
-            if !state.check_satisfies(arg, &required_schemas[i]) {
+        for (i, (arg, arg_schema)) in arg_schemas.iter().enumerate() {
+            if !state.check_satisfies(arg_schema, &required_schemas[i]) {
                 return Err(Error::SchemaChecking {
                     name: self.as_str(),
                     required: required_schemas[i].clone().into(),
-                    found: arg.clone().into(),
+                    found: arg_schema.clone().into(),
+                    var_cause: arg.as_var_cause(),
```
mattChiaravalloti, line 1500: The whole stack of functions changed signature essentially just for this right here: at this point, we are returning the `SchemaChecking` error but we previously had fully lost the context of which `Expression` this problematic schema was associated with. For now, I still think it's worth it to do argument schema checking earlier (i.e., where it is currently implemented), so that's why I didn't overhaul this to pass around a `&[Expression]`. We're only passing a `&Expression` so it isn't expensive to move around. It just makes all the function signatures a little bigger and a little whackier.

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
     }
 }
 
+trait HigherOrderFunction: SqlFunction {
```
mattChiaravalloti, line 2177: Finally, here it starts, at the bottom of one of the last files you'll see in github-order. This is the start of the schema-checking implementation for `MAP`, `FILTER`, `REDUCE`, and variable expressions. I commented heavily in the code so I don't think I'll leave any additional comments here on the PR beyond this one.

mattChiaravalloti:
evergreen retry

mattChiaravalloti:
evergreen retry

mongosql/src/mir/schema/errors.rs
```diff
@@ -211,6 +253,24 @@ impl UserError for Error {
...
+                    HigherOrderFunctionErrorCause::InitialValueUsage => ("function argument", "Invalid usage of the variable `value` because of the initial value. Recall that usages of the `value` variable must be satisfied by the both the schema of the initial value and the schema of the result of the accumulator function."),
+                    HigherOrderFunctionErrorCause::ThisUsage => ("function argument", "Invalid usage of variable `this`. Recall that usages of the `this` variable must be satisfied by the schema of the elements of the array."),
+                    HigherOrderFunctionErrorCause::FunctionArgument => ("function argument", "Ensure the function argument is semantically valid. It must have the correct number of arguments and the arguments must have the correct type."),
+                };
+                let sub_error_message = error
+                    .user_message()
+                    .unwrap_or_else(|| error.technical_message());
+                Some(format!(
+                    "Invalid {cause_desc} for `{name}`: {cause_message}. Sub-Error Code {}: {}",
```
jchemburkar:
```suggestion
                    "Invalid {cause_desc} for `{name}`: {cause_message} Sub-Error Code {}: {}",
```
claude pointed out the cause messages already end in "."

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
...
+    /// That error is clear enough to a particularly well-informed user, however this error does
+    /// not generally make it clear that the problem is the function argument to the higher order
+    /// function MAP.
+    ///
+    /// This method wraps an error that occurs during schema checking of a higher order function's
+    /// function argument in additional context so users can understand better what the problem is.
+    /// Not only do the wrappers indicate that the error occurred within the context of a higher
+    /// order function, but they also provide additional information about the specific cause of
+    /// the error. For example, in the case of `MAP(['a'], this + 1)`, the error will be wrapped
+    /// in a `HigherOrderFunctionWrapper` with the cause `InvalidThisUsage`. See the
```
jchemburkar:
```suggestion
    /// in a `HigherOrderFunctionWrapper` with the cause `ThisUsage`. See the
```

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
...
+    /// `HigherOrderFunctionErrorCause` type for the full set of possible causes.
+    ///
+    /// Callers of this method should provide the specific `value_cause` that best describes the
+    /// cause of the error. The two options are `InvalidInitialValue` and `InvalidAccumulatedValue`.
```
jchemburkar:
```suggestion
    /// cause of the error. The two options are `InitialValueUsage` and `AccumulatedValueUsage`.
```

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
...
+    /// For example, if the error occurred because the `value` variable was schema-checked with the
+    /// initial value schema, the `value_cause` should be `InvalidInitialValue`. Otherwise it should
```
jchemburkar:
```suggestion
    /// initial value schema, the `value_cause` should be `InitialValueUsage`. Otherwise it should
```

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
...
+    /// be `InvalidAccumulatedValue`.
```
jchemburkar:
```suggestion
    /// be `AccumulatedValueUsage`.
```

Sorry probably could have did this as one but bad at github

mongosql/src/mir/schema/mod.rs
```diff
@@ -2060,6 +2174,298 @@ impl CastExpr {
...
+                var_cause: Some(ref var_cause),
+                ..
+            } => match var_cause.as_str() {
+                THIS_VARIABLE => HigherOrderFunctionErrorCause::ThisUsage,
+                VALUE_VARIABLE => {
+                    value_cause.unwrap_or(HigherOrderFunctionErrorCause::FunctionArgument)
+                }
+                _ => HigherOrderFunctionErrorCause::FunctionArgument,
+            },
+            // Otherwise, we use the generic cause `InvalidFunctionArgument`.
```
jchemburkar:
```suggestion
            // Otherwise, we use the generic cause `FunctionArgument`.
```

mongosql/src/mir/schema/test/expressions/higher_order_function.rs
```diff
@@ -0,0 +1,1062 @@
...
+                is_nullable: false,
+            })),
+            is_nullable: false,
+        })),
+    );
+
+    // MAP(['a'], [this, string_field + 1]) => error
+    test_schema!(
+        do_not_incorrectly_blame_variable_usage_when_variable_is_valid_but_function_is_still_valid,
+        expected_error_code = 1020,
```
EthanHardyMongo, line 138:
```suggestion
    // MAP(['a'], [this, string_field + 1]) => error
    test_schema!(
        do_not_incorrectly_blame_variable_usage_when_variable_is_valid_but_function_is_still_invalid,
        expected_error_code = 1020,
```
tiny naming nit

EthanHardyMongo (approved):
This looks really good! Great job on the test coverage and commenting. The comments made this a lot easier to review! I just have one test name fix.

mongosql/src/mir/schema/test/expressions/aggregate.rs
jchemburkar: agree with this conclusion, thanks for calling it out.

mongosql/src/mir/schema/test/usererror.rs
jchemburkar: agreed. I think coverage of the HigherOrderFunctionErrorCause variant of HigherOrderFunctionWrapper plus  NoSuchVariable would be good here, thanks for calling it out

mongosql/src/mir/schema/errors.rs
jchemburkar: This looks good, and I feel pretty happy with how it turned out looking at the pr in totality. Thanks for putting so much effort into this!

jchemburkar (approved):
LGTM! had some thoughts/requests (mostly nits). Awesome work, this came out great

PR #174, opened 2026-07-16
SQL-3311: Migrate fold_converts from desugarer to optimizer

EthanHardyMongo:
This PR moves convert/cast folding from the AIR's Desugarer to the MIR's optimizer to make the pipelines more efficient.


mongosql/src/air/desugarer/test.rs
```diff
@@ -117,391 +115,10 @@ mod sql_null_semantics {
...
-        };
-        static ref INT_DEC: bson::Decimal128 = bson::Decimal128::from_str("3").unwrap();
-        static ref DOUBLE_DEC: bson::Decimal128 =
-            bson::Decimal128::from_str(&format!("{}", std::f64::consts::PI)).unwrap();
-        static ref STRING_DEC: bson::Decimal128 = bson::Decimal128::from_str("3.14159265").unwrap();
-    }
-
-    test_desugar_manual!(
-        name = no_convert,
-        desugarer = FoldConvertsDesugarerPass,
```
EthanHardyMongo, line 146: All the tests were moved over to the `mir` except for this one because there are already `constant_folding` tests covering this behavior.

mongosql/src/air/desugarer/mod.rs
```diff
@@ -29,8 +27,6 @@ pub type Result<T> = std::result::Result<T, Error>;
 pub enum Error {
     #[error("pattern for $like must be literal")]
     InvalidLikePattern,
-    #[error("could not statically evaluate constant $convert to Type {0:?}, due to improper constant input value")]
-    InvalidConstantConvert(air::Type),
 }
```
EthanHardyMongo, line 30: Whenever the cast fails now, I return the `on_error` field. I'm curious why we didn't do that originally?

mattChiaravalloti (approved):
Lgtm 💯

mongosql/src/air/desugarer/mod.rs
mattChiaravalloti: Oh you know what, I think I know why now. Just because we fail to convert it doesn't necessarily mean the server will, I guess. It depends on the input. For something like converting `"a"` to an int, sure, we'll behave the same as server pretty trivially. But perhaps for some date formats or other conversion input/to combinations we may fail locally but server could succeed, so we should let them try instead of returning `on_error`? Maybe we'd need to audit fully exactly how the server converts everything to be sure about this.

EthanHardyMongo: I mean that previously we just returned the `InvalidConstantConvert(air::Type)` error and crashed the program even though the `on_error` field was always there.

mattChiaravalloti: Oh very interesting! Yes, we don't need to crash the whole program. ~I do still wonder now, though, if we should go with the `on_error` value or if we should instead simply not optimize and let the server attempt the conversion.~

mattChiaravalloti: Actually, never mind, you're right. Let's stick with what you did here. The only cases where we're return `Error` and then using `on_error` are certainly equivalent to what server would do since we're using driver code for the conversions. Great catch, great update.

jpowell-mongo (commented):
Question: Do we need to explicitly call the code for the fold_convert.rs? On main, we call the fold_converts_desugarer_pass in mod.rs for the desugarer. Do we need the same in the mod.rs for the Optimizer now? Or was this meant to just move the code?

Otherwise I can give this a stamp.

jpowell-mongo (approved):

PR #175, opened 2026-07-22
SQL-3312: Expand support for CAST constant folding

mattChiaravalloti:
This PR updates the `constant_folding` optimization to support more constant folding of `CAST` expressions with literal input arguments.

Apologies that the PR turned out so big -- it is primarily testing that led to so many line changes. The tests are (loosely) adapted from MongoDB docs ([link](https://www.mongodb.com/docs/manual/reference/operator/aggregation/convert/)) so you have a more easily readable source of truth available when reviewing.

The main motivation for this ticket came from the linked [HELP ticket](https://jira.mongodb.org/browse/HELP-95941) which revealed users of the Tableau driver who attempt to use `Date` literals ended up with expressions like `CAST('2026-07-22' AS TIMESTAMP)` in their queries. Until SQL-3311 was implemented (#174), most `CAST` expressions with literal inputs were not constant folded until desugaring time which definitively prevented index utilization when such expressions appeared in filters. To enable this type of expression to utilize an index when used in a filter, we need to constant fold it to a literal value _before_ translating from `mir` to `air` since that is when we determine semantics (SQL vs MQL). (By default, a `CAST` expression without `ON NULL` and `ON ERROR` specified is considered nullable since those two branches return `null` for their respective cases. Positive nullability of what is effectively a non-null literal prevents the happy path for translation to MQL-semantic comparison operations, which are typically index-eligible.)

As a follow-up to SQL-3311 (#174), this PR expands constant folding to support converting date-formatted strings that use our Tableau connector's prescribed format ([link](https://github.com/mongodb/mongo-tableau-connector/blob/91baa895c6355bff5a6220925b9c1755245226e3/connector/dialect.tdd#L1271)) into datetime literals. Additionally, this PR expands constant folding for most other valid target types as well. The full list of new supported constant folding `CAST` conversions is:
- `array`, `bindata`, `boolean`, `date`, `decimal`, `document`, `double`, `int`, `javascript`, `javascript_with_scope`, `long`, `max_key`, `min_key`, `object_id`, `regex`, `string`, and `timestamp` literals to boolean
- `boolean`, `decimal`, `double`, `int`, `long`, and `string` literals to int
- `boolean`, `decimal`, `double`, `int`, `long`, `string`, and `date` literals to decimal128
- `boolean`, `decimal`, `double`, `int`, `long`, `string`, and `date` literals to double
- `boolean`, `decimal`, `double`, `int`, `long`, `string`, and `date` literals to long
- `string` and `object_id` literals to object_id
- `decimal`, `double`, `long`, `string`, `date`, and `object_id` literals to date
- `boolean`, `date`, `decimal`, `double`, `int`, `long`, and `object_id` literals to string

### DIFF
--- a/mongosql/src/mir/optimizer/constant_folding/lib.rs
+++ b/mongosql/src/mir/optimizer/constant_folding/lib.rs
@@ -1201,8 +1201,116 @@ impl ConstantFoldExprVisitor<'_> {
         }
     }
 
-    fn convert_to_datetime(_l: &LiteralValue) -> Option<Result<Expression, ()>> {
-        todo!()
+    fn convert_to_datetime(l: &LiteralValue) -> Option<Result<Expression, ()>> {
+        match l {
+            // Null literal values are handled directly by the `fold_cast_expr` method. Here, we
+            // return None to indicate folding did not happen, but still could in `fold_cast_expr`.
+            LiteralValue::Null => None,
+
+            // Datetimes are trivially converted to themselves.
+            LiteralValue::DateTime(v) => Some(Ok(Expression::Literal(LiteralValue::DateTime(*v)))),
+
+            // Numbers may be converted to datetime by converting to long and then to datetime.
+            // Numbers that are out of range of i64 result in an error.
+            LiteralValue::Long(v) => Some(Ok(Expression::Literal(LiteralValue::DateTime(
+                bson::DateTime::from_millis(*v),
+            )))),
+            LiteralValue::Double(v) => {
+                // MongoDB truncates when converting a double to a date.
+                let truncated = v.trunc();
+                if truncated.is_finite()
+                    && truncated >= i64::MIN as f64
+                    && truncated <= i64::MAX as f64
+                {
+                    Some(Ok(Expression::Literal(LiteralValue::DateTime(
+                        bson::DateTime::from_millis(truncated as i64),
+                    ))))
+                } else {
+                    Some(Err(()))
+                }
+            }
+            LiteralValue::Decimal128(v) => {
+                // MongoDB truncates when converting a decimal to a date.
+                // Since we cannot operate on Decimal128 values directly, we convert to string and
+                // truncate by taking only the part before the decimal point.
+                let s = v.to_string();
+                let truncated = s.split('.').next().unwrap_or(&s);
+                Some(
+                    i64::from_str(&truncated)
+                        .map(|i| {
+                            Expression::Literal(LiteralValue::DateTime(
+                                bson::DateTime::from_millis(i),
+                            ))
+                        })
+                        .map_err(|_| ()),
+                )
+            }
+
+            // ObjectIds may be converted to datetime by getting the timestamp from the ObjectId.
+            LiteralValue::ObjectId(v) => Some(Ok(Expression::Literal(LiteralValue::DateTime(
+                v.timestamp(),
+            )))),
+
+            // Strings can be converted into the dates that correspond to the date string.
+            // Valid formats that MongoDB supports are:
+            //   - "2018-03-03"
+            //   - "2018-03-03T12:00:00Z"
+            //   - "2018-03-03T12:00:00+0500"
+            LiteralValue::String(v) => {
+                // Check if string ends with timezone offset: +HHMM or -HHMM
+                let has_timezone_offset = v.len() >= 5 && {
+                    let last_5 = &v[v.len() - 5..];
+                    let bytes = last_5.as_bytes();
+                    (bytes[0] == b'+' || bytes[0] == b'-')
+                        && bytes[1..].iter().all(|b| b.is_ascii_digit())
+                };
+
+                // If it ends with 'Z' or '+/-HHMM', we can attempt to parse it directly. That is,
+                // if it matches the format YYYY-MM-DD(T| )HH:mm:SS((.ms)?Z|(+|-)HHMM)?, where `?`
+                // denotes optional.
+                let chrono_dt: Option<chrono::DateTime<Utc>> =
+                    if v.ends_with('Z') || has_timezone_offset {
+                        v.parse().ok()
+                    } else if v.len() == 10 {
+                        // If it does not end with 'Z' or '+/-HHMM' and is length 10, we assume it
+                        // is in YYYY-MM-DD format. We append the T00:00:00Z time to the string so
+                        // that we can attempt to parse it into a chrono::DateTime<Utc>.
+                        let mut s = v.to_string();
+                        s.push_str("T00:00:00Z");
+                        s.parse().ok()
+                    } else {
+                        // If it does not end with 'Z' or '+/-HHMM' and is not length 10, we assume
+                        // it is in YYYY-MM-DD(T| )HH:mm:SS(.ms)? format. We append the 'Z' to the
+                        // string so we can attempt to parse it into a chrono::DateTime<Utc>.
+                        let mut s = v.to_string();
+                        s.push('Z');
+                        s.parse().ok()
+                    };
+
+                // We cannot guarantee our format here supports everything that MongoDB supports,
+                // so we instead fall back to the original cast expression when our attempt to
+                // create a Date fails, and allow MongoDB to succeed or fail as it will.
+                chrono_dt.map(|chrono_dt| {
+                    Ok(Expression::Literal(LiteralValue::DateTime(
+                        chrono_dt.into(),
+                    )))
+                })
+            }
+
+            // These types are not supported for conversion to int.
+            LiteralValue::Boolean(_)
+            | LiteralValue::Binary(_)
+            | LiteralValue::Integer(_)
+            | LiteralValue::JavaScriptCode(_)
+            | LiteralValue::JavaScriptCodeWithScope(_)
+            | LiteralValue::MaxKey
+            | LiteralValue::MinKey
+            | LiteralValue::RegularExpression(_)
+            | LiteralValue::Timestamp(_)
+            | LiteralValue::DbPointer(_)
+            | LiteralValue::Symbol(_)
+            | LiteralValue::Undefined => None,
+        }
     }
 
     fn convert_to_object_id(_l: &LiteralValue) -> Option<Result<Expression, ()>> {
--- a/mongosql/src/mir/optimizer/constant_folding/lib.rs
+++ b/mongosql/src/mir/optimizer/constant_folding/lib.rs
@@ -1112,7 +1112,7 @@ impl ConstantFoldExprVisitor<'_> {
                 v.timestamp_millis() as f64,
             )))),
 
-            // These types are not supported for conversion to int.
+            // These types are not supported for conversion to double.
             LiteralValue::Binary(_)
             | LiteralValue::JavaScriptCode(_)
             | LiteralValue::JavaScriptCodeWithScope(_)
@@ -1186,7 +1186,7 @@ impl ConstantFoldExprVisitor<'_> {
                 v.timestamp_millis(),
             )))),
 
-            // These types are not supported for conversion to int.
+            // These types are not supported for conversion to long.
             LiteralValue::Binary(_)
             | LiteralValue::JavaScriptCode(_)
             | LiteralValue::JavaScriptCodeWithScope(_)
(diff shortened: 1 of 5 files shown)


tests/index_usage_tests/filter.yml
```diff
@@ -259,3 +259,8 @@ tests:
     current_db: index_usage_filter
     query: "SELECT * FROM nullable_fields WHERE b IN (SELECT b FROM non_nullable_fields WHERE a < 300)"
     expected_utilization: COLL_SCAN
+
+  - description: CAST literal date using Tableau Date format to datetime uses index
```
mattChiaravalloti, line 263: I debated back and forth with myself whether we should introduce extensive index utilization testing for all sorts of `CAST` constant folding. At the end of the day, to avoid blowing up the test count, I went with just one test that covers the primary case from the HELP ticket. Maybe I'll file a ticket to add more tests (possibly via generation like how we do `type_constraint_tests`) but I'm not even sure that's necessary. What do you guys think?

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+    /// Cast's on_error.
+    ///
+    /// Some(Ok(_)) means we successfully converted a literal.
     fn convert_literal(l: &LiteralValue, to: Type) -> Option<Result<Expression, ()>> {
-        match l {
-            LiteralValue::String(s) => Self::convert_string_literal(s, to),
-            LiteralValue::Integer(i) => Self::convert_numerical_literal(*i, to),
-            LiteralValue::Long(l) => Self::convert_numerical_literal(*l, to),
-            LiteralValue::Double(d) => Self::convert_numerical_literal(*d, to),
+        match to {
```
mattChiaravalloti, line 837: I changed this to match on the _target_ type instead of the source type. To me, this is much more readable since each helper function encodes how to convert every eligible type _to_ a given type. This is also more extensible. In the future, if we support more types being converted to, say, `String`, we won't need to update multiple helper functions (one per new type supported); we'll just have to update one.

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+            Type::Boolean => Some(Ok(Expression::Literal(LiteralValue::Boolean(true)))),
+            _ => None,
+        }
+    }
+
+    fn convert_to_boolean(l: &LiteralValue) -> Option<Result<Expression, ()>> {
+        match l {
+            // Null literal values are handled directly by the `fold_cast_expr` method. Here, we
+            // return None to indicate folding did not happen, but still could in `fold_cast_expr`.
+            LiteralValue::Null => None,
```
mattChiaravalloti, line 894: This is a pattern you'll see in each `convert_to_*` function. I didn't want to plumb the `on_null` value through since we have pretty elegant handling of that already in one location in `fold_cast_expr`. I think the duplicated comment and `LiteralValue::Null => None` line is a worthy tradeoff for the avoided plumbing and inevitable cloning that would follow.

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+            // Doubles are trivially converted to themselves.
+            LiteralValue::Double(v) => Some(Ok(Expression::Literal(LiteralValue::Double(*v)))),
+
+            // Other numeric types may be converted to double if they are in range.
+            LiteralValue::Integer(v) => {
+                Some(Ok(Expression::Literal(LiteralValue::Double(*v as f64))))
+            }
+            LiteralValue::Long(v) => {
+                // MongoDB's conversion of long to double can result in loss of precision, so we do
+                // the same by using Rust's `as f64` operation which also loses precision.
```
mattChiaravalloti, line 1073: Experimentally, it loses precision in the same way MongoDB does. Of course, I cannot exhaustively test this for all i64 values. If anyone has any compelling examples for the min and max values to use as bounds for "i64 as f64" conversions, let me know and we can impose those here. We can then defer all values outside those bounds to MongoDB.

I'm not yet convinced that's necessary, but definitely curious to hear your thoughts on this.

mongosql/src/mir/optimizer/constant_folding/test.rs
```diff
@@ -1,3523 +1,6268 @@
-mod constant_folding {
```
mattChiaravalloti, line 1: ❗ ❗ **worth reading** ❗ ❗ 

Apologies about this. This module, `mir::optimizer::constant_folding::test` contained a submodule called `constant_folding` which contained the tests. So the full path to the tests was `mir::optimizer::constant_folding::test::constant_folding`. I removed that nested module since it is completely unnecessary. As a result, every line in this file was shifted to the left by one tab's worth of whitespace.

To make reviewing the new tests easier, I created a new submodule at the _end_ of the file called `cast` which contains all of the updated/new tests. You can truly safely ignore 3100 lines of this file since they are just whitespace-edited lines. **The real stuff worth reviewing (new and updated unit test coverage for `CastExpr` constant folding) is located at the end of the file.**

mongosql/src/mir/optimizer/constant_folding/test.rs
```diff
@@ -1,3523 +1,6268 @@
...
-            cache: SchemaCache::new(),
-        }),
-    );
-    test_constant_fold!(
-        mul_one_ref_is_ref,
-        expected = Stage::Array(ArraySource {
-            alias: "".into(),
-            array: vec![Expression::Reference(("a", 0u16).into())],
-            cache: SchemaCache::new(),
-        }),
```
mattChiaravalloti, line 568: sigh, I'm not sure why git determined this huge deleted chunk needed to appear in the middle of the new test coverage 😢 keep scrolling to see the end of this `cast::to_document::from_document` test. After that, you'll see all the other new tests...

mongosql/src/mir/optimizer/constant_folding/test.rs
```diff
@@ -1,3523 +1,6268 @@
...
-    );
-    test_constant_fold_no_op!(
-        filter_non_literal,
-        Stage::Filter(Filter {
-            source: Box::new(test_source()),
-            condition: Expression::Reference(("a", 0u16).into()),
-            cache: SchemaCache::new(),
-        })
-    );
+                )],
```
mattChiaravalloti: This is where the end of that `cast::to_document::from_document` test starts! Then there's one more `to_document` test after this. Then, finally, its smooth sailing to the end of the file with just new unit test coverage 👍

mattChiaravalloti:
~hold off on reviewing this morning -- I'm attempting to shrink the number of lines dedicated to new tests by introducing a new macro~

Done! Ready for review

tests/index_usage_tests/filter.yml
jpowell-mongo: This makes sense. I think BSON_DATE is one of our most complex types, so I have confidence in this from illustration. Let's leave this here.

jpowell-mongo (approved):

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+
+    fn convert_to_int(l: &LiteralValue) -> Option<Result<Expression, ()>> {
+        match l {
+            // Null literal values are handled directly by the `fold_cast_expr` method. Here, we
+            // return None to indicate folding did not happen, but still could in `fold_cast_expr`.
+            LiteralValue::Null => None,
+
+            // False converts to 0, true converts to 1.
+            LiteralValue::Boolean(false) => Some(Ok(Expression::Literal(LiteralValue::Integer(0)))),
+            LiteralValue::Boolean(true) => Some(Ok(Expression::Literal(LiteralValue::Integer(1)))),
```
nicholascioli: Suggestion: integer types implement `From<bool>`. See [1]

```suggestion
            LiteralValue::Boolean(b) => Some(Ok(Expression::Literal(LiteralValue::Integer(b.into())))),
```

[1]: https://doc.rust-lang.org/std/primitive.u8.html#impl-From%3Cbool%3E-for-u8

nicholascioli: This can also be done in the other spots where `bool`s are converted.

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+                        f64::NEG_INFINITY,
+                    )))),
+                    v => match f64::from_str(v) {
+                        Ok(f) if f.is_finite() => {
+                            Some(Ok(Expression::Literal(LiteralValue::Double(f))))
+                        }
+                        _ => Some(Err(())),
+                    },
+                }
+            }
```
nicholascioli, line 1121: This might not work as expected. I tried seeing how this conversion works in practice and it seems that if it is out of range, it does not fail. Check the playground link at [1] for more info. Also, it seems that Rust can already parse inf / -inf / nan as a string without issue, so manually checking it might not be necessary.

[1]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ddbb501829c7c135bd9cc5aba89842a3

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -819,38 +819,474 @@ impl ConstantFoldExprVisitor<'_> {
...
+            // ObjectIds may be converted to datetime by getting the timestamp from the ObjectId.
+            LiteralValue::ObjectId(v) => Some(Ok(Expression::Literal(LiteralValue::DateTime(
+                v.timestamp(),
+            )))),
+
+            // Strings can be converted into the dates that correspond to the date string.
+            // Valid formats that MongoDB supports are:
+            //   - "2018-03-03"
+            //   - "2018-03-03T12:00:00Z"
+            //   - "2018-03-03T12:00:00+0500"
```
nicholascioli, line 1273: Suggestion: it might be better to just try parsing in these formats using chrono's `parse_from_str` at [1]

[1]: https://docs.rs/chrono/0.4.45/chrono/struct.DateTime.html#method.parse_from_str

nicholascioli (approved):
Looks great! I think there might be an issue with the float parsing, but I'll leave it up to you.

mongosql/src/mir/optimizer/constant_folding/lib.rs
mattChiaravalloti: Done!

mongosql/src/mir/optimizer/constant_folding/lib.rs
mattChiaravalloti: Discussed offline: ultimately, it is a deliberate decision on Rust's part to parse strings into the "closest" f64 value. For some out-of-range values, that could be `f64::{MAX/MIN}`, and for other (larger/smaller) out-of-range values, that could be `f64::{INFINITY/NEG_INFINITY}`. Given that, I decided to stick with the manual `inf`/`-inf` checks and the `f.is_finite()` check, since this still guards against _very_ out of range values. For only "slightly" out of range values, I added guards that assert `f64` equality to `MAX/MIN` _and_ string equality to their corresponding string representations. It is certainly a little odd to look at but is the correct way to handle this so our constant folding matches mongodb's behavior. Fortunately, I don't anticipate many analytics users attempting to cast literal string values of "slightly" out of range f64 values into doubles, so if this is in any way "inefficient" it won't get executed very often 🤷‍♀️ 

thank you for calling this out and providing an example test case. I added that plus a corresponding min test in addition to the existing max/min tests.

mongosql/src/mir/optimizer/constant_folding/lib.rs
mattChiaravalloti: I attempted this but cannot get it to work in an acceptable way. I am going to defer this to a follow-up tech-debt ticket.

mattChiaravalloti: https://jira.mongodb.org/browse/SQL-3383

PR #177, opened 2026-07-24
SQL-3350: Bump mongodb rust driver version to 3.8

mattChiaravalloti:
This PR bumps the mongodb rust driver dependency from 3.7 to 3.8. The evergreen waterfall's sbom task is failing because the version of `serde_with` currently in `Cargo.lock` has a vulnerability. `serde_with` is a transitive dependency we get from `mongodb`, and it is updated to the version with the vuln fix in mongodb version 3.8.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -320,11 +320,11 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.10.4"
+version = "0.12.1"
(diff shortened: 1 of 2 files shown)


mattChiaravalloti:
I scheduled the 2 failing `sbom` tasks in the evergreen patch, so if it goes green we'll know this problem is resolved 👍

bucaojit (approved):
LGTM

PR #179, opened 2026-07-27
SQL-3293: Refactor algebrizer to pass expression context via self instead of algebrize_* method api

mattChiaravalloti:
This PR refactors the `algebrizer/definitions.rs` file so that expression-related context is passed via a field on the `Algebrizer` struct (`expression_context: ExpressionContext`) instead of via the `algebrize_*` method api (e.g., `algebrize_expression(&self, ast_node: ast::Expression, in_implicit_type_conversion_context: bool)`). This PR is a preliminary PR for the actual work of SQL-3293 which will introduce algebrization for the new higher order functions.

The reason this change is useful is because for higher order function algebrization I need to introduce additional expression-related context to the `Algebrizer`. The new context is whether or not we are in a higher order function's function argument, in which unqualified `ast::Expression::Identifier`s may be considered variables instead of fields. Instead of blowing up the `algebrize_*` method signatures with additional boolean arguments, I decided to create an `ExpressionContext` struct which will store expression-level context in the `Algebrizer` itself. I have work done locally already that extends `ExpressionContext` to support this new flag, but I omitted that from this PR since it is purely related to higher order function algebrization (which will be put up as a follow-up to this).

In the interest of keeping PRs more focused and keeping reviews a bit smaller, I am putting this up first and separately. This PR will block the follow-up until it is merged.

### DIFF
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -17,7 +17,7 @@ use crate::{
     util::unique_linked_hash_map::UniqueLinkedHashMap,
     SchemaCheckingMode,
 };
-use std::{cell::RefCell, collections::BTreeSet};
+use std::{cell::RefCell, collections::BTreeSet, rc::Rc};
 
 type Result<T> = std::result::Result<T, Error>;
 
@@ -231,17 +231,21 @@ impl std::fmt::Display for ClauseType {
 #[derive(Debug, Clone)]
 pub struct Algebrizer<'a> {
     current_db: &'a str,
-    pub schema_env: SchemaEnvironment,
+    pub schema_env: Rc<SchemaEnvironment>,
     catalog: &'a Catalog,
     scope_level: u16,
     schema_checking_mode: SchemaCheckingMode,
     allow_order_by_missing_columns: bool,
     clause_type: RefCell<ClauseType>,
+    expression_context: ExpressionContext,
 }
 
+/// ExpressionContext contains information about the context in which an expression is being
+/// algebrized. In certain contexts, certain ast::Expressions are resolved differently. See fields
+/// for more information.
 #[derive(Debug, Clone, Copy, Default)]
 pub(crate) struct ExpressionContext {
-    // Indicates if the expression is being algebrized is in an implicit type conversion context.
+    // Indicates if the expression is being algebrized in an implicit type conversion context.
     // An expression is "in an implicit type conversion context" whenever it appears in a place
     // where a string is unexpected. For example, the operands of a numeric operator like + are
     // unexpected to be strings. MongoSQL supports implicit type conversion from extended-JSON
@@ -250,9 +254,9 @@ pub(crate) struct ExpressionContext {
 }
 
 impl ExpressionContext {
-    pub(crate) fn with_implicit_type_conversion_ctx(self) -> Self {
+    pub(crate) fn with_implicit_type_conversion_ctx(self, value: bool) -> Self {
         Self {
-            in_implicit_type_conversion_ctx: true,
+            in_implicit_type_conversion_ctx: value,
             ..self
         }
     }
@@ -275,6 +279,7 @@ impl<'a> Algebrizer<'a> {
             schema_checking_mode,
             allow_order_by_missing_columns,
             clause_type,
+            ExpressionContext::default(),
         )
     }
 
@@ -286,21 +291,23 @@ impl<'a> Algebrizer<'a> {
         schema_checking_mode: SchemaCheckingMode,
         allow_order_by_missing_columns: bool,
         clause_type: ClauseType,
+        expression_context: ExpressionContext,
     ) -> Self {
         Self {
             current_db,
-            schema_env,
+            schema_env: Rc::new(schema_env),
             catalog,
             scope_level,
             schema_checking_mode,
             allow_order_by_missing_columns,
             clause_type: RefCell::new(clause_type),
+            expression_context,
         }
     }
 
     pub fn schema_inference_state(&self) -> SchemaInferenceState<'_> {
         SchemaInferenceState {
-            env: self.schema_env.clone(),
+            env: (*self.schema_env).clone(),
             catalog: self.catalog,
             scope_level: self.scope_level,
             schema_checking_mode: self.schema_checking_mode,
@@ -315,12 +322,33 @@ impl<'a> Algebrizer<'a> {
             catalog: self.catalog,
             scope_level: self.scope_level + 1,
             schema_checking_mode: self.schema_checking_mode,
+            allow_order_by_missing_columns: self.allow_order_by_missing_columns,
             // subquery should use a copy of the current ClauseType, not share a RefCell with the
             // parent. It probably does not matter, but we do not want subqueries to modify parent
             // state.
+            clause_type: RefCell::new(*self.clause_type.borrow()),
+            expression_context: self.expression_context,
+        }
+    }
+
+    pub(crate) fn with_expression_context(&self, context: ExpressionContext) -> Self {
+        Self {
+            current_db: self.current_db,
+            schema_env: self.schema_env.clone(),
+            catalog: self.catalog,
+            scope_level: self.scope_level,
+            schema_checking_mode: self.schema_checking_mode,
             allow_order_by_missing_columns: self.allow_order_by_missing_columns,
             clause_type: RefCell::new(*self.clause_type.borrow()),
+            expression_context: context,
         }
+    }
+
+    fn with_implicit_type_conversion_ctx(&self, value: bool) -> Self {
+        self.with_expression_context(
+            self.expression_context
+                .with_implicit_type_conversion_ctx(value),
+        )
     }
 
     fn args_are_nullable(args: &[mir::Expression]) -> bool {
@@ -334,8 +362,8 @@ impl<'a> Algebrizer<'a> {
         if matches!(func, mir::ScalarFunction::In | mir::ScalarFunction::NotIn) {
             Self::determine_in_expression_nullability(args)
         } else {
-            // some functions can always be nullable regardless of argument nullablity,
-            // we check those first. If this function is not one of those, we set nullablity
+            // some functions can always be nullable regardless of argument nullability,
+            // we check those first. If this function is not one of those, we set nullability
             // based off the arguments.
             func.is_always_nullable() || Self::args_are_nullable(args)
         }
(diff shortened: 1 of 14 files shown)


mongosql/src/algebrizer/test/expressions/between.rs
mattChiaravalloti, line 1: Turns out we really don't have adequate unit testing for ITC! I will file a tech-debt ticket to come back and add this testing.

In general in these test files, the following updates were made:
- `in_implicit_type_conversion_context = false` ==> `expression_context = ExpressionContext::default()`
- `in_implicit_type_conversion_context = true` ==> `expression_context = ExpressionContext::default().with_implicit_type_conversion_ctx(true)`

Those are the only changes in these test files! No true content or coverage changes have been made.

mongosql/src/algebrizer/test/mod.rs
mattChiaravalloti, line 1: This file contains the incidental updates to the macros and method calls to match the new api.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -231,12 +231,35 @@ impl std::fmt::Display for ClauseType {
 #[derive(Debug, Clone)]
 pub struct Algebrizer<'a> {
     current_db: &'a str,
-    pub schema_env: SchemaEnvironment,
+    pub schema_env: Rc<SchemaEnvironment>,
```
mattChiaravalloti, line 234: Since the `Algebrizer` generally relies on being immutable and the new `expression_context` setting requires cloning, I updated this to be an `Rc<SchemaEnvironment>` to keep cloning cheap. There isn't really any reason to have it be a pure `SchemaEnvironment` since we never actually write updates to it permanently. Ultimately, this change makes the introduction of the `with_expression_context`/`with_implicit_type_conversion_ctx` methods on `Algebrizer` _much_ more palatable.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -315,8 +362,8 @@ impl<'a> Algebrizer<'a> {
         if matches!(func, mir::ScalarFunction::In | mir::ScalarFunction::NotIn) {
             Self::determine_in_expression_nullability(args)
         } else {
-            // some functions can always be nullable regardless of argument nullablity,
-            // we check those first. If this function is not one of those, we set nullablity
+            // some functions can always be nullable regardless of argument nullability,
```
mattChiaravalloti, line 365: drive-by typo fixes!

mongosql/src/algebrizer/definitions.rs
```diff
@@ -346,11 +393,16 @@ impl<'a> Algebrizer<'a> {
         }
     }
 
-    fn with_merged_mappings(mut self, mappings: SchemaEnvironment) -> Result<Self> {
-        self.schema_env
+    fn with_merged_mappings(&self, mappings: SchemaEnvironment) -> Result<Self> {
+        let mut merged_env = (*self.schema_env).clone();
```
mattChiaravalloti, line 397: This is now the one place we do true deep cloning of the `SchemaEnvironment`, but overall this PR represents a cloning win.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -471,14 +523,15 @@ impl<'a> Algebrizer<'a> {
...
         // The SchemaEnvironment from the source is merged into the SchemaEnvironment from the
         // current Algebrizer, correctly giving us the correlated bindings with the bindings
         // available from the current query level.
-        #[allow(unused_variables)]
-        let expression_algebrizer = expression_algebrizer
-            .with_merged_mappings(source.schema(&self.schema_inference_state())?.schema_env)?;
+        let expression_algebrizer = self
+            .with_merged_mappings(source.schema(&self.schema_inference_state())?.schema_env)?
+            // We should not algebrize the select values body exprs in an implicit type conversion
+            // context since there is no sepcific type expected for these values.
```
jchemburkar:
```suggestion
            // context since there is no specific type expected for these values.
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -659,10 +712,13 @@ impl<'a> Algebrizer<'a> {
         if !array_is_literal {
             return Err(Error::ArrayDatasourceMustBeLiteral);
         }
+        // We should not algebrize the array element exprs in an implicit type conversion context
+        // since there is no sepcific type expected for these values.
```
jchemburkar:
```suggestion
        // since there is no specific type expected for these values.
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1208,7 +1288,10 @@ impl<'a> Algebrizer<'a> {
         *self.clause_type.borrow_mut() = ClauseType::OrderBy;
         let expression_algebrizer = self
             .clone()
-            .with_merged_mappings(source.schema(&self.schema_inference_state())?.schema_env)?;
+            .with_merged_mappings(source.schema(&self.schema_inference_state())?.schema_env)?
+            // We should not algebrize the sort key exprs in an implicit type conversion
+            // context since there is no sepcific type expected for these values.
```
jchemburkar:
```suggestion
            // context since there is no specific type expected for these values.
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1260,9 +1343,14 @@ impl<'a> Algebrizer<'a> {
...
-                let expression_algebrizer = self.clone().with_merged_mappings(
-                    source.schema(&self.schema_inference_state())?.schema_env,
-                )?;
+                let expression_algebrizer = self
+                    .clone()
+                    .with_merged_mappings(
+                        source.schema(&self.schema_inference_state())?.schema_env,
+                    )?
+                    // We should not algebrize the group key exprs in an implicit type conversion
+                    // context since there is no sepcific type expected for these values.
```
jchemburkar:
```suggestion
                    // context since there is no specific type expected for these values.
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1355,26 +1442,23 @@ impl<'a> Algebrizer<'a> {
         Ok(mir_node)
     }
 
-    pub fn algebrize_expression(
-        &self,
-        ast_node: ast::Expression,
-        in_implicit_type_conversion_context: bool,
-    ) -> Result<mir::Expression> {
+    /// Algebrizes an `ast::Expression` node into a `mir::Expression` node. Callers must ensure
+    /// that the `Algeberizer`'s `expression_context` is set correctly for the given context of the
```
jchemburkar:
```suggestion
    /// that the `Algebrizer`'s `expression_context` is set correctly for the given context of the
```

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1799,30 +1903,39 @@ impl<'a> Algebrizer<'a> {
...
 
+        // Create an algebrizer that will algebrize expressions in an implicit type conversion
+        // context when necessary (i.e. when the operator expects non-String operands), and one
+        // that will not.
+        let itc_algebrizer = self.with_implicit_type_conversion_ctx(true);
+        let non_itc_algebrizer = self.with_implicit_type_conversion_ctx(false);
+
         // First, we must determine if the left and right operands each need to
         // be algebrized in an implicit type conversion context (this is done
         // by toggling the bool argument to aglebrize_expression). The different
```
jchemburkar:
```suggestion
        // by toggling the bool argument to algebrize_expression). The different
```

mongosql/src/algebrizer/definitions.rs
jchemburkar: locally augment pointed out that we might be able to remove clones before with_merged_mappings in algebrize_join, algebrize_filter_clause, algebrize_order_by_clause and algebrize_group_by. I checked out the branch and tried them, and seems like they compile fine. worth verifying if we can get those for free here

jchemburkar (approved):
LGTM! Just some nitpicks

mongosql/src/algebrizer/definitions.rs
mattChiaravalloti: Whoa, brilliant catch, thank you for catching that. I'll apply those and see how it goes.

mattChiaravalloti: Done

mongosql/src/algebrizer/definitions.rs
```diff
@@ -1040,13 +1106,25 @@ impl<'a> Algebrizer<'a> {
...
-                )?;
+                let expression_algebrizer = self
+                    .with_merged_mappings(
+                        source.schema(&self.schema_inference_state())?.schema_env,
+                    )?
+                    // Although the condition expression is expected to Boolean (or nullish), we
+                    // choose not to algebrize it in an implicit type conversion context. We do this
+                    // because we do not want to confuse users with seemingly inconsistent behavior.
+                    // For example, if the user wrote `... FROM foo WHERE 'false'` and we implicitly
+                    // converted the `'false'` to `fasle`, the filter would fail for all rows. But
```
bucaojit:
```suggestion
                    // converted the `'false'` to `false`, the filter would fail for all rows. But
```
nit: typo

mongosql/src/algebrizer/definitions.rs
```diff
@@ -712,18 +768,27 @@ impl<'a> Algebrizer<'a> {
...
-            .clone()
+        let condition_algebrizer = self
             .with_merged_mappings(left_src_result_set.schema_env)?
-            .with_merged_mappings(right_src_result_set.schema_env)?;
+            .with_merged_mappings(right_src_result_set.schema_env)?
+            // Although the condition expression is expected to Boolean (or nullish), we choose not
+            // to algebrize it in an implicit type conversion context. We do this because we do not
+            // want to confuse users with seemingly inconsistent behavior. For example, if the user
+            // wrote `... FROM foo JOIN bar ON 'false'` and we implicitly converted the `'false'` to
+            // `fasle`, the join condition would fail for all rows. But if they wrote `... FROM foo
```
bucaojit:
```suggestion
            // `false`, the join condition would fail for all rows. But if they wrote `... FROM foo
```
nit

bucaojit (approved):
LGTM, a nice improvement!

mongosql/src/algebrizer/test/expressions/between.rs
mattChiaravalloti: https://jira.mongodb.org/browse/SQL-3381

mattChiaravalloti:
evergreen refresh

PR #180, opened 2026-07-28
SQL-3363: Lateral join fix

nicholascioli:
This PR fixes two somewhat related issues that may appear when chaining multiple `LateralJoin`s together with correlated conditions:

## `LateralJoin` Invalid Invariant

`LateralJoin`s might have the data source defined in only the left branch but used in the right branch. This caused a panic due to a "missing" schema definition for the data source.

## `StageMovement` Source Bubbling Non-Determinism

When deciding which branch of a multi-data-source stage to bubble up, the original code iterated over a `HashSet` (which does not have stable iteration guarantees) in a stateful way which caused queries to bubble non-deterministically. This caused the pipeline to differ between runs on the same query.

### DIFF
--- a/mongosql/src/mir/optimizer/stage_movement/mod.rs
+++ b/mongosql/src/mir/optimizer/stage_movement/mod.rs
@@ -582,7 +582,12 @@ impl StageMovementVisitor<'_> {
             // Recall that the LHS (source) datasources are considered in-scope
             // inside the RHS (subquery) so this is safe.
             Stage::MqlIntrinsic(MqlStage::LateralJoin(ref n)) => {
-                let right_schema = n.subquery.schema(self.schema_state).unwrap();
+                let source_result_set = n.source.schema(self.schema_state).unwrap();
+                let state = self
+                    .schema_state
+                    .with_merged_schema_env(source_result_set.schema_env.clone());
+                let right_schema = n.subquery.schema(&state).unwrap();
+
                 // If this is a filter, we cannot move it, if the Join's JoinType is Left and any use is in the RHS.
                 // It is not semantically correct to merge WHERE conditions into lateral JOIN RHS clauses.
                 if (node.is_filter() && n.join_type == JoinType::Left) || node.is_sort() {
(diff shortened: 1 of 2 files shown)


mongosql/src/mir/optimizer/stage_movement/mod.rs
```diff
@@ -582,7 +582,12 @@ impl StageMovementVisitor<'_> {
             // Recall that the LHS (source) datasources are considered in-scope
             // inside the RHS (subquery) so this is safe.
             Stage::MqlIntrinsic(MqlStage::LateralJoin(ref n)) => {
-                let right_schema = n.subquery.schema(self.schema_state).unwrap();
+                let source_result_set = n.source.schema(self.schema_state).unwrap();
+                let state = self
+                    .schema_state
+                    .with_merged_schema_env(source_result_set.schema_env.clone());
+                let right_schema = n.subquery.schema(&state).unwrap();
```
mattChiaravalloti, line 589: It's very funny to me that the comment on lines 579-583 even calls this out "Recall that the LHS (source) datasources are considered in-scope inside the RHS (subquery)...", and yet we had this bug 😂 😭

mongosql/src/mir/optimizer/stage_movement/test.rs
```diff
@@ -1688,6 +1698,34 @@ test_move_stage_no_op!(
...
+test_move_stage_no_op!(
+    cannot_move_filter_above_lateral_left_join_if_correlated_conditions,
+    Stage::MqlIntrinsic(MqlStage::LateralJoin(LateralJoin {
+        join_type: JoinType::Left,
+        source: mir_collection("foo", "bar"), // "Table3"
+        subquery: Box::new(Stage::Filter(Filter {
+            source: Box::new(Stage::MqlIntrinsic(MqlStage::EquiJoin(EquiJoin {
+                join_type: JoinType::Inner,
+                source: mir_collection("foo", "bar2"), // "Table2"
+                from: mir_collection("foo", "bar3"),   // "Table7"
```
mattChiaravalloti:
```suggestion
        source: mir_collection("foo", "bar"),
        subquery: Box::new(Stage::Filter(Filter {
            source: Box::new(Stage::MqlIntrinsic(MqlStage::EquiJoin(EquiJoin {
                join_type: JoinType::Inner,
                source: mir_collection("foo", "bar2"),
                from: mir_collection("foo", "bar3"), 
```
[nit] I think we don't need these comments since they were only relevant to our investigation, not necessarily to posterity.

mattChiaravalloti (approved):
lgtm 💯

jpowell-mongo (approved):

PR #181, opened 2026-07-28
SQL-3293: Implement algebrization for HigherOrderFunction expressions

mattChiaravalloti:
This PR implements algebrization for `HigherOrderFunction` expressions (including the disambiguation of unqualified identifiers into `Variable`s), as described by the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.ntgguxk2r0tf). This builds off of the refactor that was previously merged in #179 .

This PR touched 10 files, but 6 of those files are simply incidental changes because of updated test macro apis and imports.

The relevant changes are in `algebrizer/definitions.rs`, where the actual algebrization implementations are written, and in `algebrizer/tests/expressions/{higher_order_functions.rs, identifier_and_subpath.rs}`, where the unit tests for all new algebrization is written.

Similar to schema-checking for higher order functions, this implementation introduces a `HigherOrderFunctionWrapper` error variant that wraps algebrizer errors in some additional context to help users understand (1) that an issue happened _within_ a higher order function and (2) what specific component of the higher order function caused it. Fortunately, wrapping errors in the algebrizer is _much_ more straightforward than in the schema checking module.

Building off of #179 , I added an additional field to the `ExpressionContext` that tracks if we are in a higher order function argument context and, if so, which one. Recall from the design doc that since we are not allowing users to specify variable names (they must use `this` and `value`) we also implicitly accepted that shadowing will happen if higher order functions are nested within each other. Keep that in mind when you observe uses of the new context field! Notice that we never actually "restore" previous context since it will never really matter! A higher order function function argument is algebrized with a new algebrizer with properly set state, so subsequent algebrization afterwards is unaffected by its state since it uses a different algebrizer struct.

### DIFF
--- a/mongosql/src/algebrizer/definitions.rs
+++ b/mongosql/src/algebrizer/definitions.rs
@@ -2782,11 +2782,11 @@ impl<'a> Algebrizer<'a> {
     fn algebrize_hof_common(
         &self,
         name: &'static str,
-        array: Box<ast::Expression>,
-        f: Box<ast::FunctionArgument>,
+        array: ast::Expression,
+        f: ast::FunctionArgument,
         hof_ctx: HigherOrderFunctionCtx,
     ) -> Result<(mir::Expression, mir::Expression, bool)> {
-        let array = self.algebrize_expression(*array).map_err(|e| {
+        let array = self.algebrize_expression(array).map_err(|e| {
             Self::wrap_error_in_hof_context(name, e, HigherOrderFunctionErrorCause::ArrayArg)
         })?;
 
@@ -2813,7 +2813,7 @@ impl<'a> Algebrizer<'a> {
         // Nullability is based only on the array argument. The function argument being nullable
         // does not affect the nullability of the Map expression, just the nullability of the
         // elements of the output array.
-        let is_nullable = Self::args_are_nullable(&[array.clone()]);
+        let is_nullable = Self::args_are_nullable(std::slice::from_ref(&array));
 
         Ok((array, f, is_nullable))
     }
@@ -2821,9 +2821,9 @@ impl<'a> Algebrizer<'a> {
     fn algebrize_hof_function_argument(
         &self,
         name: &'static str,
-        f: Box<ast::FunctionArgument>,
+        f: ast::FunctionArgument,
     ) -> Result<mir::Expression> {
-        match *f {
+        match f {
             ast::FunctionArgument::Expr(e) => self.algebrize_expression(e).map_err(|e| {
                 Self::wrap_error_in_hof_context(name, e, HigherOrderFunctionErrorCause::FunctionArg)
             }),
@@ -2835,7 +2835,7 @@ impl<'a> Algebrizer<'a> {
 
     fn algebrize_map(&self, expr: ast::MapExpr) -> Result<mir::Expression> {
         let (array, f, is_nullable) =
-            self.algebrize_hof_common("Map", expr.array, expr.f, HigherOrderFunctionCtx::Map)?;
+            self.algebrize_hof_common("Map", *expr.array, *expr.f, HigherOrderFunctionCtx::Map)?;
 
         Ok(mir::Expression::HigherOrderFunction(
             mir::HigherOrderFunctionApplication::Map(mir::MapExpr {
@@ -2849,8 +2849,8 @@ impl<'a> Algebrizer<'a> {
     fn algebrize_filter_expr(&self, expr: ast::FilterExpr) -> Result<mir::Expression> {
         let (array, f, is_nullable) = self.algebrize_hof_common(
             "Filter",
-            expr.array,
-            expr.f,
+            *expr.array,
+            *expr.f,
             HigherOrderFunctionCtx::Filter,
         )?;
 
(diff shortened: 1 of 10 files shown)


mongosql/src/algebrizer/test/expressions/identifier_and_subpath.rs
```diff
@@ -549,3 +549,176 @@ test_algebrize!(
     expected_error_code = 3008,
     input = ast::Expression::Identifier("bar".into()),
 );
+
+test_algebrize!(
+    unqualified_this_in_map_context_is_variable,
+    method = algebrize_expression,
+    expression_context = ExpressionContext::default()
+        .with_higher_order_function_arg_ctx(HigherOrderFunctionCtx::Map),
```
mattChiaravalloti: This mimics how algebrization within the context of a `Map` expression works. It's saying "if the `ExpressionContext` indicates we're in a `Map`, how would we algebrize this unqualified `Identifier`"?

mongosql/src/algebrizer/test/expressions/identifier_and_subpath.rs
```diff
@@ -549,3 +549,176 @@ test_algebrize!(
...
+                "this".into() => Schema::Atomic(Atomic::Integer),
+            },
+            required: set! {"this".to_string()},
+            additional_properties: false,
+            ..Default::default()
+        }),
+    },
+    variables = map! {
+        "this" => Schema::Atomic(Atomic::Integer),
+    },
```
mattChiaravalloti: Keep an eye on the `variables` presence/value in these tests! If it is present and does not indicate the variable is nullable, `is_nullable` is `false`. If it is not present or is present and indicates the variable is nullable, `is_nullable` is `true`.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -238,6 +242,18 @@ pub struct Algebrizer<'a> {
     allow_order_by_missing_columns: bool,
     clause_type: RefCell<ClauseType>,
     expression_context: ExpressionContext,
+    variables: BTreeMap<&'static str, schema::Schema>,
```
mattChiaravalloti: Similar to `SchemaInferenceState`, we need to track which variables are in scope. I considered putting this in the `ExpressionContext` but ultimately decided against it for reasons I can no longer justify. I think it may make more sense to put this `ExpressionContext` but I'm not sure... Curious to hear your thoughts!

jchemburkar: I think I’m pro ExpressionContext here. Since variables is expression-scoped anyway, it seems cleaner to keep all of that context in one place rather than split across multiple Algebrizer fields. I think that’ll be easier to maintain over time.

mongosql/src/algebrizer/test/expressions/higher_order_functions.rs
```diff
@@ -0,0 +1,1124 @@
+use super::*;
```
jchemburkar, line 1: nit: do we have any tests that demonstrate the desired shadowing behavior you mention in the main pr comment?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -2671,6 +2765,223 @@ impl<'a> Algebrizer<'a> {
...
+            .get_array_item_schema();
+
+        let mut fn_algebrizer = self.with_higher_order_function_arg_ctx(hof_ctx);
+        if let Some(this_schema) = this_schema {
+            fn_algebrizer = fn_algebrizer.with_variables(map! { THIS_VARIABLE => this_schema });
+        }
+        let f = fn_algebrizer.algebrize_hof_function_argument(name, f)?;
+
+        // Nullability is based only on the array argument. The function argument being nullable
+        // does not affect the nullability of the Map expression, just the nullability of the
```
jchemburkar:
```suggestion
        // does not affect the nullability of the HOF expression, just the nullability of the
```

because this can be filter too right

jchemburkar (approved):
LGTM!

mongosql/src/algebrizer/test/expressions/higher_order_functions.rs
mattChiaravalloti: We do not yet, I'll add those. I have shadowing tests for schema checking, but it's important to have them for algebrization too. Will add.

mongosql/src/algebrizer/definitions.rs
mattChiaravalloti: Makes sense, thanks for the push. I'll update this.

mattChiaravalloti: Done!

mongosql/src/algebrizer/test/expressions/higher_order_functions.rs
mattChiaravalloti: Thanks for suggesting this. Turns out there was a minor issue with shadowing -- we weren't merging the local variables with the outer context's variables! We do this properly in schema checking, and it's a very trivial change (use `append` instead of fully overwriting). I wrote small, targeted tests for each HOF, but also one very large test with multiple layers of nesting to ensure this works at multiple scopes at the same time.

bucaojit (approved):
LGTM!

PR #176, opened 2026-07-23
SQL-3295: Implement HigherOrderFunctionRewritePass stub

mattChiaravalloti:
This PR introduces the `HigherOrderFunctionRewritePass` with a stub implementation. The stub creates and invokes the two visitors that will be implemented in follow-up tickets ([SQL-3296](https://jira.mongodb.org/browse/SQL-3296) and SQL-[3297](https://jira.mongodb.org/browse/SQL-3297)). This new rewrite pass is already included in the rewrites list in the correct location, as described by the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.he6cyip700kh). Since the visitors do nothing right now, the pass is effectively a no-op.

PR #183, opened 2026-07-31
SQL-3297: Implement HigherOrderFunctionAliasVisitor

mattChiaravalloti:
This PR implements the `HigherOrderFunctionAliasVisitor` which was previously stubbed out as part of #176 . The design doc describes this visitor [here](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.he6cyip700kh).

This implementation required adding the new array function aliases to the `ast` as `FunctionName` variants. For `ARRAY_CAST`, since the second argument is a `Type`, it could not be added as a normal `FunctionExpr`. Instead, I created an `ArrayCastExpr` which is more akin to `CastExpr`. That was really the only unexpected bump in the road for this one.

I was able to unskip most of the `spec/query/array_functions.yml` tests since we can now parse and rewrite the array function aliases and all the rest of the higher order function work from algebrization onwards is implemented and merged. I was not able to unskip `ARRAY_REMOVE` since we decided to shoot for Snowflake-style semantics on that instead of typical SQL semantics (re: should `ARRAY_REMOVE([1, null, 3], 3)` result in `[1, null]` (like Snowflake) or `[1]` (using SQL 3-value semantics)). I filed a separate ticket to address this.

### DIFF
--- a/mongosql/src/ast/definitions.rs
+++ b/mongosql/src/ast/definitions.rs
@@ -289,6 +289,7 @@ pub enum Expression {
     Extract(ExtractExpr),
     Cast(CastExpr),
     Array(Vec<Expression>),
+    ArrayCast(ArrayCastExpr),
     Subquery(Box<Query>),
     Exists(Box<Query>),
     SubqueryComparison(SubqueryComparisonExpr),
@@ -326,6 +327,12 @@ pub struct CastExpr {
     pub to: Type,
     pub on_null: Option<Box<Expression>>,
     pub on_error: Option<Box<Expression>>,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct ArrayCastExpr {
+    pub expr: Box<Expression>,
+    pub to: Type,
 }
 
 #[derive(PartialEq, Debug, Clone)]
@@ -490,7 +497,6 @@ pub enum FunctionName {
     Millisecond,
 
     // Higher Order Function aliases
-    ArrayCast,
     ArrayExtract,
     ArrayCompact,
     ArrayRemove,
@@ -646,7 +652,6 @@ impl TryFrom<&str> for FunctionName {
             "SECOND" => Ok(FunctionName::Second),
             "MILLISECOND" => Ok(FunctionName::Millisecond),
 
-            "ARRAY_CAST" => Ok(FunctionName::ArrayCast),
             "ARRAY_EXTRACT" => Ok(FunctionName::ArrayExtract),
             "ARRAY_COMPACT" => Ok(FunctionName::ArrayCompact),
             "ARRAY_REMOVE" => Ok(FunctionName::ArrayRemove),
@@ -722,7 +727,6 @@ impl FunctionName {
             FunctionName::Minute => "MINUTE",
             FunctionName::Second => "SECOND",
             FunctionName::Millisecond => "MILLISECOND",
-            FunctionName::ArrayCast => "ARRAY_CAST",
             FunctionName::ArrayExtract => "ARRAY_EXTRACT",
             FunctionName::ArrayCompact => "ARRAY_COMPACT",
             FunctionName::ArrayRemove => "ARRAY_REMOVE",
@@ -795,7 +799,6 @@ impl FunctionName {
             | FunctionName::Minute
             | FunctionName::Second
             | FunctionName::Millisecond
-            | FunctionName::ArrayCast
             | FunctionName::ArrayExtract
             | FunctionName::ArrayCompact
             | FunctionName::ArrayRemove
--- a/mongosql/src/ast/definitions.rs
+++ b/mongosql/src/ast/definitions.rs
@@ -1010,41 +1010,41 @@ pub enum Type {
 
 #[derive(PartialEq, Debug, Clone, VariantCount)]
 pub enum HigherOrderFunctionExpr {
-  Map(MapExpr),
-  Filter(FilterExpr),
-  Reduce(ReduceExpr),
+    Map(MapExpr),
+    Filter(FilterExpr),
+    Reduce(ReduceExpr),
 }
 
 #[derive(PartialEq, Debug, Clone)]
 pub struct MapExpr {
-  pub array: Box<Expression>,
-  pub f: Box<FunctionArgument>,
+    pub array: Box<Expression>,
+    pub f: Box<FunctionArgument>,
 }
 
 #[derive(PartialEq, Debug, Clone)]
 pub struct FilterExpr {
-  pub array: Box<Expression>,
-  pub f: Box<FunctionArgument>,
+    pub array: Box<Expression>,
+    pub f: Box<FunctionArgument>,
 }
 
 #[derive(PartialEq, Debug, Clone)]
 pub struct ReduceExpr {
-  pub array: Box<Expression>,
-  pub init_value: Box<Expression>,
-  pub f: Box<FunctionArgument>,
+    pub array: Box<Expression>,
+    pub init_value: Box<Expression>,
+    pub f: Box<FunctionArgument>,
 }
 
 #[derive(PartialEq, Debug, Clone, VariantCount)]
 pub enum FunctionArgument {
-  Expr(Expression),
-  NamedFunction(NamedFunction)
+    Expr(Expression),
+    NamedFunction(NamedFunction)
 }
 
 #[derive(PartialEq, Debug, Clone, VariantCount)]
 pub enum NamedFunction {
-  UnaryOp(UnaryOp),
-  BinaryOp(BinaryOp),
-  Function(FunctionName),
+    UnaryOp(UnaryOp),
+    BinaryOp(BinaryOp),
+    Function(FunctionName),
 }
 
 } // end of generate_visitors! block
--- a/mongosql/src/ast/definitions.rs
+++ b/mongosql/src/ast/definitions.rs
@@ -488,6 +488,19 @@ pub enum FunctionName {
     Minute,
     Second,
     Millisecond,
+
+    // Higher Order Function aliases
+    ArrayCast,
+    ArrayExtract,
+    ArrayCompact,
+    ArrayRemove,
+    ArrayCountIf,
+    ArraySum,
+    ArrayProduct,
+    ArrayAverage,
+    ArrayAll,
+    ArrayAny,
+    ArrayJoin,
 }
 
 impl TryFrom<FunctionName> for TrimSpec {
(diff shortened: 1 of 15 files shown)


mongosql/src/ast/rewrites/test.rs
```diff
@@ -1397,3 +1397,119 @@ mod scalar_functions {
...
+        array_count_if,
+        pass = HigherOrderFunctionsRewritePass,
+        expected = Ok("SELECT SIZE(FILTER(a, this > y))"),
+        input = "SELECT ARRAY_COUNT_IF(a, this > y)",
+    );
+
+    test_rewrite!(
+        array_sum,
+        pass = HigherOrderFunctionsRewritePass,
+        expected = Ok("SELECT REDUCE(a, 0, `value` + this)"),
```
mattChiaravalloti, line 1507: `value` is a keyword in mongosql (see: `SELECT VALUE`), which is why it is delimited. When we close this project and submit docs changes, I'll be sure to note that `value` and `this` should both be delimited when used in user queries.

mongosql/src/ast/definitions.rs
```diff
@@ -960,41 +1010,41 @@ pub enum Type {
 
 #[derive(PartialEq, Debug, Clone, VariantCount)]
 pub enum HigherOrderFunctionExpr {
-  Map(MapExpr),
-  Filter(FilterExpr),
-  Reduce(ReduceExpr),
+    Map(MapExpr),
```
mattChiaravalloti, line 1013: Spacing in this section was inconsistent with the rest of the file (2 spaces vs 4)

mongosql/src/ast/pretty_print.rs
```diff
@@ -57,7 +57,7 @@ lazy_static! {
         r"(?i)inner$",
         r"(?i)int$",
         r"(?i)integer$",
-        r"(?i)is$",
+        r"(?i)^is$",
```
mattChiaravalloti, line 60: Arguably we should do this for all keywords (exact matches) but I didn't want to make such a large/drastic change as part of this mostly unrelated PR. For now, this pattern was forcing anything ending in `is` (e.g., `this`) to be delimited. I don't want that to be an incidental thing, though. Currently, I left it so `this` does not actually need to be delimited because it is not actually a keyword. I could, though, make it a keyword. If we choose to go that route, I'd rather explicitly add `this` to the keyword list here than have it be caught accidentally by this pattern. Let me know what you think!

mongosql/src/mir/schema/test/expressions/higher_order_function.rs
```diff
@@ -237,6 +237,24 @@ mod map {
             is_nullable: false,
         })),
     );
+
+    // The output array must not contain missing, since it is upconverted to null.
```
mattChiaravalloti, line 241: I encountered a small schema bug with higher order functions that caused some spec test failures! The issue is that we were not properly upconverting missing to null within `map`'s output array. That has been addressed in this PR and this test demonstrates the fix works.

mongosql/src/mir/schema/test/expressions/higher_order_function.rs
```diff
@@ -1059,4 +1077,28 @@ mod reduce {
                 is_nullable: false,
             })),
     );
+
+    // The output schema must not be possibly missing, as missing should be upconverted to null.
```
mattChiaravalloti, line 1081: The same issue applies to `reduce` when its output is possibly missing. Generally, functions do not evaluate to missing in mongosql -- those missing values are upconverted to null (just like in MongoDB).

testdata/spec_tests/query_tests/array_functions.yml
```diff
@@ -9,21 +9,21 @@ dataset:
...
-            bsonType: "int"
-          a:
-            anyOf:
-              - bsonType: "array"
-                items:
-                  anyOf:
-                    - bsonType: "int"
-                    - bsonType: !!str "null"
-              - bsonType: !!str "null"
+    schema:
```
mattChiaravalloti, line 12: Spacing was incorrect in this file, so schemas were not being written to the db. Fixed here.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -18,12 +25,377 @@ impl Pass for HigherOrderFunctionsRewritePass {
...
+    fn rewrite_array_extract(args: &[Expression]) -> Result<Expression> {
+        assert_arg_count("ARRAY_EXTRACT", args.len(), ArgCount::Exactly(2))?;
+
+        let array = &args[0];
+        let extract_expr = &args[1];
+
+        // If the second argument is a field-path-like Expression, we should prepend "this" to it.
+        // If it is not, we should wrap it in an AccessExpr with "this" as the base.
+        let f = prepend_parent_to_field_path_expr(THIS, extract_expr).unwrap_or_else(|| {
+            Expression::Access(AccessExpr {
```
mattChiaravalloti, line 178: Recall that an `AccessExpr` is typically translated to a `ComputedFieldAccess`, whereas a `SubpathExpr` is translated into a `FieldAccess` (which ultimately becomes a field (or variable) reference in aggregation, e.g. `$$this.a.b.c`). This `unwrap_or_else` is effectively saying "if this expression does not resemble something like a field path expression, just make it an `AccessExpr`".

mongosql/src/ast/rewrites/higher_order_functions.rs
bucaojit, line 26: I was looking at other `apply()` implementations and I see that the error is being checked first, such as [here](https://github.com/mongodb/mongosql/blob/ffad08845a58cb5a17c9cd745f1df3f77e963ecc/mongosql/src/ast/rewrites/scalar_functions.rs#L43).   Looks to me like we should be adding a similar check?

bucaojit (approved):
LGTM, I had one comment about the error checking.

mongosql/src/mir/schema/test/expressions/higher_order_function.rs
bucaojit: Nice!

mongosql/src/ast/rewrites/higher_order_functions.rs
mattChiaravalloti: Oh, ha! Of course. Thank you for catching this -- I got so focused on adding the `error` field and setting it throughout the rewrite, I forgot to actually use the value. Updated!

mongosql/src/ast/rewrites/mod.rs
```diff
@@ -100,3 +100,49 @@ pub fn rewrite_query(query: ast::Query) -> Result<ast::Query> {
...
+    /// Either of these many arguments.
+    Either(usize, usize),
+}
+
+impl ArgCount {
+    /// Returns whether `n` satisfies this argument count specification.
+    pub(crate) fn contains(&self, n: usize) -> bool {
+        match self {
+            ArgCount::Exactly(count) => n == *count,
+            ArgCount::Either(min, max) => n >= *min && n <= *max,
```
nicholascioli: Nit: Shouldn't this be an equality check on the bounds? Or is this supposed to be an inclusive range check?


```suggestion
            ArgCount::Either(min, max) => n == *min || n == *max,
```

mongosql/src/ast/rewrites/mod.rs
```diff
@@ -100,3 +100,49 @@ pub fn rewrite_query(query: ast::Query) -> Result<ast::Query> {
     }
     Ok(rewritten)
 }
+
+/// Specifies how many arguments a function accepts.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum ArgCount {
```
nicholascioli, line 106: Suggestion (non-blocking): After looking through the changes, I see that this is mostly used in a `const` way, so it might be cool to allow for this to act as a type-safe way to extract the arguments themselves with compiler guarantees.

I wrote a quick example of how this might look like here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f245266c5c3edf3e4102c0e992cc81fc

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -1,29 +1,405 @@
...
+                    }
+                }
+            }
+            _ => node,
+        }
+    }
+}
+
+impl HigherOrderFunctionsAliasVisitor {
+    fn make_map(array: Expression, f: Expression) -> Expression {
```
nicholascioli, line 104: Nit (non-blocking): You could probably mark these smaller helper functions as `#[inline(always)]`.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -1,29 +1,405 @@
...
+        Ok(Self::make_binary(
+            rewritten_sum,
+            BinaryOp::Div,
+            Self::make_size(array.clone()),
+        ))
+    }
+
+    /// Rewrite `ARRAY_JOIN(a)` or `ARRAY_JOIN(a, '')` into `REDUCE(a, '', value || this)`,
+    /// and rewrite `ARRAY_JOIN(a, sep)` into
+    /// `TRIM(LEADING sep FROM REDUCE(a, '', value || sep || this))`.
```
nicholascioli, line 260: Question (non-blocking): Why do we need to trim here? Since reduce is given '' as its initial value, won't it never start with the separator?

mongosql/src/ast/pretty_print.rs
nicholascioli, line 60: Question (non-blocking): Is there a reason for the `^` here but not in `mongosql.lalrpop`? Also, why is `is` the only case in this list that has a `^`?

nicholascioli (approved):
LGTM! I left a few non-blocking comments, but everything looks great as-is.

mongosql/src/ast/rewrites/higher_order_functions.rs
mattChiaravalloti: Since `''` is the initial value, it will always start with the separator since the initial evaluation of `value || sep || this` will be `'' || '<sep>' || '<this>'`. That means if the array has at least one element, the accumulated result after processing the first element will be `'<sep><this>'`, which has a leading separator that should not be there! 

We could alternatively do some conditional checking inside the rewritten function, like `CASE value = '' THEN this ELSE value || sep || this` but that leads to a less efficient translation since it uses a `$switch` expression within the `$reduce`.

mongosql/src/ast/pretty_print.rs
mattChiaravalloti: I discussed that here: https://github.com/mongodb/mongosql/pull/183#discussion_r3692547931

The tldr is:
1. in the parser, we don't use end all keyword patterns with `$`, we do exact matching, so there's no need for `^`. (I suppose I can remove the `$` instead of introducing the `^` here!)
2. `is` is the only one since I don't want to make _too_ many unrelated changes in this PR. Changing how we delimit all of these would be potentially pretty drastic so I'd rather have a separate ticket to track that.

mongosql/src/ast/rewrites/mod.rs
mattChiaravalloti: Oh whoops yes, I initially implemented this as "Between" then realized later that it was used as an "Either" so I changed the name but not the condition! Good catch.

mongosql/src/ast/rewrites/higher_order_functions.rs
nicholascioli: Whoops I'm dumb. I'll resolve this one

mongosql/src/ast/pretty_print.rs
nicholascioli: Wow for some reason that comment did not render for me. Sounds good!

mattChiaravalloti: Filed https://jira.mongodb.org/browse/SQL-3409 to track this

mongosql/src/ast/rewrites/higher_order_functions.rs
mattChiaravalloti: done!

mongosql/src/ast/rewrites/mod.rs
mattChiaravalloti: Done!

PR #184, opened 2026-07-31
SQL-673: Implement support for ComputedFieldAccess

mattChiaravalloti:
This PR addresses a long-standing issue that came to light as part of the recent array function project: our lack of support for computed field access expressions. SQL-673 has been unblocked for several years, and with `ARRAY_EXTRACT` now implemented and potentially creating `AccessExpr`s syntactically, I thought it would be good to avoid the error that results from attempting to codegen a `ComputedFieldAccess` expression.

This PR updates `air::GetField` to accept an arbitrary `Expression` instead of a `String`. It also replaces the `air::SqlOperator::ComputedFieldAccess` with `air::GetField`. This flows from the update to the mir which removes `ScalarFunction::ComputedFieldAccess` in favor of a native `mir::Expression::ComputedFieldAccess` variant. These changes are plumbed through the full translation pipeline and the spec test, along with migrated/new unit test and manually testing, works as expected.


mongosql/src/air/agg_ast/ast_definitions.rs
```diff
@@ -443,7 +443,7 @@ impl From<TaggedOperator> for air::Expression {
     fn from(ast_op: TaggedOperator) -> Self {
         match ast_op {
             TaggedOperator::GetField(gf) => air::Expression::GetField(air::GetField {
-                field: gf.field,
+                field: air::Expression::Literal(air::LiteralValue::String(gf.field.clone())).into(),
```
mattChiaravalloti: We still support existing `GetField` uses which exclusively expect a literal string for the `field` value. This change now also supports arbitrary expressions for `GetField`. This feature was only introduced in server version 7.2, which technically breaks our "6.0 baseline" support. However, since `ComputedFieldAccess` usage previously always resulted in an error, I thought this was an acceptable breaking of that baseline.

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -3771,166 +3771,6 @@ mod merge_objects {
     );
 }
 
-mod computed_field_access {
```
mattChiaravalloti, line 3774: tests migrated to their own file

mongosql/src/mir/schema/test/expressions/computed_field_access.rs
```diff
@@ -0,0 +1,132 @@
+use crate::{
```
jchemburkar, line 1: nit: file is missing an s!

mongosql/src/mir/schema/test/expressions/mod.rs
```diff
@@ -1,5 +1,6 @@
 mod aggregate; // mir::Aggregate
 mod case; // mir::Expression::{SearchedCase, SimpleCase}
+mod computed_field_acces; // mir::Expression::ComputedFieldAccess
```
jchemburkar:
```suggestion
mod computed_field_access; // mir::Expression::ComputedFieldAccess
```

mongosql/src/codegen/expressions.rs
```diff
@@ -185,9 +181,15 @@ impl MqlCodeGenerator {
     }
 
     fn codegen_get_field(&self, gf: air::GetField) -> Result<Bson> {
+        let input = self.codegen_expression(*gf.input)?;
+        let field = match *gf.field {
+            air::Expression::Literal(air::LiteralValue::String(s)) => {
+                Self::wrap_in_literal_if(s.clone(), |s| s.starts_with('$'))
```
jchemburkar: is the clone necessary?

mongosql/src/algebrizer/definitions.rs
```diff
@@ -2269,16 +2269,21 @@ impl<'a> Algebrizer<'a> {
...
+                // Since this expression is certainly not a StringConstructor, we can just use
+                // `self` to algebrize it since the implicit type conversion context does not
+                // matter.
+                field: Box::new(self.algebrize_expression(sf)?),
+                // We cannot know reliably one way or another if the field access is nullable, so we
+                // conservatively say it is. This is because the name of the `field` itself is a
+                // computed expression that will not be known until runtime. It is possible for
+                // both `expr` and `field` to be non-nullable, but for `field` to evaluate at
+                // runtime to a value that does not exist in the document `expr` and therefore
+                // result in missing. It is also possible for `field` to evaluate to a value that
```
jchemburkar, line 2283: I just want to double check this is the intended behavior, and if so, I think we might want to add a test. Here is a test that I think describes the missing behavior from this comment, but fails:
```
test_constant_fold_no_op!(
    computed_field_missing_should_be_noop,
    Stage::Array(ArraySource {
        alias: "".into(),
        array: vec![Expression::ComputedFieldAccess(ComputedFieldAccess::new(
            Box::new(Expression::Document(
                    unchecked_unique_linked_hash_map! {"a".into() => Expression::Literal(LiteralValue::Integer(2))}
                .into())),
            Box::new(Expression::Literal(LiteralValue::String("b".into()))),
        ))],
        cache: SchemaCache::new(),
    })
);
```
let me know if I'm following correctly!

jchemburkar (changes requested):
I have one potential change in mind (and a couple nits), but otherwise looks good to me!

mongosql/src/air/agg_ast/ast_definitions.rs
jchemburkar: agreed!

mongosql/src/algebrizer/definitions.rs
mattChiaravalloti: This test currently exists in `optimizer/constant_folding/test.rs`:
```
test_constant_fold!(
    computed_field_missing,
    expected = Stage::Array(ArraySource {
        alias: "".into(),
        array: vec![Expression::Literal(LiteralValue::Null)],
        cache: SchemaCache::new(),
    }),
    expected_changed = true,
    input = Stage::Array(ArraySource {
        alias: "".into(),
        array: vec![Expression::ComputedFieldAccess(ComputedFieldAccess::new(
            Box::new(Expression::Document(
                    unchecked_unique_linked_hash_map! {"a".into() => Expression::Literal(LiteralValue::Integer(2))}
                .into())),
            Box::new(Expression::Literal(LiteralValue::String("b".into()))),
        ))],
        cache: SchemaCache::new(),
    }),
);
```
We can constant fold a `ComputedFieldAccess` expression only when the `field` is string literal and the `expr` is a document literal. So in this test, as in yours, the inputs are a literal document (`{a: 2}`) and a literal string (`'b'`). Since accessing `{a: 2}['b']` results in `Missing`, we know when constant folding that `Missing` is upconverted to `Null`. That's where the existing test diverges from your test -- we _can_ and _do_ upconvert this missing value to null, whereas your test suggests no-oping. Since both inputs are literals, we know they won't change at runtime so it is safe to do this.

That last sentence might be where the misunderstanding is? Or is it something else with between the algebrizer comment here and the constant folding behavior in the optimizer?

One thing I'm realizing is that we _could_ do a little better than "always nullable" but the infrastructure to support it would be pretty bulky. Specifically, if the `expr` has a static `Document` schema with `additional_fields: false` and the the `field` is a string literal, we can know for sure if a field is _not_ nullable. That type of input could also be a no-op in the constant folder, if that's what you have in mind. Something like `doc_field_with_non_required_b_field['b']` -- just because `'b'` isn't "required" in `doc_field_with_non_required_b_field` doesn't mean we can constant fold this field access to `null` since some instances of `doc_field_with_non_required_b_field` may actually contain data for `'b'`. Is this more in line with what you were thinking, or does the earlier test/explanation already address what you had in mind?

mongosql/src/codegen/expressions.rs
mattChiaravalloti: Ah good catch, might have been a remnant of a different factoring of this code while I was working on this or could've just been a thoughtless mistake introducing it. Either way, removed 👍

mongosql/src/mir/schema/test/expressions/mod.rs
mattChiaravalloti: Done

mongosql/src/mir/schema/test/expressions/computed_field_access.rs
mattChiaravalloti: Done

mongosql/src/air/agg_ast/ast_definitions.rs
nicholascioli: Nit: Does the `clone` need to be here?

mongosql/src/codegen/expressions.rs
```diff
@@ -185,9 +181,15 @@ impl MqlCodeGenerator {
     }
 
     fn codegen_get_field(&self, gf: air::GetField) -> Result<Bson> {
+        let input = self.codegen_expression(*gf.input)?;
+        let field = match *gf.field {
+            air::Expression::Literal(air::LiteralValue::String(s)) => {
+                Self::wrap_in_literal_if(s, |s| s.starts_with('$'))
```
nicholascioli, line 192: Question (non-blocking): Do we not have a constant or function somewhere that describes what checking for a leading `$` means? I think it would help to use something like that over raw characters and it would help describe the _intent_ behind why we check for a leading `$`.

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -593,20 +593,23 @@ impl ConstantFoldExprVisitor<'_> {
...
 
-    // Constant folds the computed field function
-    fn fold_computed_field_function(
-        &mut self,
-        left: Expression,
-        right: Expression,
-    ) -> Option<Expression> {
+    // Constant folds the computed field access expr
+    fn fold_computed_field_access_expr(&mut self, expr: ComputedFieldAccess) -> (Expression, bool) {
+        if self.has_null_arg(&[*expr.expr.clone(), *expr.field.clone()]) {
```
nicholascioli: Nit (non-blocking): The clone here makes me a little sad. I looked into `has_null_arg` and saw that it only really needs references to expressions. Its signature could be refactored to look like the following:

```rust
fn has_null_arg<'a>(&self, args: impl IntoIterator<Item = &'a Expression>) -> bool { ... }
```

Then this code code simply be:


```suggestion
        if self.has_null_arg([&expr.expr, &expr.field]) {
```

This should be considered non-blocking; I just like removing `.clone`s preemptively

nicholascioli (approved):
LGTM! I left a few nits / questions, but none should be considered blocking.

mongosql/src/codegen/expressions.rs
mattChiaravalloti: Ah yea, I see that we only document this in the `translator` and not again here in `codegen`. I'll add comments to explain why.

For your understanding here in this github comment: in aggregation language, a string that starts with a `$` is considered a field reference expression (e.g., `"$foo"` is the field reference expression for `foo`, not the string literal `"$foo"`). So, when setting the `field` argument to the `$getField` expression, if the field's name begins with `$`, we need to wrap that value in the `$literal` expression to indicate to mongodb that the value is a string literal, not a field reference.

mongosql/src/mir/optimizer/constant_folding/lib.rs
mattChiaravalloti: Thanks, happy to update 🎉

mongosql/src/air/agg_ast/ast_definitions.rs
mattChiaravalloti: done!

mongosql/src/algebrizer/definitions.rs
jchemburkar: ah ok. maybe i was just over-eager with where i placed this test / didn't get the optimization quite right. if we have coverage for this i think it makes sense, I had just had a hard time initially following the null+missing semantics of this case. the explanation here is really what i was interested in, I dont think we need to edge case this _too_ hard!

jchemburkar (approved):

PR #185, opened 2026-08-05
SQL-3296: Implement FunctionArgumentVisitor

mattChiaravalloti:
This PR implements the `FunctionArgumentVisitor` which was previously stubbed out as part of #176. ~**Importantly, I rebased this branch on top of #183. When reviewing, I advise omitting the first commit in this PR since it is a squashed version of #183, which is quite large!**~ The design doc describes this visitor [here](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.he6cyip700kh).

This implementation is more self-contained than #183 since it does not involve introducing any new function names or any new parser support.

No spec or end-to-end tests can be skipped as a result of this change. Those are all still waiting on parser support for higher order functions, which will be unblocked once this and #183 are merged.

Edit: #183 has been merged so I rebased on top of that and this is now in much cleaner state to review.

### DIFF
--- a/mongosql/src/ast/rewrites/higher_order_functions.rs
+++ b/mongosql/src/ast/rewrites/higher_order_functions.rs
@@ -339,7 +339,33 @@ impl FunctionArgumentVisitor {
             FunctionArgument::NamedFunction(NamedFunction::BinaryOp(BinaryOp::Sub)) => {
                 FunctionArgument::Expr(Self::rewrite_named_function_to_unary_expr(UnaryOp::Neg))
             }
-            FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op)) => {
+            // Exhaustively match other BinaryOps so in the future if we add new binary operators
+            // with overloaded forms, we will be forced to handle them here.
+            FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::And))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::Concat))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::Div))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::In))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::Mul))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::NotIn))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op @ BinaryOp::Or))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(
+                op @ BinaryOp::Comparison(ComparisonOp::Eq),
+            ))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(
+                op @ BinaryOp::Comparison(ComparisonOp::Gt),
+            ))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(
+                op @ BinaryOp::Comparison(ComparisonOp::Gte),
+            ))
+            | FunctionArgument::NamedFunction(NamedFunction::BinaryOp(
+                op @ BinaryOp::Comparison(ComparisonOp::Lt),
+            ))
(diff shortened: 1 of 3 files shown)


mongosql/src/ast/rewrites/test.rs
```diff
@@ -20,6 +20,24 @@ macro_rules! test_rewrite {
     };
 }
 
+// For any features that do not yet have parser support but do have `ast` support, use this macro.
```
mattChiaravalloti, line 23: As this comment says, I added this macro to help support writing `rewriter` tests for new `ast` features that do not yet have parser support. For now, that is just the `HigherOrderFunctionExpr` variants. In the future, there may be other new additions that could benefit from this macro.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -122,16 +126,6 @@ impl HigherOrderFunctionsAliasVisitor {
         }))
     }
 
-    /// Returns the `this` identifier expression used within higher order function bodies.
```
mattChiaravalloti, line 128: These are useful for both visitors, so I made them free functions in this file as opposed to methods on `HigherOrderFunctionsAliasVisitor`.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -287,20 +281,155 @@ impl HigherOrderFunctionsAliasVisitor {
...
+            FunctionArgument::NamedFunction(NamedFunction::UnaryOp(op)) => {
+                FunctionArgument::Expr(Self::rewrite_named_function_to_unary_expr(op))
+            }
+            FunctionArgument::NamedFunction(NamedFunction::BinaryOp(BinaryOp::Add)) => {
+                FunctionArgument::Expr(Self::rewrite_named_function_to_unary_expr(UnaryOp::Pos))
+            }
+            FunctionArgument::NamedFunction(NamedFunction::BinaryOp(BinaryOp::Sub)) => {
+                FunctionArgument::Expr(Self::rewrite_named_function_to_unary_expr(UnaryOp::Neg))
+            }
+            FunctionArgument::NamedFunction(NamedFunction::BinaryOp(op)) => {
```
mattChiaravalloti: Do you think we should expand this pattern to all `BinaryOp` variants like:
```
FunctionArgument::NamedFunction(NamedFunction::BinaryOp(BinaryOp::Mul))
| FunctionArgument::NamedFunction(NamedFunction::BinaryOp(BinaryOp::Div))
| ...
```
The reason we would do this is so that in future, if we add new `BinaryOp`s, if they happen to have a corresponding overloaded unary operator version, we would need to explicitly consider this code because the compiler would force us to look here. Without an exhaustive list, new operators will just fall to this wildcard case.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -287,20 +281,155 @@ impl HigherOrderFunctionsAliasVisitor {
...
+    fn rewrite_binary_context_arg(&mut self, f: FunctionArgument) -> FunctionArgument {
+        match f {
+            FunctionArgument::Expr(_) => f,
+            FunctionArgument::NamedFunction(NamedFunction::UnaryOp(UnaryOp::Pos)) => {
+                FunctionArgument::Expr(Self::rewrite_named_function_to_binary_expr(BinaryOp::Add))
+            }
+            FunctionArgument::NamedFunction(NamedFunction::UnaryOp(UnaryOp::Neg)) => {
+                FunctionArgument::Expr(Self::rewrite_named_function_to_binary_expr(BinaryOp::Sub))
+            }
+            FunctionArgument::NamedFunction(NamedFunction::UnaryOp(op)) => {
```
mattChiaravalloti: Same question here, except for explicitly listing all `UnaryOp` variants.

mongosql/src/ast/rewrites/higher_order_functions.rs
jchemburkar: Yes, I think I'm in favor of avoiding wildcards / any behavior that could be sneaky. This is a good callout

jchemburkar (approved):
LGTM! Agree with your callout, Im in favor of spelling out the matches in both cases. Nice work.

nicholascioli (approved):
LGTM!

PR #186, opened 2026-08-10
SQL-3305: Wrap Expr Language Binary Comparisons in Null Guards

jpowell-mongo:
## Summary

This pull request updates the `SqlNullSemanticsOperatorsDesugarerPass` to add a null guard to field references and variables that are used in binary comparisons in the context of a Match Stage. 

Aggregation pipelines that look like this:
```javascript
[{ $match: { $expr: { $gt: ["$a", 0] }} }]
```

Would now add an $and operation with a null check:
```javascript
[{ $match: { $expr: { $and: [ { $gt: ["$a", null] }, { $gt: ["$a", 0] }]}}}]
```

This optimization is meant to encourage index utilization for nullable fields.

## Changes

1. Introduce a new method `null_guard_match_predicates` that traverses an Expression, and wraps binary comparisons in null guards.
2. `SqlNullSemanticsOperatorsDesugarerPass` now runs this method before further desugaring.
3. Unit tests to validate changes.


mongosql/src/air/desugarer/sql_null_semantics_operators.rs
```diff
@@ -363,9 +363,136 @@ impl SqlNullSemanticsOperatorsDesugarerVisitor {
...
+                            op: sql_op_to_mql_op(sql_operator.op).unwrap(),
+                            args: new_args,
+                        })
+                    }
+                    // Any other operator is left for the main desugarer pass to handle.
+                    _ => SqlSemanticOperator(sql_operator),
+                }
+            }
+            // We match on the switch (Case statement) because we just want to wrap the expressions within the cases, not the entire statement
+            Switch(switch_expr) => {
```
jpowell-mongo: TODO: Revisit if we actually need this match arm

mongosql/src/air/desugarer/testdata/desugar_sql_null_semantics.yml
mattChiaravalloti, line 1: Can we add coverage for the remaining comparison operators that are not covered yet:
- `$sqlEq` in a match stage (at least one field ref that gets null-guarded)
- `$sqlLte` in a match stage (at least one field ref that gets null-guarded)
- `$sqlNe` in a match stage (at least one field ref that gets null-guarded)

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
```diff
@@ -363,9 +363,114 @@ impl SqlNullSemanticsOperatorsDesugarerVisitor {
...
         }
     }
+
+    /// Null-guards a `$match` predicate's `FieldRef`/`Variable` comparison operands.
+    ///
+    /// For a binary comparison (`Lt`/`Lte`/`Gt`/`Gte`/`Eq`/`Ne`), any `FieldRef` or `Variable`
+    /// operand is guarded by a preceding `$gt(<operand>, null)` check. This
+    ///  keeps the comparison index-eligible. Compound expressions like function calls
+    /// are not index-eligible, so those operators are left untouched.
+    /// Also recurses into `And`/`Or` (applying the above to each argument).
```
mattChiaravalloti, line 373: Can this also recurse into `Not`?

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
```diff
@@ -363,9 +363,114 @@ impl SqlNullSemanticsOperatorsDesugarerVisitor {
...
+                        // 2. Add the comparison itself: rebuilt as Mql if both sides are simple,
+                        // otherwise left as the original Sql operator for desugar_sql_op to handle.
+                        and_args.push(if is_simple(&lhs) && is_simple(&rhs) {
+                            MqlSemanticOperator(air::MqlSemanticOperator {
+                                op: sql_op_to_mql_op(sql_operator.op).unwrap(),
+                                args: vec![lhs, rhs],
+                            })
+                        } else {
+                            SqlSemanticOperator(sql_operator)
+                        });
```
mattChiaravalloti: [question/suggestion] 2 thoughts on this:
1. For readability, can we extract the conditional out of the `and_args.push(...)` expression into a named variable? It is a very large expression to provide as an argument directly to `and_args.push`.
2. Since we only update the expression if `is_simple(&lhs) && is_simple(&rhs)`, can we perform that check before we create the null guards? There's no need to even spend time/memory creating the null guards if we know upfront that we aren't going to be able to rewrite.

mattChiaravalloti: I'd propose rewriting this like:
```
Lt | Lte | Gt | Gte | Eq | Ne => {
    // We use is_simple to determine if we can convert the comparison to MQL.
    // Literals, FieldRefs and variables could potentially use an index so we consider those.
    // Scalar functions and other complex expressions don't qualify, so we just exclude them here.
    let is_simple = |e: &Expression| matches!(e, Literal(_) | FieldRef(_) | Variable(_));

    // Only FieldRef/Variable operands are worth guarding here - that's what
    // keeps the comparison index-eligible.
    let requires_null_guard = |e: &&Expression| matches!(e, FieldRef(_) | Variable(_));

    if !sql_operator.args.iter().all(is_simple) {
        return SqlSemanticOperator(sql_operator)
    }
    
    let lhs = sql_operator.args[0].clone();
    let rhs = sql_operator.args[0].clone();

    let and_args = [&lhs, &rhs]
        .into_iter()
        .filter(requires_null_guard)
        .cloned()
        .map(|operand| {
            // Return an expression that checks if the operand is null
            MqlSemanticOperator(air::MqlSemanticOperator {
                op: MqlOperator::Gt,
                args: vec![operand, Literal(LiteralValue::Null)],
            })
        })
        .chain(std::iter::once(MqlSemanticOperator(
            air::MqlSemanticOperator {
                op: sql_op_to_mql_op(sql_operator.op).unwrap(),
                args: sql_operator.args, // or, `vec![lhs, rhs]`
            },
         )))
        .collect();

    MqlSemanticOperator(air::MqlSemanticOperator {
        op: MqlOperator::And,
        args: and_args,
    })
}
```

This exits as early as possible and avoids extra conditional logic later for creating the new set of arguments to `And`.

mattChiaravalloti (changes requested):
Nice changes. I agree with what you said offline that this might not end up applying to too many expressions since match_null_filtering and rewrite_to_match_language should mostly prevent us from ever having sql-semantic comparison operators in `$match`/`$expr` stages. Still, it's good to have this implemented optimally and correctly! I left some suggestions for a few missing tests and a proposed rewrite to make certain code exit earlier when possible.

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
jpowell-mongo: Yes, I add Not to the Match Arm

mongosql/src/air/desugarer/testdata/desugar_all.yml
```diff
@@ -1276,29 +1259,7 @@ tests:
...
-                                        ],
-
-                                      { "$literal": null },
-                                      { "$not": [ "$$desugared_sqlNot_input0" ] },
-                                    ],
-                                },
-                            },
-                        }
-                    }
+                  "$match": { "$expr": { "$not": ["$expr"] } },
```
mattChiaravalloti: Is this correct? Don't we still need the guard since `{$not: null}` is true in mongodb but should still be false in the sql context? Example:
```
Enterprise test> db.test.insertMany([{_id: 0, a: true}, {_id: 1, a: false}, {_id: 2, a: null}, {_id: 3}])
{ acknowledged: true, insertedIds: { '0': 0, '1': 1, '2': 2, '3': 3 } }
Enterprise test> db.test.aggregate([{$match: {$expr: {$not: ["$a"]}}}])
[ { _id: 1, a: false }, { _id: 2, a: null }, { _id: 3 } ]
Enterprise test> db.test.aggregate([{$match: {$expr: {$let: {vars: {desugared_sqlNot_input0: "$a"}, in: {$cond: [{$lte: ["$$desugared_sqlNot_input0", null]}, {$literal: null}, {$not: ["$$desugared_sqlNot_input0"]}]}}}}}])
[ { _id: 1, a: false } ]
```
Notice the difference in result sets when using the new translation vs the old.

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
```diff
@@ -363,9 +363,118 @@ impl SqlNullSemanticsOperatorsDesugarerVisitor {
...
+                            // Append the original comparison to the end of the list of null guards
+                            .chain(std::iter::once(comparison))
+                            .collect();
+
+                        MqlSemanticOperator(air::MqlSemanticOperator {
+                            op: MqlOperator::And,
+                            args: and_args,
+                        })
+                    }
+                    And | Or | Not => {
```
mattChiaravalloti: It seems `$not` actually cannot operate the same way as `$and` and `$or`:
```
Enterprise test> db.test.aggregate([{$project: {_id: 0, v: {$not: ["$a"]}}}])
[ { v: false }, { v: true }, { v: true }, { v: true } ]
Enterprise test> db.test.aggregate([{$project: {_id: 0, v: {$and: [true, "$a"]}}}])
[ { v: true }, { v: false }, { v: false }, { v: false } ]
Enterprise test> db.test.aggregate([{$project: {_id: 0, v: {$and: [false, "$a"]}}}])
[ { v: false }, { v: false }, { v: false }, { v: false } ]
Enterprise test> db.test.aggregate([{$project: {_id: 0, v: {$or: [true, "$a"]}}}])
[ { v: true }, { v: true }, { v: true }, { v: true } ]
Enterprise test> db.test.aggregate([{$project: {_id: 0, v: {$or: [false, "$a"]}}}])
[ { v: true }, { v: false }, { v: false }, { v: false } ]
```

We don't need to hold up this PR supporting it. My previous comment was truly a question when I asked "Can this also recurse into `Not`?" I didn't know the answer but appears the answer is actually no, or at least "not in the same way as `And` and `Or` and not without extra work".

mattChiaravalloti (changes requested):
Refactor and additional testing looks good. There are now correctness issues though so we need to address those before merging. I think we can just roll back the `Not` changes for now.

mongosql/src/air/desugarer/sql_null_semantics_operators.rs
jpowell-mongo: Okay, I'll revert the Not changes and fix the correctness issues in the tests. Yeah, Not doesn't work exactly the same as AND and OR in this case.

mattChiaravalloti (approved):

EthanHardyMongo (approved):
LGTM!

PR #187, opened 2026-08-11
SQL-3298: Implement parser support for higher order functions

mattChiaravalloti:
This PR implements parsing for higher order functions according to the [design doc](https://docs.google.com/document/d/15bcFJ1wH6BcQYem7slPP8eG142BSvmjY3UuTNve8r7I/edit?tab=t.0#heading=h.he6cyip700kh). This is the final step that enables higher order functions in the dialect.

In addition to updating the parser, this PR also unskips all remaining higher order function e2e and spec tests, including `error`, `query`, `rewrite`, and `type_constraint` tests. Several of those unskipped tests needed minor alterations since they were incorrectly written initially. The `type_constraint` tests revealed a minor bug in the schema checking module, which was fixed as part of this PR (hence changes in `mir/schema/mod.rs`).

Also, as expected with any `parser` updates, the `pretty_printer` and `pretty_printer_fuzz_test` modules were updated to reflect new parsing behavior.

Finally, the "biggest" change you'll observe in this PR ("biggest" meaning "most files touched") is the addition of `is_delimited` to the `ast::Expression::Identifier` variant, which was necessary to avoid shift-reduce conflicts in the parser for the `FunctionArgument` generation. The issue there was that, syntactically, `NamedFunction::Function`s look exactly the same as `Identifiers`, which obviously causes an ambiguity in the grammer. An example of how this is problematic are these 2 queries:
```
// Query 1
// Here, we refer to the named function, LOG10
SELECT MAP([1, 2, 3], LOG10) FROM [{'LOG10': 'a'}] AS arr

// That query is rewritten into:
SELECT MAP([1, 2, 3], LOG10(this)) FROM [{'LOG10': 'a'}] AS arr

// Query 2
// Here, we refer to a _field_ named "LOG10"
SELECT MAP([1, 2, 3], `LOG10`) FROM [{'LOG10': 'a'}] AS arr
```
Query 1 outputs an array of 3 numbers: `[LOG10(1), LOG10(2), LOG10(3)]`. Query 2 outputs an array of 3 string values, all the same: `['a', 'a', 'a']`. We need to be able to distinguish when an identifier was supposed to be a "named function" vs when it is truly supposed to be an identifier.

Of course, delimiters always indicate that it is an identifier. However, as previously implemented, the parser/ast did not retain information about whether or not a parsed identifier was delimited. That led to the ambiguity I mentioned between `Identifier` expressions vs `NamedFunction::Function`s in the context of higher order functions. `lalrpop` does not tolerate any ambiguities and does not have any knobs or flags to alleviate them -- the only solution is to truly get rid of them. To do this, I created `IdentifierExpr` which tracks `is_delimited`. Using that flag allows us to unambiguously choose if an identifier that happens to use the same name as a function is intended to be a "named function" or an identifier.

This change was far-reaching since there were something like 300 uses of `Expression::Identifier` across many different files. Primarily, these uses are in test files. There were fewer than 10 uses of `Expression::Identifier` in "real" code. I'll highlight in the PR which files were just updated to include `is_delimited` info.

### DIFF
--- a/mongosql/src/algebrizer/test/expressions/between.rs
+++ b/mongosql/src/algebrizer/test/expressions/between.rs
@@ -82,7 +82,7 @@ test_algebrize!(
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"2\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("a".to_string())),
+        max: Box::new(ast::Expression::Identifier("a".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -126,7 +126,7 @@ test_algebrize!(
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"2\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("a".to_string())),
+        max: Box::new(ast::Expression::Identifier("a".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -163,7 +163,7 @@ test_algebrize!(
         arg: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        min: Box::new(ast::Expression::Identifier("a".to_string())),
+        min: Box::new(ast::Expression::Identifier("a".into())),
         max: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"2\"}".to_string()
         )),
@@ -207,7 +207,7 @@ test_algebrize!(
         arg: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        min: Box::new(ast::Expression::Identifier("a".to_string())),
+        min: Box::new(ast::Expression::Identifier("a".into())),
         max: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"2\"}".to_string()
         )),
@@ -244,7 +244,7 @@ test_algebrize!(
         }
     )),
     input = ast::Expression::Between(ast::BetweenExpr {
-        arg: Box::new(ast::Expression::Identifier("a".to_string())),
+        arg: Box::new(ast::Expression::Identifier("a".into())),
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
@@ -288,7 +288,7 @@ test_algebrize!(
         }
     )),
     input = ast::Expression::Between(ast::BetweenExpr {
-        arg: Box::new(ast::Expression::Identifier("a".to_string())),
+        arg: Box::new(ast::Expression::Identifier("a".into())),
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
@@ -335,8 +335,8 @@ test_algebrize!(
         arg: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        min: Box::new(ast::Expression::Identifier("a".to_string())),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        min: Box::new(ast::Expression::Identifier("a".into())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -378,8 +378,8 @@ test_algebrize!(
         arg: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        min: Box::new(ast::Expression::Identifier("a".to_string())),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        min: Box::new(ast::Expression::Identifier("a".into())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -423,8 +423,8 @@ test_algebrize!(
         arg: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        min: Box::new(ast::Expression::Identifier("a".to_string())),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        min: Box::new(ast::Expression::Identifier("a".into())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -463,11 +463,11 @@ test_algebrize!(
         }
     )),
     input = ast::Expression::Between(ast::BetweenExpr {
-        arg: Box::new(ast::Expression::Identifier("a".to_string())),
+        arg: Box::new(ast::Expression::Identifier("a".into())),
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -506,11 +506,11 @@ test_algebrize!(
         }
     )),
     input = ast::Expression::Between(ast::BetweenExpr {
-        arg: Box::new(ast::Expression::Identifier("a".to_string())),
+        arg: Box::new(ast::Expression::Identifier("a".into())),
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
@@ -551,11 +551,11 @@ test_algebrize!(
         }
     )),
     input = ast::Expression::Between(ast::BetweenExpr {
-        arg: Box::new(ast::Expression::Identifier("a".to_string())),
+        arg: Box::new(ast::Expression::Identifier("a".into())),
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"1\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("b".to_string())),
+        max: Box::new(ast::Expression::Identifier("b".into())),
     }),
     env = map! {
         ("foo", 1u16).into() => Schema::Document( Document {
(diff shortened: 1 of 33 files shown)


mongosql/src/algebrizer/test/expressions/between.rs
```diff
@@ -82,7 +82,7 @@ test_algebrize!(
         min: Box::new(ast::Expression::StringConstructor(
             "{\"$numberInt\": \"2\"}".to_string()
         )),
-        max: Box::new(ast::Expression::Identifier("a".to_string())),
+        max: Box::new(ast::Expression::Identifier("a".into())),
```
mattChiaravalloti, line 85: If you're reading this PR in github-file-order, this will be the first file you see. Note that it is a test file and that I switched `"a".to_string()` to `"a".into()`. You'll see in `ast/definitions.rs` that I implemented the following for `IdentifierExpr`:
```
impl From<&str> for IdentifierExpr { ... }
impl From<(&str, bool)> for IdentifierExpr { ... }
impl From<String> for IdentifierExpr { ... }
impl From<(String, bool)> for IdentifierExpr { ... }
```
These are useful throughout the repo, especially in the test cases. If you're using a basic identifier like `"a"` that does not need to be delimited, you can just do `"a".into()` which will create `IdentifierExpr: { name: "a".to_string(), is_delimited: false }`.

If you need to indicate the `is_delimited` field explicitly (which the _vast_ majority of test cases do not), you can do `("<name>", <is_delimited>).into()`. I included the `String` counterparts for places where we already have the name as a `String`.

Anyway, the point is, this change you see on this line is something you will see literally hundreds of times in this PR. The files in `algebrizer/test/*` especially should only have these types of changes.

mongosql/src/algebrizer/test/expressions/higher_order_functions.rs
```diff
@@ -749,7 +749,7 @@ mod reduce {
                 )])),
                 init_value: Box::new(ast::Expression::Literal(ast::Literal::Integer(1))),
                 f: Box::new(ast::FunctionArgument::Expr(ast::Expression::Identifier(
-                    "value".to_string()
+                    "value".into()
```
mattChiaravalloti, line 752: You may notice that here we are not indicating that `"value"` should be delimited. (Recall that `value` is a keyword in MongoSQL.) The reason I chose not to is because this is a algebrizer test and the algebrizer does not consider delimiter information so it is therefore extraneous to set it.

Many of our unit tests use fragmentary, semantically invalid expression and/or stage snippets that focus only on relevant information to assert certain test behavior. This fits into that same idea.

mongosql/src/ast/rewrites/test.rs
```diff
@@ -1946,20 +1946,20 @@ mod higher_order_functions {
...
             ast::HigherOrderFunctionExpr::Reduce(ast::ReduceExpr {
-                array: Box::new(ast::Expression::Identifier("a".to_string())),
+                array: Box::new(ast::Expression::Identifier("a".into())),
                 init_value: Box::new(ast::Expression::Literal(ast::Literal::Integer(0))),
                 f: Box::new(ast::FunctionArgument::Expr(ast::Expression::Binary(
                     ast::BinaryExpr {
                         op: ast::BinaryOp::Add,
-                        left: Box::new(ast::Expression::Identifier("value".to_string())),
-                        right: Box::new(ast::Expression::Identifier("this".to_string())),
+                        left: Box::new(ast::Expression::Identifier(("value", true).into())),
```
mattChiaravalloti, line 1954: Here, in the `rewrite/test` suite, it actually _is_ relevant to use "correct" delimiter information since the rewriter explicitly _does_ indicate `"value"` is a delimited identifier. The tests need to assert that information as well.

mongosql/src/ast/pretty_print_test.rs
```diff
@@ -537,14 +537,23 @@ mod identifier {
...
         expected = "foo1",
         input = "foo1"
     );
     expression_printer_test!(empty, expected = "``", input = "``");
     expression_printer_test!(is_keyword, expected = "`is`", input = "`is`");
+    expression_printer_test!(
+        delimited_non_keyword_retains_delimiters,
+        expected = "`foo`",
+        input = "`foo`"
+    );
```
mattChiaravalloti, line 556: With this change, we will actually now represent user queries exactly as they were provided to us. Previously, if a user wrote:
```
SELECT `foo` FROM `bar`
```
We would pretty print this back to them as
```
SELECT foo FROM bar
```

mongosql/src/ast/pretty_print_test.rs
```diff
@@ -1255,43 +1264,42 @@ mod precedence_tests {
...
+
+    expression_printer_test!(
+        filter_with_expr_arg,
+        expected = "FILTER(a, this = 1)",
+        input = "FILTER(a, this=1)"
+    );
+
+    expression_printer_test!(
+        filter_with_named_function_arg,
+        expected = "FILTER(a, NOT )", // we intentionally print a space after unary ops
```
mattChiaravalloti: I don't like it either, but I don't think we should change this as part of this PR. It's not as straightforward as I would hope :(

mongosql/src/mir/schema/mod.rs
```diff
@@ -2263,6 +2263,37 @@ trait HigherOrderFunction: SqlFunction {
             error: Box::new(error),
         }
     }
+
+    /// Validate that the given array expression satisfies ANY_ARRAY_OR_NULLISH. Return the array
+    /// schema and return its item schema.
+    fn validate_array_and_get_item_schema(
```
mattChiaravalloti, line 2269: This is helper method is used below by all 3 higher order function variants. They all duplicated this code, so I factored it out to make this. Now, instead of returning `Null` as the schema of a higher order function when an array input is definitely not an array, we instead proceed with algebrization (treating the `this` schema as missing).

This is important for correctness and was caught by the type_constraint tests. Consider the expression: `FILTER(NULL, "bad")`. That expression is invalid since the second argument _must_ satisfy bool, null, or missing, but it is a string! Without this change, we'd accept this expression and say it's schema is null, but actually we should fail to compile the query with a schema checking error.

Interestingly, MongoDB can tolerate invalid types in this situation, however MongoSQL is meant to fail on type constraint validations even if an expression would otherwise evaluate validly at runtime.

mongosql/src/mir/schema/mod.rs
```diff
@@ -2283,26 +2314,8 @@ impl SqlFunction for MapExpr {
...
-        }
-
-        let Some(array_item_schema) = array_schema.get_array_item_schema() else {
-            // At this point, we know the schema satisfies ANY_ARRAY_OR_NULLISH. In STRICT
-            // mode, that means the schema must be exactly NULLISH if there is no array item
-            // schema. In RELAXED mode, technically there may be other non-array/non-nullish
-            // schemas, but those would cause runtime errors. Therefore, we can just return
-            // Null as the expression schema at this point since we know the `array` arg is
-            // never an array.
-            return Ok(Schema::Atomic(Atomic::Null));
```
mattChiaravalloti, line 2304: See here, for example, we used to return `Null`. As you can see above in the helper, we now set the `array_item_schema` to `Missing` if we encounter this situation.

mongosql/src/parser/mongosql.lalrpop
```diff
@@ -402,6 +402,55 @@ Substring: FunctionExpr = {
...
+    },
+}
+
+FunctionArgument: FunctionArgument = {
+    ADD => {
+        FunctionArgument::NamedFunction(NamedFunction::UnaryOp(UnaryOp::Pos))
+    },
+    SUB => {
+        FunctionArgument::NamedFunction(NamedFunction::UnaryOp(UnaryOp::Neg))
+    },
```
mattChiaravalloti, line 433: Intentionally choosing to interpret `+` and `-` as the Unary versions instead of the Binary versions, as noted in the design doc. For now, these are the only overloaded operators in the dialect.

mongosql/src/parser/mongosql.lalrpop
```diff
@@ -402,6 +402,55 @@ Substring: FunctionExpr = {
...
+        FunctionArgument::NamedFunction(NamedFunction::UnaryOp(UnaryOp::Not))
+    },
+    <b:NonOverloadedBinaryOp> => {
+        FunctionArgument::NamedFunction(NamedFunction::BinaryOp(b))
+    },
+    <e:Expression> => {
+        // Check if this is a bare identifier that could be a function name
+        match &e {
+            Expression::Identifier(IdentifierExpr { name: id, is_delimited: false }) => {
+                match FunctionName::try_from(id.as_str()) {
```
mattChiaravalloti, line 444: Here it is: this is where we use `is_delimited: false` to determine if we should actually attempt to consider an identifier as a named function!

mongosql/src/parser/test.rs
mattChiaravalloti, line 1: Although the first several lines of this file are just `IdentiferExpr`-related changes, at the bottom you will find the `higher_order_functions` module with all new tests covering the new expected parser behavior for higher order functions.

testdata/spec_tests/query_tests/higher_order_functions.yml
```diff
@@ -9,21 +9,21 @@ dataset:
...
-            bsonType: "int"
-          a:
-            anyOf:
-              - bsonType: "array"
-                items:
-                  anyOf:
-                    - bsonType: "int"
-                    - bsonType: !!str "null"
-              - bsonType: !!str "null"
+    schema:
```
mattChiaravalloti, line 12: I set up the spacing incorrectly when I first created this file, so the schema wasn't being set properly by the data-loader. This change fixes that.

tests/errors/rewriter.yml
```diff
@@ -0,0 +1,6 @@
+tests:
+  - description: IncorrectArgumentCount with Minimum requirement in Higher Order Function
```
mattChiaravalloti, line 2: Moved from `tests/errors/schema.yml`. This now actually always caught as a rewriter issue and will not make it to schema-checking.

mongosql/src/ast/pretty_print_test.rs
jpowell-mongo: Q: What was the context here?

mattChiaravalloti: I believe it's because we want to leave no space for the unary `+` and `-` operators, but we want a space for `NOT`, so we define pretty printing for `UnaryExpr` as the pretty print of the `UnaryOp` followed directly by its pretty printed argument, letting the `UnaryOp` themselves define how they're written (space or no space).

I'm realizing now I can probably `trim` the on the `FunctionArgument` pretty print!

tests/errors/rewriter.yml
EthanHardyMongo: So does `Error 1001` now only apply to functions that require an exact number of arguments? Should we update this in the `errors.md` and add another section for rewriter errors?

EthanHardyMongo (approved):
LGTM! I just had one comment about updating the `errors.md` file.

jpowell-mongo (approved):

tests/errors/rewriter.yml
mattChiaravalloti: > So does `Error 1001` now only apply to functions that require an exact number of arguments?

Error 1001 still applies at schema checking to any expression with an incorrect number of arguments. Some functions expect a minimum number and some expect an exact number. This test was initially added as part of #162 to account for this error case in the context of higher order functions because I thought that was the only way to trigger it. At that time, I did not realize we would inevitably need to address nearly all minimum requirement operators at either parse or syntactic rewrite time. Now, though, I realize we actually can test this error directly via `COALESCE()` since we require a minimum of 1 argument to `COALESCE`.

I'll update this file to include the `COALESCE()` test so minimum arg errors in the schema module are covered here.

I'll also update this file to include a test for a higher order function wrapper error, since I did not do that at the time of #173.

> Should we update this in the `errors.md` and add another section for rewriter errors?

As for this question, I have two things to say:
1. Yes, we should eventually update this to support rewrite errors, but first we should update the rewrite error type to implement `UserError`. That is tracked here: [SQL-3220](https://jira.mongodb.org/browse/SQL-3220). For this PR, I will leave this as-is.
2. I will update `errors.md` to account for the higher order function wrapper error since I forgot to do that as part of #173. I'll be sure to note in the docs ticket related to the next release that the errors page needs to be re-rendered based on changes from this project.

PR #189, opened 2026-08-13
[SQL-3349] Merge adjacent match filter stages

jpowell-mongo:
## Summary

Extends the`MergeNeighboringMatchesOptimizer` to merge adjacent Mql MatchFilter stages. Adds parity to our existing logic for regular Filter stages. 

Today, a query like:

```sql
 SELECT * FROM movies WHERE year > 1950 AND num_mflix_comments > 20
```

Creates an aggregation pipeline with two match stages. One for each side of the AND expression:
```javascript
[
    { "$match": { "year": { "$gt": 1950 } } },
    { "$match": { "num_mflix_comments": { "$gt": 20 } } },
    { "$project": { "movies": "$$ROOT", "_id": 0 } },
]
```

There are potential performance gains by merging the two match stages so we don't need to perform two collection scans at each match stage. These changes now generate:
```javascript
[
    { "$match": { "$and": [{ "year": { "$gt": 1950 } }, { "num_mflix_comments": { "$gt": 20 } }] } },
    { "$project": { "movies": "$$ROOT", "_id": 0 } },
]
```

Where we create a single match stage that uses the $and operator to combine the conjoin the conditions, just as we do for regular filter stages.

### Note
It appears that this optimization may already exist on the server based on their docs:
 https://www.mongodb.com/docs/manual/core/aggregation-pipeline-optimization/#match--match-coalescence
 
 However, when running the non-coalesced aggregation pipeline, the explain plan doesn't seem to optimize the query:
 ```javascript
[
    { "$match": { "year": { "$gt": 1950 } } },
    { "$match": { "num_mflix_comments": { "$gt": 20 } } },
    { "$project": { "movies": "$$ROOT", "_id": 0 } },
]
```

But I don't know if indexes or other factors impact this.
 
 I think this makes sense to add from a completeness perspective, we do the same optimization for regular filters, so there's not reason not to include here, but I suspect this will get taken care of for us.

### DIFF
--- a/mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
+++ b/mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
@@ -78,12 +78,11 @@ impl Visitor for MergeNeighboringMatchesVisitor {
                     // instead of nesting a new $and inside it
                     MatchQuery::Logical(MatchLanguageLogical {
                         op: MatchLanguageLogicalOp::And,
-                        args,
+                        mut args,
                         ..
                     }) => {
-                        let mut conditions = args;
-                        conditions.push(this_condition);
-                        conditions
+                        args.push(this_condition);
+                        args
                     }
                     // otherwise, create a vector with the child condition, and the condition of this filter
                     combined_conditions => vec![combined_conditions, this_condition],
--- a/mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
+++ b/mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
@@ -57,42 +57,53 @@ impl MergeNeighboringMatchesVisitor {
 }
 
 impl Visitor for MergeNeighboringMatchesVisitor {
-    fn visit_mql_stage(&mut self, node: MqlStage) -> MqlStage {
-        match node {
-            MqlStage::MatchFilter(match_filter) => {
-                let match_filter = self.visit_match_filter(*match_filter);
-                match match_filter.source.as_ref() {
-                    Stage::MqlIntrinsic(MqlStage::MatchFilter(child)) => {
-                        let conditions = match &child.condition {
-                            // the child is already a $and, append this condition to that and
-                            // instead of nesting a new $and inside it
-                            MatchQuery::Logical(MatchLanguageLogical {
-                                op: MatchLanguageLogicalOp::And,
-                                args,
-                                ..
-                            }) => {
-                                let mut conditions = args.clone();
-                                conditions.push(match_filter.condition.clone());
-                                conditions
-                            }
-                            // otherwise, create a $and with the two match filter conditions
-                            _ => vec![child.condition.clone(), match_filter.condition.clone()],
-                        };
-                        MqlStage::MatchFilter(Box::new(MatchFilter {
-                            source: child.source.clone(),
-                            condition: MatchQuery::Logical(MatchLanguageLogical {
-                                op: MatchLanguageLogicalOp::And,
-                                args: conditions,
-                                cache: Default::default(),
-                            }),
-                            cache: match_filter.cache.clone(),
-                        }))
+    fn visit_match_filter(&mut self, node: MatchFilter) -> MatchFilter {
+        // Recurse first so that we've already merged children before we try to merge this node.
+        let MatchFilter {
+            source,
+            condition,
+            cache,
+        } = node.walk(self);
+        match *source {
+            Stage::MqlIntrinsic(MqlStage::MatchFilter(child)) => {
+                let MatchFilter {
+                    source: child_source,
+                    condition: child_condition,
+                    ..
+                } = *child;
+
+                // Combine the child condition with the parent condition
+                let conditions: Vec<MatchQuery> = match child_condition {
+                    // the child is already a $and, append this condition to that and
+                    // instead of nesting a new $and inside it
+                    MatchQuery::Logical(MatchLanguageLogical {
+                        op: MatchLanguageLogicalOp::And,
+                        args,
+                        ..
+                    }) => {
+                        let mut conditions = args;
+                        conditions.push(condition);
+                        conditions
                     }
-                    // nothing to merge, return the filter as-is
-                    _ => MqlStage::MatchFilter(Box::new(match_filter)),
+                    // otherwise, create a vector with the child condition, and the condition of this filter
+                    child_match_query => vec![child_match_query, condition],
+                };
+                MatchFilter {
+                    source: child_source,
+                    condition: MatchQuery::Logical(MatchLanguageLogical {
+                        op: MatchLanguageLogicalOp::And,
+                        args: conditions,
+                        cache: Default::default(),
+                    }),
+                    cache,
                 }
             }
-            _ => node.walk(self),
+            // nothing to merge; rebuild the filter from its own (unchanged) fields
+            other => MatchFilter {
+                source: Box::new(other),
+                condition,
+                cache,
+            },
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs [hunk 1]
```diff
@@ -56,6 +57,45 @@ impl MergeNeighboringMatchesVisitor {
 }
 
 impl Visitor for MergeNeighboringMatchesVisitor {
+    fn visit_mql_stage(&mut self, node: MqlStage) -> MqlStage {
```
jpowell-mongo: Note: 

I wasn't sure if I should have this on visit_mql_stage, or actually just implement the visit_match_filter method instead? 

The mql_stage has a few variants so I felt like it made sense to match directly on the MatchFilter type. It's a small change, but lmk what you all prefer.

mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
```diff
@@ -56,6 +57,56 @@ impl MergeNeighboringMatchesVisitor {
...
+                    }
+                    // otherwise, create a vector with the child condition, and the condition of this filter
+                    child_match_query => vec![child_match_query, condition],
+                };
+                MatchFilter {
+                    source: child_source,
+                    condition: MatchQuery::Logical(MatchLanguageLogical {
+                        op: MatchLanguageLogicalOp::And,
+                        args: conditions,
+                        cache: Default::default(),
```
jpowell-mongo, line 95: Note: I wasn't sure what to put here for the Cache, is it okay that I set the cache for the inner And condition to be Default::default() since I re-use the 'cache' field for the outer match filter?

mongosql/src/mir/optimizer/merge_neighboring_matches/test.rs
```diff
@@ -76,6 +77,244 @@ mod merge_neighboring_matches_tests {
...
+                function: MatchLanguageComparisonOp::Eq,
+                input: Some(mir_field_path("bar", vec!["y"])),
+                arg: LiteralValue::Integer(43),
+                cache: SchemaCache::new(),
+            }),
+            cache: SchemaCache::new(),
+        }))),
+    );
+
+    test_merge_neighboring_matches!(
```
mattChiaravalloti: [suggestion] While we're working on this file, let's create a `test_merge_neighboring_matches_noop!` macro, similar to other optimization test modules, so that this kind of test is easier to write/read. After adding that macro (follow the pattern of other `*_noop!` macros: [example](https://github.com/mongodb/mongosql/blob/1372410213e3f77d2316af2897cc58361802e1ff/mongosql/src/mir/optimizer/constant_folding/test.rs#L46-L50), [example](https://github.com/mongodb/mongosql/blob/1372410213e3f77d2316af2897cc58361802e1ff/mongosql/src/mir/optimizer/dead_code_elimination/test.rs#L55-L59), ...), you can update this test to use it instead, just providing an input.

mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
```diff
@@ -56,6 +57,56 @@ impl MergeNeighboringMatchesVisitor {
...
+        // Recurse first so that we've already merged children before we try to merge this node.
+        let MatchFilter {
+            source,
+            condition,
+            cache,
+        } = node.walk(self);
+        match *source {
+            Stage::MqlIntrinsic(MqlStage::MatchFilter(child)) => {
+                let MatchFilter {
+                    source: child_source,
```
mattChiaravalloti: [super-nit] (super-non-blocking) (super optional, truly!) I always considered the `source` relationship to be more "parent"-like than "child"-like since it is the "source" of this stage's inputs.

Alternatively, to avoid the question entirely, we may consider naming the `node`'s fields something like `this_source` and `this_condition` or `node_source` and `node_condition`, and `node.source`'s fields like `other_source` and `other_condition` or `source_source` and `source_condition`. Or, similar to how we do it in `visit_filter` for standard `Filter` stages, we could skip deconstruction and always refer to the fields via namespacing `node.condition` vs `f.condition` (though, that `f` could do with a rename to something like `source` or `source_f`). Anyway, "naming is hard", etc. Wanted to share some thoughts and ideas!

It's totally fine to also just leave this as-is!

mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
mattChiaravalloti: TLDR: this is correct at the `Expression`/`MatchQuery` level.

<hr>

Oh wow I didn't realize we still had `cache` fields on `mir::MatchQuery` structs. We removed `cache` from all `mir::Expression` variants because it proved too difficult to keep them valid and in sync. We may want to file a quick-win ticket to remove `cache` from `MatchQuery` structs, but that's certainly out of scope of this ticket.

Anyway: It's always correct to use `Default::default()` for a `SchemaCache` since it can always be recalculated. Sure, that is sometimes inefficient, but it is sometimes necessary. I'd say this is a case where it is necessary since we don't have any existing primitives for combining caches (and it is out of scope of this ticket to introduce such primitives). We probably _could_ work out some way of combining caches so we could avoid re-computation, but it is always possible that there are some expression combinations that would meaningfully change the schema so the combination functions could end up being equivalent in terms of work to re-computation.

mattChiaravalloti: Btw, the reason we're able to use the `node`'s `cache` for the combined stage is because `node` comes _after_ the `source` so its cache already includes information from the source. As in, if the source says field `a` is not nullable and then this node says field `b` is not nullable, then this node's cache already included info about `a` being non-null. (Consider the alternative where we use `source.cache` instead of `node.cache` -- well, then, the combined stage would still say `a` is non-null, but it would _not_ say `b` is non-null even though the combined stage ensure that!)

At the expression-level, though, that's not always the case. It _might_ be safe for `MatchQuery` combining since they all need to evaluate to boolean or nullish, but still it's not worth the "risk" since recalculation is not so difficult.

mattChiaravalloti (approved):
lgtm 💯 One optional suggestion about `noop` testing, and one SUPER optional comment about naming. No need to retag unless you think it's necessary 👍

mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
```diff
@@ -56,6 +57,56 @@ impl MergeNeighboringMatchesVisitor {
...
+                    // instead of nesting a new $and inside it
+                    MatchQuery::Logical(MatchLanguageLogical {
+                        op: MatchLanguageLogicalOp::And,
+                        args,
+                        ..
+                    }) => {
+                        let mut conditions = args;
+                        conditions.push(this_condition);
+                        conditions
+                    }
```
nicholascioli, line 86: You can just mark `args` as `mut`!


```suggestion
                    MatchQuery::Logical(MatchLanguageLogical {
                        op: MatchLanguageLogicalOp::And,
                        mut args,
                        ..
                    }) => {
                        args.push(this_condition);
                        args
                    }
```

nicholascioli (approved):
LGTM! Left a small nit about `mut`.

mongosql/src/mir/optimizer/merge_neighboring_matches/mod.rs
jpowell-mongo: Ahhhh, this is nice. Every time I get a review from you, it's like "you know Rust just lets you do xyz"

PR #191, opened 2026-08-21
SQL-3432: add marker verification as a separate crate

nicholascioli:
This PR attempts to standardize the way mongosql clients enforce token entitlement.


mongosql-token-verifier/src/providers.rs
```diff
@@ -0,0 +1,277 @@
...
+            if url != JWKS_WELL_KNOWN_URL {
+                return Ok(json!({}));
+            }
+
+            self.0.clone().ok_or(TestError::IntentionallyEmpty)
+        }
+    }
+
+    fn generate_keys(seed: u8, key_id: &str) -> (SigningKey, Jwk) {
+        let signing_key = SigningKey::from_bytes(&[seed; 32]);
```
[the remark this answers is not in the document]
nicholascioli: So this crate is only used for tests, so I think this is fine to ignore?

jchemburkar:
@mattChiaravalloti tagging you in since I took over and made a few small changes. This pr is for verifying the token to ensure the clusters have access to atlas sql. The basic workflow:
1. the driver will access the underlying collection to pull the token
2. here we pull the public jwks from a trusted url and cache it (this will mean for drivers holding a long connection we dont spam the trusted url)
3. we verify the token (passed as a string) to make sure it is valid

the most relevant design doc is here: https://github.com/10gen/engineering-documents/blob/main/teams/sql-engines/SQL-3026-atlas-sql-dedicated-cluster/AtlasSQLDirectCluster_marker.md

happy to sync monday if that would be of any help

mongosql-token-verifier/src/providers.rs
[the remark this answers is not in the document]
mattChiaravalloti: Yes if it is test-only then we can ignore it.

mongosql-token-verifier/src/clients/file.rs
```diff
@@ -0,0 +1,12 @@
+use std::{future::Future, path::Path};
+
+/// A file client
```
mattChiaravalloti, line 3:
```suggestion
/// A file client. Intended for use by offline clients.
```
[suggestion] I know it should be obvious from context given the rest of this package, but I still think we may benefit from including "why" this trait exists in this file. From my reading of the design doc and the rest of this PR, it seems this is the solution that lines up with
> Client applications in environments with no outbound internet access need a fallback mechanism. We will add a new URI/CLI option to provide a path to a the local jwks.json file.

Do you think it's worth saying something to that affect here?

mongosql-token-verifier/src/providers.rs
```diff
@@ -0,0 +1,403 @@
...
+            *atime -= JWKS_TTL;
+        }
+    }
+}
+
+impl<H: HttpsClient, F: FileClient> NetworkedCachedJwksProvider<H, F> {
+    /// Refresh the JWKS, ignoring any cached entry: upstream, then file, then
+    /// falling back to the last-known-good cache on failure
+    async fn refresh(&mut self) -> Result<JwkSet, CachedError<H::Error, F::Error>> {
+        // Always attempt to fetch from upstream first
```
mattChiaravalloti, line 115: [question] Why are we always attempting to fetch from the upstream first when we anticipate there will be offline use cases? Do we expect users to always provide a fallback file, or will that file only be provided when the user expects to be offline? If its the latter, then maybe we should use the presence of `self.path` to indicate we should read from the file first?

What kind of timeout do we expect have set on the `fetch_from_upstream` codepath? Should we document in the library that a strict timeout should be set if a consumer (driver) expects to use this exclusively in offline mode? I suppose they can just use a dummy `HttpsClient` implementation in that case that immediately errors. Still, it would be nice to have the library skip `HttpsClient` fetching entirely in the offline case.

mongosql-token-verifier/src/providers.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// Backdate the cached entry so it reads as expired, exercising the refresh path
+    #[cfg(test)]
+    fn expire_cache(&mut self) {
+        if let Some((_, atime)) = &mut self.jwks {
+            *atime -= JWKS_TTL;
+        }
+    }
+}
+
+impl<H: HttpsClient, F: FileClient> NetworkedCachedJwksProvider<H, F> {
```
mattChiaravalloti, line 111: [question] Why is this a separate `impl` block from the one above? It seems both are exactly `impl<H: HttpsClient, F: FileClient> NetworkedCachedJwksProvider<H, F>`, so do they need to be separate? If so, can we include a code-comment explaining why?

mongosql-token-verifier/src/validator.rs
```diff
@@ -0,0 +1,565 @@
...
+            result
+        };
+
+        let decoded = jsonwebtoken::decode::<Claims>(marker, &key, &validation)?;
+        if !decoded.claims.enabled {
+            return Err(ValidatorError::MarkerDisabled);
+        }
+
+        // Validate that the marker is for the specified cluster
+        if decoded.claims.sub != self.cluster {
```
mattChiaravalloti, line 87: [suggestion/question] Shouldn't this check happen before the `!decoded.claims.enabled` check? If we somehow end up in the state where an invalid token is set (because of the wrong cluster) and also happens to have the marker disabled, it feels like the "wrong cluster" error is more important than the "disabled" error. It indicates to the user that something more important is incorrect than simply being disabled.

mongosql-token-verifier/src/validator.rs
```diff
@@ -0,0 +1,565 @@
...
+
+        // Also ensure that the marker is not expired
+        // Note: It would be nice if the library could be dynamically do this for us,
+        // but we are kind of stuck in a catch-22 because we don't know the claims
+        // until we have decoded them but we can't decode them without knowing
+        // if `exp` should be there or not :(
+        if decoded.claims.iss == Issuer::Emergency && decoded.claims.exp.is_none() {
+            return Err(ValidatorError::MissingExp);
+        }
+        if let Some(exp) = decoded.claims.exp {
```
mattChiaravalloti, line 102: [suggestion/question] Same goes for these error cases: we should probably enforce these before considering "is the feature enabled?"

mongosql-token-verifier/src/validator.rs
```diff
@@ -0,0 +1,565 @@
...
+                return Err(ValidatorError::MarkerExpired);
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
```
mattChiaravalloti, line 114: [chore] Looks like we're missing a test for when the token is emergency-issued and missing an `exp`. This should assert the `ValidtorError::MissingExp` error.

mattChiaravalloti (changes requested):
This is super impressive! I have some thoughts, questions, and suggestions, but nothing that requires too many changes. This is looking nearly ready 💯

nicholascioli:
Will close for now; we can re-open when we revisit this work.

PR #194, opened 2026-08-26
SQL-3387: Implement support for ARRAY_REMOVE with proper semantics

mattChiaravalloti:
This PR concludes the array functions work for the Dialect Improvement project. It implements support for `ARRAY_REMOVE` with the intended semantics; that is, it supports `ARRAY_REMOVE` such that null elements are not removed from the array unless the specified value to remove is itself null. See the spec/query tests for details.

The ticket proposed indicating the `FilterExpr` include a flag to force MQL semantics (as opposed to SQL 3-value null semantics) but I ultimately decided to push that lower, to the expression nested within the `FilterExpr`. The reason I did this is because it is really the `Neq` expression _within_ the rewrite that needs to force MQL semantics, not the `FilterExpr` itself. That flag would be too powerful and ambiguous if it was available at the `FilterExpr` level.

I documented at the `ast` and `mir` levels why the flag exists so that future implementors could avoid misusing it.

Apologies for the size of the PR. It is mostly a consequence of adding a new flag to two pretty popular structs. For the most part, the only relevant files are `parsers/mongosql.lalrpop`, `ast/definitions.rs`, `ast/rewrites/higher_order_functions.rs`,  `algebrizer/definitions.rs`, `mir/definitions.rs`, `translator/{expression.rs, util.rs}`, and `tests/spec_tests/query_tests/array_functions.yml`. All other files are incidental updates that simply introduce the new flag with the default value (false). I did add unit testing for relevant functions, if you are interested in those.


mongosql/src/algebrizer/test/expressions/between.rs
mattChiaravalloti, line 1: If you are looking in GitHub order, this is the first file you'll see. This is pretty much what you'll see in all test files! The flag is introduced with the false value wherever necessary. In some files my IDE also made local edits removing extraneous commas and newlines. It looks like none of those types of edits are in this file but that's the only other type of change you'll see in test files.

mongosql/src/algebrizer/test/expressions/binary.rs
mattChiaravalloti, line 1: Thile includes a `mod force_mql_semantics` at the bottom with new unit test coverage.

mongosql/src/algebrizer/definitions.rs
```diff
@@ -2071,14 +2074,16 @@ impl<'a> Algebrizer<'a> {
...
+        let force_mql_semantics = b.force_mql_semantics;
 
         // here we don't use the new constructor because we're setting the
         // calculated nullability
         let scalar_function_expr =
             mir::Expression::ScalarFunction(mir::ScalarFunctionApplication {
                 function,
-                is_nullable,
                 args,
+                force_mql_semantics,
```
mattChiaravalloti, line 2085: The only `ast` struct that defines `force_mql_semantics` is `ast::BinaryExpr`. That is why this is the only spot where we set the value. All other locations use `false`.

mongosql/src/mir/optimizer/constant_folding/lib.rs
```diff
@@ -335,21 +337,27 @@ impl ConstantFoldExprVisitor<'_> {
...
         let new_sf = Expression::ScalarFunction(ScalarFunctionApplication {
             function: ScalarFunction::And,
-            is_nullable: sf.is_nullable,
             args: vec![
                 Expression::ScalarFunction(ScalarFunctionApplication {
                     function: ScalarFunction::Lte,
-                    is_nullable: sf.is_nullable,
                     args: vec![arg.clone(), top],
+                    force_mql_semantics: sf.force_mql_semantics,
+                    is_nullable: arg_is_nullable || top_is_nullable,
```
mattChiaravalloti, line 350: 🚗  drive-by minor correctness improvement here. It's actually inconsequential given the way this entire module is implemented but it's better to be correct here in case other changes are made around this later.

mongosql/src/translator/test/expressions.rs
mattChiaravalloti, line 1: This file contains unit tests for relevant functions.

jpowell-mongo (commented):
I did some investigation, and I'm wondering if we could translate the ARRAY_REMOVE to include a null-check if we're trying to remove a literal value. This would mean the changes only need to happen in rewrite_array_remove and we don't need to proliferate a boolean argument across the rest of the codebase

I made a POC branch here:
https://github.com/mongodb/mongosql/pull/195/changes

```rust
   /// Rewrite `ARRAY_REMOVE(a, NULL)` into `FILTER(a, this <> NULL)`.
    /// For any other `x`, null elements must be retained, so rewrite
    /// `ARRAY_REMOVE(a, x)` into `FILTER(a, this IS NULL OR this <> x)`.
    fn rewrite_array_remove(args: &[Expression]) -> Result<Expression> {
        let [array, remove_expr] = try_exact_args("ARRAY_REMOVE", args)?;

        let is_filtering_out_nulls = matches!(remove_expr, Expression::Literal(Literal::Null));
        if is_filtering_out_nulls {
            // If we're explicitly checking for null, then this is fine.
            Ok(Self::make_filter(
                array.clone(),
                Self::make_binary(
                    this(),
                    BinaryOp::Comparison(ComparisonOp::Neq),
                    remove_expr.clone(),
                ),
            ))
        } else {
            // Not explicitly removing nulls, so null elements must be retained:
            // FILTER(a, this IS NULL OR this <> remove_expr)
            let include_null_values = Expression::Is(IsExpr {
                expr: Box::new(this()),
                target_type: TypeOrMissing::Type(Type::Null),
            });

            Ok(Self::make_filter(
                array.clone(),
                Self::make_binary(
                    include_null_values,
                    BinaryOp::Or,
                    Self::make_binary(
                        this(),
                        BinaryOp::Comparison(ComparisonOp::Neq),
                        remove_expr.clone(),
                    ),
                ),
            ))
        }
    }
```

mattChiaravalloti:
Closing in favor of #195

PR #196, opened 2026-08-28
[SQL-3387] Support Array Remove with a null check

jpowell-mongo:
## Summary

This pull request updates the rewrite_array_remove() function to add null checks against the remove_expr, and the element being compared 'this', so that we can ensure we keep null elements if null is not explicitly being filtered out.

### DIFF
--- a/mongosql/src/ast/rewrites/higher_order_functions.rs
+++ b/mongosql/src/ast/rewrites/higher_order_functions.rs
@@ -225,48 +225,36 @@ impl HigherOrderFunctionsAliasVisitor {
             })
         };
 
-        let is_filtering_out_nulls = matches!(remove_expr, Expression::Literal(Literal::Null));
-        if is_filtering_out_nulls {
-            Ok(Self::make_filter(
-                array.clone(),
+        // NOT (this IS NULL AND x IS NULL): drop the element when both sides are NULL.
+        let not_both_null = Expression::Unary(UnaryExpr {
+            op: UnaryOp::Not,
+            expr: Box::new(Self::make_binary(
+                is_null(this()),
+                BinaryOp::And,
+                is_null(remove_expr.clone()),
+            )),
+        });
+
+        // this IS NULL OR x IS NULL OR this <> x: keep the element when exactly one side is
+        // NULL, and otherwise defer to `<>` on two non-NULL operands.
+        let either_null_or_unequal = Self::make_binary(
+            is_null(this()),
+            BinaryOp::Or,
+            Self::make_binary(
+                is_null(remove_expr.clone()),
+                BinaryOp::Or,
                 Self::make_binary(
                     this(),
                     BinaryOp::Comparison(ComparisonOp::Neq),
                     remove_expr.clone(),
                 ),
-            ))
-        } else {
-            // NOT (this IS NULL AND x IS NULL): drop the element when both sides are NULL.
-            let not_both_null = Expression::Unary(UnaryExpr {
-                op: UnaryOp::Not,
-                expr: Box::new(Self::make_binary(
-                    is_null(this()),
-                    BinaryOp::And,
-                    is_null(remove_expr.clone()),
-                )),
-            });
+            ),
+        );
 
-            // this IS NULL OR x IS NULL OR this <> x: keep the element when exactly one side is
-            // NULL, and otherwise defer to `<>` on two non-NULL operands.
-            let either_null_or_unequal = Self::make_binary(
-                is_null(this()),
-                BinaryOp::Or,
-                Self::make_binary(
-                    is_null(remove_expr.clone()),
-                    BinaryOp::Or,
-                    Self::make_binary(
-                        this(),
-                        BinaryOp::Comparison(ComparisonOp::Neq),
-                        remove_expr.clone(),
-                    ),
-                ),
-            );
-
-            Ok(Self::make_filter(
-                array.clone(),
-                Self::make_binary(not_both_null, BinaryOp::And, either_null_or_unequal),
-            ))
-        }
+        Ok(Self::make_filter(
+            array.clone(),
+            Self::make_binary(not_both_null, BinaryOp::And, either_null_or_unequal),
+        ))
     }
 
     /// Rewrite `ARRAY_COUNT_IF(a, f)` into `SIZE(FILTER(a, f))`.
--- a/mongosql/src/ast/rewrites/higher_order_functions.rs
+++ b/mongosql/src/ast/rewrites/higher_order_functions.rs
@@ -200,7 +200,7 @@ impl HigherOrderFunctionsAliasVisitor {
 
         let is_filtering_out_nulls = matches!(remove_expr, Expression::Literal(Literal::Null));
         if is_filtering_out_nulls {
-            // If we're explicitly checking for null, then this is fine.
+            // If we're explicitly filtering out null, we can just use (x <> remove_expr)
             Ok(Self::make_filter(
                 array.clone(),
                 Self::make_binary(
@@ -210,13 +210,13 @@ impl HigherOrderFunctionsAliasVisitor {
                 ),
             ))
         } else {
-            // Not explicitly removing nulls, so null elements must be retained:
-            // FILTER(a, this IS NULL OR this <> remove_expr)
+            // If we're filtering out a literal, include an IS null checks, so we keep them.
             let include_null_values = Expression::Is(IsExpr {
                 expr: Box::new(this()),
                 target_type: TypeOrMissing::Type(Type::Null),
             });
 
+            // `FILTER(a, this IS NULL OR x <> remove_expr)
             Ok(Self::make_filter(
                 array.clone(),
                 Self::make_binary(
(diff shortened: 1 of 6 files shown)


mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -192,18 +192,81 @@ impl HigherOrderFunctionsAliasVisitor {
         ))
     }
 
-    /// Rewrite `ARRAY_REMOVE(a, x)` into `FILTER(a, this <> x)`.
+    /// Rewrite `ARRAY_REMOVE(a, NULL)` into `FILTER(a, this <> NULL)`.
+    /// For any other `x`, rewrite `ARRAY_REMOVE(a, x)` into
+    /// `FILTER(a, NOT (this IS NULL AND x IS NULL) AND (this IS NULL OR x IS NULL OR this <> x))`.
```
mattChiaravalloti: [issue] (**blocking**) We should not treat the `NULL` literal as a special case since it relies on brittle and arguably buggy/incomplete code later in the translation pipeline.

In isolation, `this <> NULL` is _always_ `NULL` via SQL 3-value semantics. Since `NULL` is falsy that result should cause all elements to be filtered out. This is the problem I was originally addressing in the closed PR (#194): creating a special marker to indicate an operator should force MQL semantics instead of SQL semantics regardless of operand nullability.

The problem with this special case is that it relies on what could either be considered a bug or at least be considered incomplete code in several places.

First, in general, the `constant_folding` optimizer folds a binary comparison operation that has a definitively nullish argument into the `NULL` literal ([link](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/mongosql/src/mir/optimizer/constant_folding/lib.rs#L777-L780)). This only fails for binary comparisons within higher order functions that put a `Variable` argument "first" (e.g., `FILTER(a, this > NULL)` will fail to fold, but `FILTER(a, NULL < this)` will successfully fold). So since our special-cased rewrite is `FILTER(a, this <> NULL)`, the optimizer fails. Future updates to the optimizer to properly handle this would be _correct_ to fold this into just `NULL`, invalidating this rewrite.

Next, the `sql_null_semantic_operators` desugarer relies on that optimizer behavior which higher order functions currently break ([link](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/mongosql/src/air/desugarer/sql_null_semantics_operators.rs#L327-L328)). Again, a fix to this in the future would be correct and consistent with the rest of the codebase but would invalidate this special-cased rewrite.

**My conclusions is: Regardless of if the argument is a literal `NULL` or an expression that evaluates to `NULL`, we should always rewrite this the same way.** Otherwise, we run the risk of this rewrite becoming invalid as we refine and correct the rest of the translation codepath over time.

(Sidenote: 
I'm going to file a ticket to fix the optimizer for handling variables since right now we can trivially produce inconsistent results. My examples above demonstrate this:
```
// query
SELECT VALUE { 'a': a, 'v': FILTER(a, this > null) } FROM [{'a': []}, {'a': [1, 2, 3]}, {'a': [1, null, 3]}, {'a': null}, {}] AS arr`

// result
    { "": { "a": [], "v": [] } }
    { "": { "a": [1, 2, 3], "v": [1, 2, 3] } }
    { "": { "a": [1, null, 3], "v": [1, 3] } }
    { "": { "a": null, "v": null } }
    { "": { "v": null } }

// logically equivalent query
SELECT VALUE { 'a': a, 'v': FILTER(a, null < this) } FROM [{'a': []}, {'a': [1, 2, 3]}, {'a': [1, null, 3]}, {'a': null}, {}] AS arr

// result
    { "": { "a": [], "v": [] } }
    { "": { "a": [1, 2, 3], "v": [] } }
    { "": { "a": [1, null, 3], "v": [] } }
    { "": { "a": null, "v": null } }
    { "": { "v": null } }
```
)

mongosql/src/ast/rewrites/test.rs
```diff
@@ -1418,6 +1418,11 @@ mod scalar_functions {
 
 mod higher_order_functions {
     use super::*;
+    test_rewrite!(array_remove_with_field,
+        pass = HigherOrderFunctionsRewritePass,
+        expected = Ok("SELECT VALUE {'a': a, 'v': FILTER(a, NOT (this IS NULL AND x IS NULL) AND (this IS NULL OR (x IS NULL OR this <> x)))} FROM array_default_numeric_data AS array_default_numeric_data"),
+        input = "SELECT VALUE { 'a': a, 'v': ARRAY_REMOVE(a, x) } FROM `array_default_numeric_data` AS `array_default_numeric_data`",
+    );
```
mattChiaravalloti:
```suggestion
```
[nit] This test is not distinct from test `array_remove` below (line 1492), so we should remove it.

tests/spec_tests/query_tests/array_functions.yml
```diff
@@ -63,6 +61,17 @@ tests:
       - { '': { "a": null, "v": null } }
       - { '': { "v": null } }
 
+  - description: ARRAY_REMOVE remove field correctness test
+    current_db: spec_query_array_functions
+    query: "SELECT VALUE { 'a': a, 'v': ARRAY_REMOVE(a, x) } FROM `array_default_numeric_data` AS `array_default_numeric_data`"
```
mattChiaravalloti: [suggestion] 2 suggestions for this test:
1. Let's use a different data set that includes more rows. We should demonstrate what happens when `x` is null and there are no null values in `a`, and what happens when `x` is not null and there is at least one null value in `a`. We already nearly cover that first case via the row with `_id: 3` so maybe we could just add one more row to cover the second case. But that would result in updating every test in this file, which seems not worth it. An additional collection is super cheap to create and use so we should probably go down that route.
2. Can you include `x` in the result set? `SELECT VALUE { 'a': a, 'x': x, 'v': ...` That way it is easy to just read this test and know the behavior without having to correlate the test with the data in a separate file.

tests/spec_tests/rewrite_tests/array_functions.yml
```diff
@@ -17,9 +17,13 @@ tests:
...
 
-  - description: Rewrite ARRAY_REMOVE to FILTER
+  - description: Rewrite ARRAY_REMOVE against a field to FILTER
     query: "SELECT ARRAY_REMOVE(`array`, x) FROM foo"
-    result: "SELECT VALUE {'_1': FILTER(`array`, this <> x)} FROM foo AS foo"
+    result: "SELECT VALUE {'_1': FILTER(`array`, NOT (this IS NULL AND x IS NULL) AND (this IS NULL OR (x IS NULL OR this <> x)))} FROM foo AS foo"
+
+  - description: Rewrite ARRAY_REMOVE against null literal to FILTER
+    query: "SELECT ARRAY_REMOVE(`array`, null) FROM foo"
+    result: "SELECT VALUE {'_1': FILTER(`array`, this <> NULL)} FROM foo AS foo"
```
mattChiaravalloti: [chore] We'll either remove or update this when we make the change to the rewriter.

mattChiaravalloti (changes requested):
Thanks again for coming up with this idea of using a more complex rewrite as opposed to making thousands of changes to the entire translation pipeline and several IRs! This is a cleaner solution. I left one critical comment that addresses correctness along with a handful of very small changes to testing. But this is effectively ready to go after those changes! The good news is the required change is simply deleting a couple lines of code and comments that handles the special casing of the `NULL` literal.

tests/spec_tests/query_tests/array_functions.yml
jpowell-mongo: Will do 👍

mongosql/src/ast/rewrites/higher_order_functions.rs
jpowell-mongo: Yeah, this is where I was split on how we're meant to handle this. In SQL, your'e note meant to use <> for null values. Instead you're meant to use a "IS NULL" or "IS NOT NULL" for null checks.

I want to test this out locally, and see where this lands, but thanks for raising this! I was back and forth on <> vs. IS NOT NULL. 

Noted 👍

mattChiaravalloti: Hm, I think there's a misunderstanding here. The non-literal rewrite you have is correct in all cases. I'm simply suggesting we don't make a special case for the null literal.

mattChiaravalloti: There are no changes needed to the conditions in the rewrite, we just need to stop rewriting `ARRAY_REMOVE(a, NULL)` into `FILTER(a, this <> NULL)` and instead rewrite to the same `FILTER` expression as everything else.

mongosql/src/ast/rewrites/higher_order_functions.rs
```diff
@@ -192,17 +192,68 @@ impl HigherOrderFunctionsAliasVisitor {
...
```
mattChiaravalloti:
```suggestion
    /// Rewrite `ARRAY_REMOVE(a, x)` into
    /// `FILTER(a, NOT (this IS NULL AND x IS NULL) AND (this IS NULL OR x IS NULL OR this <> x))`.
```

mattChiaravalloti (approved):
lgtm 🔥

nicholascioli (approved):
LGTM! Well done

PR #198, opened 2026-08-31
[SQL-3427] $expr language for mismatched bound types

jpowell-mongo:
## Summary

This pull request updates the generate_match() function in schema-builder-library to use $expr language if it encounters partition bounds that are not comparable BSON types.

Previously, generate_match() used $match language for generating the aggregation pipeline for finding partition documents. However, $match language uses type-bracketing, which means if a collection has a polymorphic "_id" field, this function would fail to find those bounds.

## Changes
Partition::generate_match previously always used match-language ($gte/$lt/$lte) against the partition key. Match language is type-bracketed. We don't want this when the min/max bounds for the partition are different types.                              
                                                                                                                                                                                     
  This change adds bounds_are_comparable and branches on it:                                                                                                                         
                                                                                                                                                                                     
  - Match language (default): used when min and max are the same BSON type, or when both are numeric (Int32/Int64/Double/Decimal128), since numerics compare across their types in   
    match language. This is the common case and keeps the simpler, index-friendlier query shape.                                                                                     
  - $expr (fallback): used when the bounds are different BSON types. $expr comparisons use BSON ordering, so a partition key of any type can fall between the bounds.

2. Added a dataset mismatched_id_fields_small.yml with a small dataset to validate that we generate partitions with mismatched types. This was uploaded to S3, and has a corresponding change in the evergreen.yaml to pull down that file for integration testing.

## How I tested this

1. I built schema-build-library locally, generated a dataset with mismatched _id fields, and ran schema-manager-rs to validate that we can derive a schema.
2. Validated that I could then write a MongoSQL query against that collection.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -792,8 +792,7 @@ functions:
         working_dir: mongosql-rs/testdata
         script: |
           ${prepare_shell}
-          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library.tar.gz
-
+          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library-new.tar.gz
   "decompress schema-builder-library testdata":
     - command: shell.exec
       params:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -804,7 +804,7 @@ functions:
           rm -rf testdata/schema-builder-library/
           echo "Removal complete"
           echo "Decompressing integration test data..."
-          tar -xzvf testdata/schema-builder-library.tar.gz --directory testdata
+          tar -xzvf testdata/schema-builder-library-new.tar.gz --directory testdata
           echo "Decompression complete"
 
   "generate static code analysis":
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -792,7 +792,7 @@ functions:
         working_dir: mongosql-rs/testdata
         script: |
           ${prepare_shell}
-          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library-new.tar.gz
+          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library.tar.gz
   "decompress schema-builder-library testdata":
     - command: shell.exec
       params:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -804,7 +804,7 @@ functions:
           rm -rf testdata/schema-builder-library/
           echo "Removal complete"
           echo "Decompressing integration test data..."
-          tar -xzvf testdata/schema-builder-library-new.tar.gz --directory testdata
+          tar -xzvf testdata/schema-builder-library.tar.gz --directory testdata
           echo "Decompression complete"
 
   "generate static code analysis":
--- a/schema-builder-library/src/partitioning/partition.rs
+++ b/schema-builder-library/src/partitioning/partition.rs
@@ -22,6 +22,33 @@ fn bounds_are_comparable(min: &Bson, max: &Bson) -> bool {
         || min.element_type() == max.element_type()
 }
 
+/// Returns the string `$type` reports for `bound`, for use in `$expr` type predicates.
+fn type_name(bound: &Bson) -> &'static str {
+    match bound {
+        Bson::Double(_) => "double",
+        Bson::String(_) => "string",
+        Bson::Array(_) => "array",
+        Bson::Document(_) => "object",
+        Bson::Boolean(_) => "bool",
+        Bson::Null => "null",
+        Bson::RegularExpression(_) => "regex",
+        Bson::JavaScriptCode(_) => "javascript",
+        Bson::JavaScriptCodeWithScope(_) => "javascriptWithScope",
+        Bson::Int32(_) => "int",
+        Bson::Int64(_) => "long",
+        Bson::Timestamp(_) => "timestamp",
+        Bson::Binary(_) => "binData",
+        Bson::ObjectId(_) => "objectId",
+        Bson::DateTime(_) => "date",
+        Bson::Symbol(_) => "symbol",
+        Bson::Decimal128(_) => "decimal",
+        Bson::Undefined => "undefined",
+        Bson::MaxKey => "maxKey",
+        Bson::MinKey => "minKey",
+        Bson::DbPointer(_) => "dbPointer",
+    }
+}
+
 #[derive(Debug, PartialEq, Clone)]
 pub struct Partition {
     pub min: Bson,
@@ -34,6 +61,9 @@ impl Partition {
     // optional schema, and a list of _id values to ignore. If the Schema is None, the $match will
     // only be based on the Partition bounds and the ignored_ids list.
     // If the min and max bounds are not comparable in match language, the $expr language will be used.
+    //
+    // Both branches share the same logic: a document is in the partition if its partition key type
+    // matches the type of one of the bounds.
     #[instrument(level = "trace", skip_all)]
     pub fn generate_match(
         &self,
@@ -51,14 +81,27 @@ impl Partition {
         // $expr language
         if !bounds_are_comparable(&self.min, &self.max) {
             let key_path = format!("${partition_key}");
+            // Each bound constrains only documents of that bound's own BSON type. Note that partition key
+            // values whose type matches neither bound are excluded.
+            // So if the bound types are (ObjectId, String) but some ids are Int32, those ids will be excluded.
             let mut expr_body = doc! {
                 "$expr": {
                     "$and": [
-                        doc! {"$gte": [&key_path, self.min.clone()]},
-                        doc! {lt_op: [&key_path, self.max.clone()]},
                         // $literal keeps a $-prefixed ignored value from being read as a
                         // field path.
                         doc! {"$not": {"$in": [&key_path, {"$literal": ignored_ids.to_vec()}]}},
+                        doc! {
+                            "$or": [
+                                doc! {"$and": [
+                                    doc! {"$eq": [{"$type": &key_path}, type_name(&self.min)]},
+                                    doc! {"$gte": [&key_path, self.min.clone()]},
+                                ]},
+                                doc! {"$and": [
+                                    doc! {"$eq": [{"$type": &key_path}, type_name(&self.max)]},
+                                    doc! {lt_op: [&key_path, self.max.clone()]},
+                                ]},
+                            ]
+                        },
                     ]
                 }
             };
--- a/schema-builder-library/src/partitioning/partition.rs
+++ b/schema-builder-library/src/partitioning/partition.rs
@@ -3,6 +3,25 @@ use tracing::instrument;
 
 pub const PARTITION_SIZE_IN_BYTES: i64 = 100 * 1024 * 1024; // 100 MB
 
+/// Returns true when `min` and `max` can be compared using match-language range operators.
+fn bounds_are_comparable(min: &Bson, max: &Bson) -> bool {
+    fn is_wildcard(bound: &Bson) -> bool {
+        matches!(bound, Bson::MinKey | Bson::MaxKey)
+    }
+
+    fn is_numeric(bound: &Bson) -> bool {
+        matches!(
+            bound,
+            Bson::Int32(_) | Bson::Int64(_) | Bson::Double(_) | Bson::Decimal128(_)
+        )
+    }
+
+    is_wildcard(min)
+        || is_wildcard(max)
+        || (is_numeric(min) && is_numeric(max))
+        || min.element_type() == max.element_type()
+}
+
 #[derive(Debug, PartialEq, Clone)]
 pub struct Partition {
     pub min: Bson,
@@ -27,18 +46,45 @@ impl Partition {
             "$lt"
         };
 
-        let mut match_body = doc! {
-            partition_key: {
-                "$nin": ignored_ids,
-                "$gte": self.min.clone(),
-                lt_op: self.max.clone(),
+        // If the min and max bounds are not comparable in match language, fall back to the
+        // $expr language, whose comparison operators do not order values across BSON types.
+        if !bounds_are_comparable(&self.min, &self.max) {
+            let key_path = format!("${partition_key}");
+            let mut expr_body = doc! {
+                "$expr": {
+                    "$and": [
+                        doc! {"$gte": [&key_path, self.min.clone()]},
+                        doc! {lt_op: [&key_path, self.max.clone()]},
+                        // $literal keeps a $-prefixed ignored value from being read as a
+                        // field path.
+                        doc! {"$not": {"$in": [&key_path, {"$literal": ignored_ids.to_vec()}]}},
+                    ]
+                }
+            };
+
+            // $jsonSchema has no $expr equivalent, so the schema exclusion stays in match
+            // language as a sibling of $expr, which is valid within a single $match.
+            if let Some(schema) = doc {
+                expr_body.insert("$nor", vec![schema]);
             }
-        };
-        if let Some(schema) = doc {
-            match_body.insert("$nor", vec![schema]);
-        }
-        doc! {
-            "$match": match_body
+
+            doc! {
+                "$match": expr_body
+            }
+        } else {
+            let mut match_body = doc! {
+                partition_key: {
+                    "$nin": ignored_ids,
+                    "$gte": self.min.clone(),
+                    lt_op: self.max.clone(),
+                }
+            };
+            if let Some(schema) = doc {
+                match_body.insert("$nor", vec![schema]);
+            }
+            doc! {
+                "$match": match_body
+            }
         }
     }
 }
--- a/schema-builder-library/src/partitioning/partition.rs
+++ b/schema-builder-library/src/partitioning/partition.rs
@@ -33,6 +33,7 @@ impl Partition {
     // generate_partition_match generates the $match stage for sampling based on the partition, an
     // optional schema, and a list of _id values to ignore. If the Schema is None, the $match will
     // only be based on the Partition bounds and the ignored_ids list.
+    // If the min and max bounds are not comparable in match language, the $expr language will be used.
     #[instrument(level = "trace", skip_all)]
     pub fn generate_match(
         &self,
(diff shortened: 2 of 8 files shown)


schema-builder-library/src/partitioning/partition.rs
```diff
@@ -27,18 +77,58 @@ impl Partition {
...
+                                doc! {"$and": [
+                                    doc! {"$eq": [{"$type": &key_path}, type_name(&self.min)]},
+                                    doc! {"$gte": [&key_path, self.min.clone()]},
+                                ]},
+                                doc! {"$and": [
+                                    doc! {"$eq": [{"$type": &key_path}, type_name(&self.max)]},
+                                    doc! {lt_op: [&key_path, self.max.clone()]},
+                                ]},
+                            ]
+                        },
```
EthanHardyMongo: Since expression language is not type bracketed, you only need to add this to make different BSON types work: 
```
            partition_key: { "$nin": ignored_ids },
            "$expr": {
                "$and": [
                    doc! {"$gte": [partition_key, self.min.clone()]},
                    doc! {lt_op: [partition_key, self.max.clone()]},
                ]
             }
```
From the [docs:](https://www.mongodb.com/docs/manual/reference/operator/aggregation/gte/)
> The [$gte](https://www.mongodb.com/docs/manual/reference/operator/aggregation/gte/#mongodb-expression-exp.-gte) compares both value and type, using the [specified BSON comparison order](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#std-label-bson-types-comparison-order) for values of different types. 

Essentially, we just need the bounds to be in a `$expr` so that the expression language comparison operators are used instead of the match (query) language ones.

Also, this way fixes the case where there's more than two BSON types in the partition (i.e., the BSON types between the two bounds won't be excluded).

EthanHardyMongo (changes requested):
This PR can be heavily simplified because expression language comparison operators don't type bracket. I explain this more in a comment.

schema-builder-library/src/partitioning/partition.rs
```diff
@@ -3,6 +3,52 @@ use tracing::instrument;
 
 pub const PARTITION_SIZE_IN_BYTES: i64 = 100 * 1024 * 1024; // 100 MB
 
+/// Returns true when `min` and `max` can be compared using match-language range operators.
+fn bounds_are_comparable(min: &Bson, max: &Bson) -> bool {
+    fn is_wildcard(bound: &Bson) -> bool {
+        matches!(bound, Bson::MinKey | Bson::MaxKey)
+    }
```
EthanHardyMongo: We should remove this as well because there could be 3 or more BSON types in a partition, so even though the MinKey or MaxKey won't cause issues, the other bound will.

schema-builder-library/src/partitioning/partition.rs
jpowell-mongo: Okay, I'm open to switching this back. I originally had the logic to have those bounds to support the polymorphic behavior. I made the decision to keep it in the current form because I thought it was unlikely that we would see collections with multiple type for the _id field.

I'll update accordingly.

schema-builder-library/src/partitioning/partition.rs
jpowell-mongo: Are you saying that we should not use $expr language if the bound includes MinKey, MaxKey? 

I think we still need to know the bound is Min or Max Key so we can validate if this goes in expr lang or not?

EthanHardyMongo: We should use `$expr` language when a wildcard is used, but this part prevents that:
```
    is_wildcard(min)
        || is_wildcard(max)
```
Why do we need special handling for `MinKey` and `MaxKey`? Match language should only be used if both bounds are numeric or the same type.

schema-builder-library/src/partitioning/partition.rs
```diff
@@ -14,6 +26,10 @@ impl Partition {
     // generate_partition_match generates the $match stage for sampling based on the partition, an
     // optional schema, and a list of _id values to ignore. If the Schema is None, the $match will
     // only be based on the Partition bounds and the ignored_ids list.
+    // If the min and max bounds are not comparable in match language, the $expr language will be used.
+    //
+    //  $expr comparisons use BSON total ort order, so a partition key of any type can fall between the bounds, while match language
+    // is type-bracketed and only matches keys that have the exact type as bound's type.
```
EthanHardyMongo:
```suggestion
    //
    //  $expr comparisons use BSON total sort order, so a partition key of any type can fall between the bounds, while match language
    // is type-bracketed and only matches keys that have the same type as the bound.
```
typo nit

schema-builder-library/src/partitioning/partition.rs
```diff
@@ -3,6 +3,18 @@ use tracing::instrument;
...
+fn bounds_are_comparable(min: &Bson, max: &Bson) -> bool {
+    fn is_numeric(bound: &Bson) -> bool {
+        matches!(
+            bound,
+            Bson::Int32(_) | Bson::Int64(_) | Bson::Double(_) | Bson::Decimal128(_)
+        )
+    }
+
+    (is_numeric(min) && is_numeric(max)) || min.element_type() == max.element_type()
+}
```
EthanHardyMongo, line 19:
```suggestion

    (is_numeric(min) && is_numeric(max)) || min.element_type() == max.element_type() || matches!((min, max), (Bson::MinKey, Bson::MaxKey))
}
```
I just realized that when the bounds are MinKey _and_ MaxKey, we can use match language. However, when just one bound is MinKey/MaxKey, $expr language is required. Although, this is such a rare edge case, so up to you if you want to add it. If you add it though, make sure to revert the tests that use MinKey and MaxKey for the bounds.

evergreen.yml
```diff
@@ -792,8 +792,7 @@ functions:
         working_dir: mongosql-rs/testdata
         script: |
           ${prepare_shell}
-          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library.tar.gz
-
```
EthanHardyMongo, line 796: The wasm tests also use this testdata, and its "listing databases works" test asserts the output of listDatabases(), so add the new DB to this list, so that it passes. 

```
    test("listing databases works", async () => {
        const dbs = await SERVICE.listDatabases();

        expect(dbs.sort()).toMatchInlineSnapshot(`
          [
            "admin",
            "config",
            "local",
            "nonuniform",
            "uniform",
          ]
        `);
    });
```
Also, this part of the README, points to the old testdata. Update this link as well.

> ## Integration testing for schema-builder-library
>The `schema-builder-library` integration tests cover library method correctness against a live
database. These tests require a running enterprise mongod server (version at least 6.0). They also
require that data be loaded in before running the tests. The data is
stored [here](https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library.tar.gz).
After decompressing the data, to load it into the database,

evergreen.yml
```diff
@@ -805,7 +804,7 @@ functions:
           rm -rf testdata/schema-builder-library/
           echo "Removal complete"
           echo "Decompressing integration test data..."
-          tar -xzvf testdata/schema-builder-library.tar.gz --directory testdata
+          tar -xzvf testdata/schema-builder-library-new.tar.gz --directory testdata
           echo "Decompression complete"
```
EthanHardyMongo: In evergreen, the tests use `testdata/schema-builder-library/`,  but I think it will now be `testdata/schema-builder-library-new`, so you'll need to update those paths as well. 

Also, it doesn't look like the `schema-builder-library` buildvariant runs for PRs (which is why you don't see any evergreen failures), so run a patch for that to ensure it works.

EthanHardyMongo (approved):
LGTM! I just have a few minor comments.

evergreen.yml
EthanHardyMongo: I just added `schema-builder-library` to the PR testing variant regex, so it should work now if you push another commit.

evergreen.yml
EthanHardyMongo: Also, there's no evergreen failures because the wasm tests were never being run because the `schema-builder-library` buildvariant was not added to the pr checks. I just added it though, so they should run when you push again.

evergreen.yml
```diff
@@ -792,8 +792,7 @@ functions:
         working_dir: mongosql-rs/testdata
         script: |
           ${prepare_shell}
-          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library.tar.gz
-
+          curl -LO https://mongosql-noexpire.s3.us-east-2.amazonaws.com/schema_manager/schema-builder-library-new.tar.gz
```
mattChiaravalloti: Are we not able to overwrite existing data in that S3 bucket? Do we need the `-new` part of the name? Why not just replace the existing `schema-builder-library.tar.gz` file with the new file and leave this code unchanged?

schema-builder-library/src/partitioning/test.rs
mattChiaravalloti, line 3: [nit/suggestion] (non-blocking) I like to do these drive-by improvements whenever we find files like this: this file already represents a module called `test` so there is no need to have this nested `test` module within it. Would you consider removing this line and the corresponding closing `}` at the end of the file and then making sure everything is tabbed back one tab? The module path can be `schema_builder_library::partitioning::test` instead of `schema_builder_library::partitioning::test::test`. 

Completely optional.

schema-builder-library/src/partitioning/test.rs
```diff
@@ -142,6 +287,64 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_generate_partition_match_with_mismatched_id_fields_dataset_bounds() {
```
mattChiaravalloti: [question] How is this test distinct from the first test in the file? It seems to be nearly identical except for the ordering of min and max bounds, which matches real bson sort ordering ([link](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/)). If it is not distinct, we should remove it.

schema-builder-library/src/partitioning/test.rs
```diff
@@ -11,6 +11,151 @@ mod test {
     static DEFAULT_PARTITION_KEY: LazyLock<String> = LazyLock::new(|| "_id".to_string());
     use crate::partitioning::partition::PARTITION_SIZE_IN_BYTES;
 
+    #[test]
+    fn test_generate_partition_match_with_mismatched_bson_types() {
```
mattChiaravalloti:
```suggestion
    fn generate_partition_match_without_schema_with_mismatched_bson_types_with_max_bound_inclusive() {
```
[nit] (non-blocking) (completely optional)
1. Consider removing the `test_` prefix from this and all tests. It is not necessary for 2 reasons: first, `#[test]` already marks this as a test function; second, it is nested within the `test` module so its output in `cargo test` will read as `schema_builder_library::partitioning::test::generate_partition_match_without_schema...`
2. Consider adding the full description of the test, similar to how the existing ones do. This test covers not only mismatched bson types, but also no schema and an inclusive max bound. Having the names describe the exact set could help identify gaps in coverage and can help future maintainers understand coverage more quickly. For example, we seem to be missing at least one test from this matrix:

schema | bounds | inclusive | covered?
-- | -- | -- | --
w/o schema | matched types | exclusive | yes - `test_generate_partition_match_without_schema_doc_max_bound_exclusive`
w/o schema | matched types | inclusive | yes - `test_generate_partition_match_without_schema_doc_max_bound_inclusive`
w/o schema | mismatched types | exclusive | yes - `test_generate_partition_match_with_mismatched_bson_types_max_bound_exclusive`
w/o schema | mismatched types | inclusive | yes - this test
w/ schema | matched types | exclusive | yes - `test_generate_partition_match_with_schema_doc_max_bound_exclusive`
w/ schema | matched types | inclusive | yes - `test_generate_partition_match_with_schema_doc_max_bound_inclusive`
w/ schema | mismatched types | exclusive | no
w/ schema | mismatched types | inclusive | yes - `test_generate_partition_match_with_mismatched_bson_types_and_schema`

Can you add coverage for that missing test?

3. Consider grouping the tests in a consistent order. The existing 4 tests were ordered: (w/o schema, inclusive), (w/o schema, exclusive), (w/ schema, inclusive), (w/ schema, exclusive). They are all inherently with "matched types" for the bounds. You can update the names to indicate that for consistency if you want. When adding the new tests, we may want to group the corresponding "mismatched types" 4 together above or below those existing 4 and then group the remaining new coverage (e.g., `test_generate_partition_match_with_mixed_numeric_bounds_uses_match_language` and `test_generate_partition_match_with_wildcard_min_uses_expr`) together above or below those.

schema-builder-library/src/partitioning/test.rs
```diff
@@ -11,6 +11,151 @@ mod test {
...
+                        "$gte": Bson::Int64(1),
+                        "$lt": Bson::Double(5000.0)
+                    }
+                }
+            }
+        );
+    }
+
+    #[test]
+    fn test_generate_partition_match_with_wildcard_min_uses_expr() {
```
mattChiaravalloti: [question] Should we add a corresponding test for wildcard max?

schema-builder-library/src/partitioning/test.rs
```diff
@@ -142,6 +287,64 @@ mod test {
...
+                            {"$lte": ["$_id", Bson::String("S".to_string())]}
+                        ]
+                    }
+                }
+            }
+        );
+    }
+
+    #[test]
+    fn test_generate_partition_match_with_long_and_object_id_bounds() {
```
mattChiaravalloti: [nit] This test also does not appear to be truly distinct in any way. Sure, it uses a different type for the max bound, but is that really very different? The important aspect we're testing is whether min and max are the same. We don't need to cover all combinations of types, one example should be sufficient.

schema-builder-library/src/partitioning/partition.rs
```diff
@@ -27,18 +46,45 @@ impl Partition {
...
         };
 
-        let mut match_body = doc! {
-            partition_key: {
-                "$nin": ignored_ids,
-                "$gte": self.min.clone(),
-                lt_op: self.max.clone(),
+        // If the min and max bounds are not comparable in match language, fall back to the
+        // $expr language
+        if !bounds_are_comparable(&self.min, &self.max) {
```
mattChiaravalloti: [supernit] (super non-blocking) (super optional) Since the negative case does not benefit from an early return, would you consider flipping the branches of the conditional so it reads positively instead of negatively?

That improves readability a bit since the result of `bounds_are_comparable` directly indicates which branch we go into, instead of needs to negate first. Right now, it reads like "bound_are_comparable returned false, which means we _do_ enter this condition". Obviously, the two options are completely functionally isomorphic so this is truly not a necessary change. I just wanted to flag it as a Rust idiom and a personal preference.

mattChiaravalloti (changes requested):
The new match condition looks great 👍  I only have a handful of comments about organization and test coverage

evergreen.yml
jpowell-mongo: Yes, I'll rename it. I was initially using the new name in case the updated data was malformed. But given it works correctly, I'll just update the file in S3. Good callout 👍

schema-builder-library/src/partitioning/partition.rs
jpowell-mongo: This was updated. Marking this as resolved!

evergreen.yml
jpowell-mongo: I've updated the tests accordingly and the Evergreen tasks are not passing.

schema-builder-library/src/partitioning/test.rs
jpowell-mongo: I'm actually going to leave this in. I know it's small, but having the "mod test" allows my IDE to run all the tests together as a module, whereas removing the mod test means I don't have a way to easily run all the tests together : / . I'm going to leave this in for now.

schema-builder-library/src/partitioning/test.rs
jpowell-mongo: It's not distinct in this case. I originally had a plan to enumerate all the pairs of types as tests, but abandoned it. I'll remove this test.

schema-builder-library/src/partitioning/test.rs
jpowell-mongo: I'll remove this test. I was aiming to exercise the variation of different types.

schema-builder-library/src/partitioning/test.rs
mattChiaravalloti: You should still be able to run it as a module in your IDE by running the `mod test` here: https://github.com/mongodb/mongosql/blob/001c8e4512aa1cbd3d5ed6551eabb76fc38ba9b4/schema-builder-library/src/partitioning.rs#L11

If you use a JetBrains IDE, at least, the play button to run tests is available there.

schema-builder-library/src/partitioning/partition.rs
jpowell-mongo: Yep. Reversed the order here:
https://github.com/mongodb/mongosql/pull/198/changes/32365ec2328729566047b3b25f0f1a23e2ca9957

schema-builder-library/src/partitioning/test.rs
jpowell-mongo: I updated the test names, and added a test for the missing case you highlighted:
   fn generate_schema_with_mismatched_bson_types_with_schema_exclusive() {}

mattChiaravalloti (approved):
lgtm 👍

mattChiaravalloti:
> Also, there's no evergreen failures because the wasm tests were never being run because the `schema-builder-library` buildvariant was not added to the pr checks. I just added it though, so they should run when you push again.

^ from @EthanHardyMongo 

@jpowell-mongo We removed these tests from the evergreen PR patch definition yesterday since they were consistently failing and blocking merges, so these tests are _not_ part of your PR patch right now. I see they passed for you. Not sure if you want to add them to the PR patch or confirm them manually one more time.

jpowell-mongo:
> > Also, there's no evergreen failures because the wasm tests were never being run because the `schema-builder-library` buildvariant was not added to the pr checks. I just added it though, so they should run when you push again.
> 
> ^ from @EthanHardyMongo
> 
> @jpowell-mongo We removed these tests from the evergreen PR patch definition yesterday since they were consistently failing and blocking merges, so these tests are _not_ part of your PR patch right now. I see they passed for you. Not sure if you want to add them to the PR patch or confirm them manually one more time.

The Schema-Builder-Library Variant test pass:
https://spruce.corp.mongodb.com/version/6aa4193e90df4f0008515a51/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #199, opened 2026-08-31
SQL-3453: Fix constant folding and desugaring in the context of higher order functions

mattChiaravalloti:
This PR updates the `constant_folding` optimizer and `sql_null_semantic_operators` desugarer to properly handle cases that could arise in the context of higher order function expressions. See the ticket or the new e2e test introduced in this PR to see a correctness bug that exists without these fixes.
  
The issue in `constant_folding` was primarily that the `SchemaInferenceState` used by the visitor did not contain any variable information, which caused `has_null_arg` to return false since it would always encounter an error when requesting the schema of a `Variable` ([link](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/mongosql/src/mir/optimizer/constant_folding/lib.rs#L41)). That meant that order mattered when specifying arguments to the function argument of a higher order function -- if a `Variable` appeared before a definitely-nullish argument, then the result was unconditionally `false` even if a literal null was present. See the e2e test for an example of this. This PR addresses this in an imprecise way, indicating that `this` and `value` are available in the `SchemaInferenceState` with the `Schema::Any` schema. SQL-3454 tracks work to make this more precise. For now, though, this is acceptable behavior.
  
The issue in the `sql_null_semantic_operators` desugarer was really a consequence of constant folding failing in the context of higher order functions ([link](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/mongosql/src/air/desugarer/sql_null_semantics_operators.rs#L327-L328)). This PR updates the desugarer to handle null literals even though it anticipates never encountering them. It is best to have that implemented correctly and unused than incorrectly in any circumstance.

### DIFF
--- a/mongosql/src/air/desugarer/testdata/desugar_sql_null_semantics.yml
+++ b/mongosql/src/air/desugarer/testdata/desugar_sql_null_semantics.yml
@@ -1709,3 +1709,87 @@ tests:
               },
           },
       }
+
+  - name: "desugar $sqlEq literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlEq": ["$a", { "$literal": null }] } } }
+    expected:
+      - { "$project": { "_id": 0, "expr": { "$literal": null } } }
+
+  - name: "desugar $sqlIndexOfCP literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlIndexOfCP": ["$a", { "$literal": null }, "$b", "$c"] } } }
+    expected:
+      - { "$project": { "_id": 0, "expr": { "$literal": null } } }
+
+  - name: "desugar $sqlLt literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlLt": ["$a", { "$literal": null }] } } }
+    expected:
+      - { "$project": { "_id": 0, "expr": { "$literal": null } } }
+
+  - name: "desugar $sqlLte literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlLte": ["$a", { "$literal": null }] } } }
+    expected:
+      - { "$project": { "_id": 0, "expr": { "$literal": null } } }
+
+  - name: "desugar $sqlGt literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlGt": ["$a", { "$literal": null }] } } }
+    expected:
+      - { "$project": { "_id": 0, "expr": { "$literal": null } } }
+
+  - name: "desugar $sqlGte literal null"
+    input:
+      - { "$project": { "_id": 0, "expr": { "$sqlGte": ["$a", { "$literal": null }] } } }
+    expected:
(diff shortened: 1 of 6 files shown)


mongosql/src/air/desugarer/testdata/desugar_sql_null_semantics.yml
mattChiaravalloti, line 1: I added a literal null test for every SQL semantic operator besides `$sqlAnd`, `$sqlOr`, and `$sqlIn` since those are handled separately already by existing tests. The list of functions is included in this file, or here for convenience: [link](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/mongosql/src/air/desugarer/sql_null_semantics_operators.rs#L477-L478).

testdata/e2e_tests/queries.yml
```diff
@@ -333,3 +333,29 @@ dataset:
...
+  - db: "higher_order_function_nested_sql_semantics"
+    collection:
+      name: "filter_array"
+      docs:
+        - { "_id": 0, "a": [] }
+        - { "_id": 1, "a": [ 1 ] }
+        - { "_id": 2, "a": [ 1, 2, 3 ] }
+        - { "_id": 3, "a": [ 1, null, 3 ] }
+        - { "_id": 4, "a": null }
+        - { "_id": 5 }
```
mattChiaravalloti, line 346: Arguably these 2 rows are not necessary but I left them in to demonstrate these changes do not impact general nullish behavior for the higher order functions.

tests/e2e_tests/queries.yml
```diff
@@ -664,3 +664,19 @@ tests:
...
         - { "foo_flatten_group": { "x": 2 } }
+
+  # The initial implementation of higher order functions relied on several incomplete and incorrect implementations in
+  # the translation pipeline. First, the optimizer did not properly handle Variable expressions when constant folding
+  # expressions nested within higher order functions. Second, the desugarer assumed the optimizer always successfully
+  # folded away null literals in expressions that required special SQL null semantics. These two issues in concert led
+  # to this test failing since `v1` would fail to properly assert SQL semantics and `v2` would fold to `NULL`.
+  - description: Higher order functions must handle SQL semantics correctly and consistently
+    current_db: higher_order_function_nested_sql_semantics
+    query: "SELECT VALUE { 'a': a, 'v1': v1, 'v2': v2, 'v1_eq_v2': v1 = v2 } FROM (SELECT VALUE { 'a': a, 'v1': FILTER(a, this > NULL), 'v2': FILTER(a, NULL < this) } FROM filter_array) q"
```
mattChiaravalloti, line 675: I only test with `FILTER` which seems sufficient in this suite, given the existing unit testing. If anyone has a compelling reason to also add e2e tests for various combinations of nested SQL-semantic functions in other higher order functions, let me know and I'll add tests as appropriate 👍

jchemburkar (approved):

EthanHardyMongo (approved):

PR #201, opened 2026-09-02
SQL-3436: Ignore first document in partition if $jsonSchema can't match it

EthanHardyMongo:
This PR fixes the infinite looping issue that some costumers were having. I also discovered that the empty key issue is no longer an issue for mongodb 8.0. Also, dotted fields in the `jsonSchema` `required` field cause this issue because they're seen as field paths instead of literal fields. There may be more instances that can cause this, but this PR will fix those cases as well.


schema-builder-library/src/schema.rs
```diff
@@ -84,7 +84,9 @@ pub async fn derive_schema_for_partition<S: LocalDataService>(
     initial_schema_doc: Option<Arc<Schema>>,
     single_partition: SinglePartition,
 ) -> Result<Schema, Error<S::Error>> {
-    let mut ignored_ids = Vec::new();
+    // At most one blocker at a time: it sits at `partition.min`, which the inclusive
+    // `$gte` bound cannot exclude. See the `ignored_min_id` assignment below.
+    let mut ignored_min_id: Option<Bson> = None;
```
EthanHardyMongo, line 89: This whole issue is from using `$gte` for the lower bound along with being unable to add the first document to the `ignored_ids`. However, the `ignore_ids` vector was never necessary because the partition bounds change, so the only document that could ever be seen in successive batches was the last document in the batch. This is why I've changed it so that we only need to ignore at most one document. All the other unmatchable documents have already been excluded because they're out of bounds. The long comment below explains this as well.

schema-builder-library/src/schema.rs
```diff
@@ -131,28 +133,35 @@ pub async fn derive_schema_for_partition<S: LocalDataService>(
...
-                if old_schema == iter_schema {
-                    ignored_ids.push(id.clone());
-                }
                 no_result = false;
             } else {
-                warn!(db, collection, "document {partition_key} field");
+                warn!(
+                    db,
+                    collection, "document is missing the {partition_key} field"
+                );
```
EthanHardyMongo, line 142: Fixed typo.

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
```diff
@@ -0,0 +1,141 @@
...
+    };
+}
+
+// A single unmatchable document is the guaranteed instance of the bug: it is alone in
+// its batch and sits at the partition minimum, so neither the schema filter nor the
+// inclusive `$gte` bound can exclude it.
+test_derivation_terminates!(
+    single_dotted_document,
+    docs = vec![doc! {"_id": 0, "a.b": 1}]
+);
```
EthanHardyMongo, line 87: I have a few tests here showing the different ways the infinite looping issue can come about, but essentially what happens is you end up with a batch of one document that `$jsonSchema` can't match. I just show the different ways you can end up in this condition.

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
```diff
@@ -0,0 +1,141 @@
...
+// More generally the loop hangs once a batch holds a single unmatchable document: it is that
+// batch's first, which the pre-fix comparison against `iter_schema` (reset to `Unsat` every
+// batch) could never recognize, and it sits at the inclusive `$gte` bound.
+//
+// A document is unmatchable only while its dotted key survives in `required`, which is the
+// *intersection* of the unioned required sets -- so every document must carry the same dotted
+// key. Differing dotted keys all drop out and then match via `properties`, which compares
+// keys literally. The counts below are specific to these documents sharing one shape:
+// duplicates within a batch are recognized and skipped, so a batch of one arises exactly at a
+// count congruent to 1 modulo PARTITION_DOCS_PER_ITERATION.
```
EthanHardyMongo, line 108: This is in reference to how the old code would push the `_id`'s to a vector.

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
```diff
@@ -0,0 +1,141 @@
...
+
+test_derivation_terminates!(
+    dotted_documents_two_over_a_full_batch,
+    docs = (0..41).map(|i| doc! {"_id": i, "a.b": i}).collect()
+);
+
+// The count is not what matters. These two share `"a.b"`, so it survives the `required`
+// intersection and both stay unmatchable, but they differ elsewhere -- so neither is ever
+// recognized as a duplicate, `min` lands on the second, and the next batch holds it alone.
+// A count of two, nowhere near 1 modulo PARTITION_DOCS_PER_ITERATION.
```
EthanHardyMongo, line 122: This is also in reference to the old code. In this case the document count doesn't matter because the second document changed the schema, so it was never pushed to `ignored_ids`.

EthanHardyMongo:
evergreen retry

EthanHardyMongo:
evergreen retry

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
mattChiaravalloti, line 108: Discussed a bit offline: let's still add a test that covers the non-required case that crosses batches.

Consider this:
```
test_derivation_terminates!(
    nonrequired_dotted_documents_multiple_batches,
    docs = (0..10)
        .map(|i| doc! {"_id": i, "a.b": i})
        .chain((10..21).map(|i| doc! {"_id": i, "c.d": i}))
        .collect()
);
```

It is passing for me on my pull of this branch. It turns out this implicit traversal does not apply to `properties` _for now_, but it may be possible for future MongoDB versions to introduce that 💀 This test could be helpful to make sure we don't regress on that front either.

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
```diff
@@ -0,0 +1,141 @@
...
+        doc! {"_id": 1, "a.b": 1, "u1": 1},
+    ]
+);
+
+// Controls: no trailing batch of one, and no unmatchable document at all. A failure here
+// means the fix broke ordinary derivation rather than that the bug is unfixed.
+test_derivation_terminates!(
+    dotted_documents_filling_whole_batches,
+    docs = (0..40).map(|i| doc! {"_id": i, "a.b": i}).collect()
+);
```
mattChiaravalloti, line 136: How is this test different from `dotted_documents_two_over_a_full_batch`? I see it doesn't go to two "full" batches, but is that really so different?

schema-builder-library/src/schema.rs
```diff
@@ -131,28 +133,35 @@ pub async fn derive_schema_for_partition<S: LocalDataService>(
...
         if no_result {
             break;
         }
 
-        schema = schema.union(&iter_schema);
+        let new_schema = schema.union(&iter_schema);
+
+        // A batch that widened nothing means `schema` already covers what came back, yet
+        // `$nor: [$jsonSchema]` returned it anyway -- so the document at `partition.min`
+        // cannot match the `$jsonSchema` rendering of the very schema it contributed to.
```
mattChiaravalloti, line 154: I'm confused how we know the problem is `min`. How do we _know_ that? Why can't it be any other document in the batch? This might be a big ask but can you possibly share a full example demonstrating exactly how it is only ever `min` that is problematic and nothing else?

mattChiaravalloti (changes requested):
Great fix! I wish we could use `$literal` but alas that is not a match-language operator. I think we should add one more test, and possibly refine the code-comment explaining this situation, but beyond that this is a really good fix to a very subtle issue.

schema-builder-library/src/schema.rs
jpowell-mongo: Could you consider updating the comment of ignored_min_id with the explanation about the partition bounds, or about the logic for excluding the min_id when you go to a new partition? 

I think it's explained here in the Github conversation but I don't think I'd understand that by "just" reading the code.

schema-builder-library/src/schema.rs
jpowell-mongo: My understanding is that in the original logic, the old_schema and iter_schema would never be equal for the first document in a collection because you're comparing UNSAT to an actual schema of the first document.

I don't have an example, but my "hunch", is that because of how the cursor iterates, we'll at least always have a schema after the first document, and if we're union-ing schemas together, we should always see an "no-op" if we've already unioned a document.  I know this is hand-wavy but I think the cursor + the behavior of union should ensure that we don't see this in other cases 🤷

jpowell-mongo (approved):
I think the tests give made this change clearer for me, the logic makes sense and seems like it resolves the infinite loops. 

I think it would be worth attempting to respond about why we only care about min-id in the comment thread, or at least adding a bit more explanation to this pull request in case we revisit this in the future 👍 . Thanks for digging into this!

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
EthanHardyMongo: You're right that they're not different under the new code. The 40/41 split is about the old code: 41 is 1 mod 20, which is what produced a trailing batch of exactly one, and that document was always its batch's first, so the pre-fix `old_schema == iter_schema` check (with `iter_schema` reset to `Unsat` per batch) could never add it to `ignored_ids` which caused the hang. 40 terminated fine pre-fix, so it's there as a control case rather than as a repro. I can remove it if you want.

schema-builder-library/src/schema.rs
EthanHardyMongo: The min doc is the only one that needs to be excluded because any other bad doc is excluded by the bounds of the next iteration. We do `$gte: partition.min` in the aggregation, and inside the batch loop, `partition.min = id` runs for every document, so the last document of a batch becomes the min for the next batch. Therefore, the only document that can be be queried repeatedly and cause looping is the min. 

from the comment:
>Otherwise, `min` has advanced past any previous blocker and `$gte` excludes it

EthanHardyMongo: I added another comment explaining it:


>ignored_min_id is the id of a document that `schema`'s `$jsonSchema` rendering cannot
match. It is remembered so the next query can exclude it by id. At most one such document
can survive: each iteration queries `partition_key >= partition.min` sorted ascending, and
the batch loop advances `partition.min` to the last document it processes, so the inclusive
`>=` bound re-admits only the document sitting exactly at `min`.
Reset to `None` whenever a batch widens the schema, since `min` has then advanced past any
earlier blocker.

schema-builder-library/src/internal_integration_tests/derive_schema_unmatchable_documents.rs
mattChiaravalloti: Oh I see. No that's a good point. This is useful in case we change the implementation again in the future.

mattChiaravalloti (approved):
lgtm 🔥

PR #202, opened 2026-09-03
[DRAFT] [POC] [WIP] Window Functions One-Shot Investigation

jpowell-mongo:


mongosql/src/ast/rewrites/window_functions.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+struct WindowFunctionsRewriteVisitor {
+    /// The clause currently being walked, used both to decide whether a window function is
+    /// legal here and to name the offending clause in the error.
+    clause: &'static str,
+    /// Whether we are inside a `SELECT VALUE` body, where window functions are rejected for
+    /// the same reason aggregations are.
+    in_select_values: bool,
```
jpowell-mongo, line 31: Note: I made the decision to exclude window functions from being support in 
SELECT VALUE style queries such as:
"SELECT VALUE { 'a': a, 'v': FILTER(a, this > 0) } FROM `filter` AS `filter`"
because window functions are a rollup across a set of rows, and also because this is would be non-standard SQL. 

During investigation I wasn't convinced there was a clean way to handle this.

mongosql/src/mir/schema/mod.rs
```diff
@@ -357,17 +358,6 @@ impl CachedSchema for Stage {
                     source_result_set.max_size
                 };
 
-                // A helper to bind a field/alias to a schema.
```
jpowell-mongo, line 360: Note: investigate removing this? Not sure why this was removed for window functions.

jpowell-mongo: Update: looks like this helper was moved to a different location

mongosql/src/algebrizer/definitions.rs
```diff
@@ -355,6 +434,199 @@ impl<'a> Algebrizer<'a> {
...
+    }
+
+    /// Hoists every window function in the SELECT clause into one or more `Stage::Window`s,
+    /// returning the new plan and, for each window call, the expression that reads its
+    /// output.
+    ///
+    /// MQL allows a single partitionBy/sortBy per `$setWindowFields` but a per-output-field
+    /// window, so functions sharing a specification share a stage and differing
+    /// specifications produce a chain.
+    fn algebrize_window_functions(
```
jpowell-mongo, line 461: TODO: Give a closer review of this.

PR #203, opened 2026-09-08
SQL-3464: Fix And and Or schema checking

mattChiaravalloti:
This PR fixes the result of `And` and `Or` schema checking. Before this change, the schema checker returned `Null` for these operators if any operator MUST be null. However, that is incorrect: `false AND null` is `false`, a boolean; and `true OR null` is `true`, also a boolean! This is a long-standing bug, it seems.

The reason this caused a problem for `ARRAY_REMOVE` is that `this IS NULL OR NULL IS NULL OR NULL` would be considered equivalent to `NULL` in the constant-folding optimizer, which wreaked havoc on the rewritten `FILTER` condition. By fixing the schema for these operators, we fix the `ARRAY_REMOVE` behavior. (Note: we do not constant fold `... IS NULL` as of [SQL-1640](https://jira.mongodb.org/browse/SQL-1640). I will file a ticket to support it, which is definitely doable -- we just need to check if the `...` expr is null _or_ missing when the target is `NULL`. Anyway, that's completely orthogonal to this bug.)

### DIFF
--- a/mongosql/src/mir/schema/test/expressions/scalar_function.rs
+++ b/mongosql/src/mir/schema/test/expressions/scalar_function.rs
@@ -302,7 +302,7 @@ mod and {
     );
 
     test_schema!(
-        may_be_null,
+        may_be_null_because_arg_may_be_null,
         expected = Ok(Schema::AnyOf(set![
             Schema::Atomic(Atomic::Boolean),
             Schema::Atomic(Atomic::Null)
@@ -318,7 +318,7 @@ mod and {
     );
 
     test_schema!(
-        may_be_missing,
+        may_be_null_because_arg_may_be_missing,
         expected = Ok(Schema::AnyOf(set![
             Schema::Atomic(Atomic::Boolean),
             Schema::Atomic(Atomic::Null)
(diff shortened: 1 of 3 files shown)


tests/spec_tests/query_tests/array_functions.yml
```diff
@@ -52,17 +52,14 @@ tests:
 
   - description: ARRAY_REMOVE remove NULL correctness test
     current_db: spec_query_array_functions
-    query: "SELECT VALUE { 'a': a, 'x': x, 'v': ARRAY_REMOVE(a, NULL) } FROM `array_remove_correctness` AS `array_remove_correctness`"
```
mattChiaravalloti, line 55: Simplified to: (1) not return `x` since it is unused in target expression, and (2) use the smaller data set since the larger set does not cover any additional unique behavior when the element-to-remove is a literal `NULL`

mongosql/src/mir/schema/test/expressions/scalar_function.rs
```diff
@@ -333,14 +333,31 @@ mod and {
         schema_env = map! {("bar", 0u16).into() => Schema::AnyOf(set![Schema::Atomic(Atomic::Boolean), Schema::Missing])},
     );
 
+    test_schema!(
+        may_be_null_because_arg_must_be_null,
```
mattChiaravalloti, line 337: Note: I did not add `may_be_null_because_arg_must_be_missing` tests for `and` or `or` because it is invalid to have an expression that _must_ be `missing`!

jpowell-mongo (approved):

EthanHardyMongo (approved):

PR #204, opened 2026-09-09
SQL-3465: Fix determine_join_semantics optimizer to not convert LEFT joins into EquiJoins when both fields are nullable

mattChiaravalloti:
This PR addresses a correctness bug in the `determine_join_semantics` optimization. The `determine_join_semantics` optimizer does not consider `join_type` at all when it really should. The issue causing this bug is [here](https://github.com/mongodb/mongosql/blob/e7f5933c47ad6b5bcd3d479bdfeade1cc1800728/mongosql/src/mir/optimizer/determine_join_semantics/mod.rs#L177) where we semi-conditionally insert a filter before the `EquiJoin` to remove null values from the LHS. For a LEFT join, we do not want to remove those values.

The fix is not to simply amend that condition to also say `&& j.join_type == JoinType::Inner` since that would fall-through to the else branch which returns an `EquiJoin` that does not use SQL semantics (i.e., "null = null" will be true when we still want that to be false).

Instead, the fix in this PR is to _not_ convert a `Join` into an `EquiJoin` when both the local and foreign fields are nullable _and_ it is a `LEFT` join; instead we return the node unchanged. (It will be converted to a `LateralJoin` instead; there is no way to represent a SQL-semantic `LEFT` join using equi-`$lookup` syntax.)

New spec tests have been added to demonstrate this works for `LEFT` and `RIGHT` joins, and a new unit test was added to the optimization to demonstrate this case does not get rewritten.

### DIFF
--- a/mongosql/src/mir/optimizer/determine_join_semantics/mod.rs
+++ b/mongosql/src/mir/optimizer/determine_join_semantics/mod.rs
@@ -191,7 +191,7 @@ impl Visitor for JoinSemanticsOptimizerVisitor<'_> {
                             }
 
                             self.changed = true;
-                            if local_field.is_nullable && foreign_field.is_nullable {
+                            if both_nullable {
                                 Stage::MqlIntrinsic(MqlStage::EquiJoin(EquiJoin {
                                     join_type: j.join_type,
                                     source: Box::new(Stage::Filter(Filter {
--- a/mongosql/src/mir/optimizer/determine_join_semantics/mod.rs
+++ b/mongosql/src/mir/optimizer/determine_join_semantics/mod.rs
@@ -1,7 +1,7 @@
 /// Optimizes Join stages by converting them to EquiJoin stages when possible.
 /// If a Join stage's right source is a collection source and the condition is
 /// a simple equality comparison of fields (one from the left and one from the
-/// right), then the Join can be converted into an EquiJoin.
+/// right), then the Join can be converted into an EquiJoin in most cases.
 ///
 /// Note that the intent here is to enable the translator and codegen to produce
 /// an equijoin-style $lookup stage in the final pipeline. This style $lookup
@@ -16,9 +16,14 @@
 ///     replace the Join condition with a literal false value.
 ///   - If the schema indicates either the local or the foreign field must NOT
 ///     be null or missing, no filter is needed.
-///   - Otherwise, the schema indicates both MAY be null or missing. Therefore,
-///     a Filter stage that asserts the existence of the local field is placed
-///     before the EquiJoin stage.
+///   - Otherwise, the schema indicates both MAY be null or missing. For INNER
+///     joins, this case can still be handled by an EquiJoin: a Filter stage
+///     that asserts the existence of the local field is placed before the
+///     EquiJoin stage. For LEFT joins, this case cannot be lowered to an
+///     EquiJoin while still preserving SQL null semantics. This is because we'd
+///     either lose rows where the local field has nullish data by adding a
+///     filter, or we'd lose proper null semantics by omitting the filter and
+///     converting to an unfiltered equijoin $lookup.
 ///
 #[cfg(test)]
 mod test;
@@ -28,7 +33,7 @@ use crate::{
         optimizer::Optimizer,
         schema::{CachedSchema, SchemaCache, SchemaInferenceState},
         visitor::Visitor,
-        EquiJoin, Expression, FieldAccess, FieldPath, Filter, MqlStage, ScalarFunction,
+        EquiJoin, Expression, FieldAccess, FieldPath, Filter, JoinType, MqlStage, ScalarFunction,
         ScalarFunctionApplication, Stage,
     },
     schema::ResultSet,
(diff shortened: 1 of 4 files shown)


tests/spec_tests/query_tests/join.yml
```diff
@@ -41,6 +48,13 @@ tests:
         }
       - { "bar": { "_id": 0, "foo": 1 } }
 
+  - description: right join retains all rows from right hand side including those with with null data on the join key
```
jpowell-mongo: Great catch to also test the right join on this!

jpowell-mongo (approved):
I checked out the code, and did the repro from the issue with a LEFT JOIN and a RIGHT JOIN, and it looked good! 

Thanks for catching this and the additional details on the ticket itself.

LGTM pending fixing the Windows Evergreen task!

mongosql/src/mir/optimizer/determine_join_semantics/mod.rs
```diff
@@ -173,8 +178,20 @@ impl Visitor for JoinSemanticsOptimizerVisitor<'_> {
...
                         Some((local_field, foreign_field)) => {
+                            let both_nullable =
+                                local_field.is_nullable && foreign_field.is_nullable;
+
+                            // LEFT JOINs cannot be converted to EquiJoins if both fields are
+                            // nullable. This is because we need to retain all rows on the LHS but
+                            // do not want those rows to match rows from the RHS when the join keys
+                            // are both null. In this case, we skip this optimization in favor of
+                            // these being lowered to LateralJoins later in the optimizer.
+                            if both_nullable && j.join_type == JoinType::Left {
```
jpowell-mongo, line 195: Comment (non-blocking): I think it could be worth adding a short note saying that we don't need to check for JoinType::Right because we convert Right Joins into Left Joins in the algebrizer in algebrize_join_datasource() function.

Alternatively, we could also update the clause to be (j.join_type == JoinType::Left ||  j.join_type == JoinType::Right)

mattChiaravalloti: I'll add that note. I initially didn't since the `mir` only has 2 join types so there isn't such a thing as "right join" in the MongoSQL abstract model ([link](https://github.com/mongodb/mongosql/blob/742aaa1b8b9c71cb2adabbce35b7aaebe436fce9/mongosql/src/mir/definitions.rs#L240-L244)).

jpowell-mongo: Ah, apologies, sorry I should have referenced the mir definitions directly 😑 . Since it's functionally not possible, then I don't think we need to add a comment.

mattChiaravalloti: I already added it. No harm in having the info there as well.

EthanHardyMongo (approved):
LGTM!

tests/spec_tests/query_tests/join.yml
mattChiaravalloti, line 1: @nbagnard I updated the spec tests to cover the full matrix of null/non-null behavior for `INNER`, `LEFT`, and `RIGHT` joins. I worked out the expected results by hand, then confirmed them against MySQL and PostgreSQL ([db-fiddle](https://www.db-fiddle.com/f/ttHrGjJxozy9NEHqcQM7UQ/0)). These tests all pass, giving me full confidence in the bug fix: it retains correct behavior that already existed (nonnull-to-nonnull comparisons and nonnull-null / null-nonnull comparisons for all join types) and fixes the incorrect behavior (null-to-null for left/right joins).

testdata/spec_tests/query_tests/join.yml
nbagnard: Can we review datasets and tests for all types of join and make sure that we handle NULL values as expected? 
I think the gap comes for here. Because we have no spec tests with NULL data, we didn't catch the issue earlier. We want to make sure that we are covering all types of joins and the behavior with NULL join condition.

testdata/spec_tests/query_tests/join.yml
```diff
@@ -60,3 +60,43 @@ dataset:
       properties:
         x:
           bsonType: "int"
+
+  - db: "spec_query_join"
+    collection:
+      name: "parents"
+      docs:
+        - { "_id": 1, "nullable_field": 10, "non_nullable_field": 10 }
+        - { "_id": 2, "nullable_field": null, "non_nullable_field": 10 }
```
nbagnard, line 69: Let's add values that are not matching to covers all cases from the JOIN spec.
```suggestion
        - { "_id": 2, "nullable_field": null, "non_nullable_field": 10 }
        - { "_id": 3, "nullable_field": 4, "non_nullable_field": 5 }
```

testdata/spec_tests/query_tests/join.yml
```diff
@@ -60,3 +60,43 @@ dataset:
...
+            - bsonType: !!str "null"
+        non_nullable_field:
+          bsonType: "int"
+
+  - db: "spec_query_join"
+    collection:
+      name: "children"
+      docs:
+        - { "_id": 1, "nullable_field": 10, "non_nullable_field": 10 }
+        - { "_id": 2, "nullable_field": null, "non_nullable_field": 10 }
```
nbagnard, line 90: Let's add values that are not matching to covers all cases from the JOIN spec.
```suggestion
        - { "_id": 2, "nullable_field": null, "non_nullable_field": 10 }
        - { "_id": 3, "nullable_field": 6, "non_nullable_field": 7 }
```

nbagnard (changes requested):
I think adding rows that don't match will give us full coverage for all expected semantics.

testdata/spec_tests/query_tests/join.yml
mattChiaravalloti: Good call. I'll update with an additional row in `parents` and in `children` that do no match anything in the other and I'll update the result sets 👍

mattChiaravalloti: Done!

testdata/spec_tests/query_tests/join.yml
mattChiaravalloti: Done!

nbagnard (approved):
LGTM! :)

mattChiaravalloti:
The schema-builder-library-wasm variant is failing on the waterfall and on this patch. I'm removing it from the PR patch variant regex for now. We rarely need to run it, so it should be added manually if needed, maybe. Or we can add it back after this and #205 are merged in.

mattChiaravalloti:
evergreen refresh

mattChiaravalloti:
evergreen retry

mattChiaravalloti:
evergreen retry

mattChiaravalloti:
evergreen refresh

mattChiaravalloti:
evergreen refresh

PR #205, opened 2026-09-09
SQL-3051: Fix e2e and spec query test result set comparison algorithm

mattChiaravalloti:
This PR addressees a bug in the e2e / spec/query result set comparison algorithm; see the jira ticket for full details. The short of it is that the old implementation allowed for the same "expected" row to be reused multiple times by different "actual" rows. This is now fixed so each expected row can only match at most one actual row.

As a result, I had to modify one spec test, `MOD correctness test` since it was only passing because of this bug. The fix was simple: we accidentally had the `Decimal128` values use `e1` (* 10) instead of `e0` (* 1). This wasn't failing before since we do cross-numeric-type comparisons and the same expected rows could be reused by actual rows. That is now fixed.

I wanted to get this one done to ensure our confidence in new spec tests added recently: not only for array functions, but also for the changes in #204 .

### DIFF
--- a/evergreen/suite-tasks.yml
+++ b/evergreen/suite-tasks.yml
@@ -383,6 +383,25 @@ tasks:
           mongodb_version: rapid
     
 
+  - name: test-spec-query_latest-standalone
+    tags: [latest, standalone]
+    commands:
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          MONGODB_VERSION: latest
+          TOPOLOGY: server
+      - func: "install rust toolchain"
+      - func: "run data loader"
+        vars:
+          working_dir: ${common_test_infra_dir}
+          data_loader_args: "--mongod-uri mongodb://localhost:27017 -d ../testdata/spec_tests/query_tests"
+      - func: "run rust tests"
+        vars:
+          description: run spec query tests
+          cargo_test_flags: --features=query --package=e2e-tests
+          mongodb_version: latest
+
+
   - name: test-index-usage_latest-standalone
     tags: [latest, standalone]
     commands:
--- a/test-utils/src/query/mod.rs
+++ b/test-utils/src/query/mod.rs
@@ -10,7 +10,7 @@ use sql_engines_common_test_infra::{
     parse_yaml_test_file, sanitize_description, Error as cti_err, TestGenerator, YamlTestCase,
     YamlTestFile,
 };
-use std::{collections::HashSet, env, fs::File, io::Write, path::PathBuf};
+use std::{env, fs::File, io::Write, path::PathBuf};
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct QueryTestExpectations {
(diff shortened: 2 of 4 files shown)


tests/spec_tests/query_tests/scalar_functions.yml
```diff
@@ -325,42 +325,42 @@ tests:
...
-      - {'': { 'v': { "$numberInt": "80" }, 'divisor': { "$numberDouble": "NaN" }, 'n': { "$numberDouble": "NaN" }}}
-      - {'': { 'v': { "$numberDouble": "Infinity" }, 'divisor': { "$numberInt": "7" }, 'n': { "$numberDouble": "NaN" }}}
-      - {'': { 'v': { "$numberInt": "80" }, 'divisor': { "$numberDouble": "Infinity" }, 'n': { "$numberDouble": "80" }}}
-      - {'': { 'v': { "$numberDouble": "-Infinity" }, 'divisor': { "$numberInt": "7" }, 'n': { "$numberDouble": "NaN" }}}
-      - {'': { 'v': { "$numberInt": "80" }, 'divisor': { "$numberDouble": "-Infinity" }, 'n': { "$numberDouble": "80" }}}
-      - {'': { 'v': null, 'divisor': { "$numberInt": "7" }, 'n': null }}
-      - {'': { 'v': null, 'divisor': null, 'n': null }}
-      - {'': { 'v': null, 'n': null }}
-      - {'': { 'n': null }}
+    query: "SELECT VALUE {'_id': _id, 'v': v, 'divisor': divisor, 'n': MOD(v, divisor)} FROM mod AS c"
```
mattChiaravalloti, line 328: Adding the `_id` helped me debug and helps ensure that these rows are compared as we expect. Honestly, we may want to update all numeric-value function spec tests to include `_id` to help ensure we are comparing expected output types more accurately. That isn't super necessary and certainly isn't worth holding up this PR for. We can file a ticket for the backlog to consider some time in the future.

mattChiaravalloti (commented):
Most of the PR is unit test code for the new `assert_result_sets_equal` function!

evergreen/suite-tasks.yml
```diff
@@ -383,6 +383,25 @@ tasks:
           mongodb_version: rapid
     
 
+  - name: test-spec-query_latest-standalone
```
mattChiaravalloti, line 386: We were missing `test-spec-query` on `latest`. All other test suites run on `latest` except this one, so it seems to be an omission by accident.

I just filed SQL-3467 to track the work to migrate us off of `-latest` builds properly. For now, though, it seems best to ensure we have this variant existing so we include it in the migration as well.

nbagnard: Does mongo-orchestration support "stable"? I checked the BI Connector and curator uses "stable" which pull the latest released version. For example, right now, 8.3-stable gives 8.3.9,

mattChiaravalloti: I'll check. I can make that change here or we can defer it to one of SQL-3467 or SQL-3232.

mattChiaravalloti: It seems we cannot just use the label "stable": https://spruce.corp.mongodb.com/task/mongosql_ubuntu2204_test_spec_query_latest_standalone_patch_742aaa1b8b9c71cb2adabbce35b7aaebe436fce9_6aa2beee983e000007c76baf_26_09_10_14_30_22/logs?execution=0

We could use `latest-stable`, but that effectively still uses `latest`. We could possibly use `8.0-stable` but that is equivalent to `8.0` since it already fetches the latest stable version of `X.Y` when `X.Y` is the specified version.

I think we should defer figuring out how to avoid the "latest" tag to a different ticket since it is orthogonal to this PR. It's still worth having this new task in `suite-task.yml` since we are missing this coverage without it and it was likely omitted by accident.

tests/spec_tests/query_tests/scalar_functions.yml
nbagnard: Both options are fine by me.
Since fixing the comparison algorithm didn't uncover any issues, it's up to you if you want to file a follow up ticket or make the changes right now.

test-utils/src/templates/query_test_body_template
```diff
@@ -83,40 +83,12 @@ pub fn {name}() {{
     let type_compare = test.options.type_compare.unwrap_or(false);
```
nbagnard, line 83: question:
Did you introduce this with future test framework updates in mind?

test-utils/src/query/mod.rs
nbagnard, line 1: todo:
I love that you added unit tests to check our comparison logic and make it is correct, because an issue in this logic can hide bugs later on :)
Can we add tests with nested documents and arrays to make sure that the recurisve behavior works as expected too.
It seems that we have uncovered bugs a few time in our resultset validation logic, so it looks like it's worth adding.

nbagnard (changes requested):

test-utils/src/templates/query_test_body_template
mattChiaravalloti: Not sure what you're asking here. Are you asking about why `type_compare` was introduced? If so, that's not newly added as part of this PR. It seems to have been added as part of [SQL-2391](https://jira.mongodb.org/browse/SQL-2391) in this commit: [3f8cbcf](https://github.com/mongodb/mongosql/commit/3f8cbcf04573f8d66c2a45d88243a12c2983c1a7).

I'm actually not sure why SQL-2391 was filed and what it aimed to do. Honestly, I kind of fundamentally disagree with it since it says we should not be asserting specific value expectations from type conversions, but maybe I'm thinking of it wrong...

test-utils/src/query/mod.rs
mattChiaravalloti: Sure thing! I'll add those now

test-utils/src/templates/query_test_body_template
nbagnard: Yes, sorry, I wasn't too sure what this was for because all the unit tests use `false`, but you are right, I see it being used in `tests/schema_derivation_tests/query_spec_tests/type_conversions.yml` tests.

test-utils/src/query/mod.rs
mattChiaravalloti: Added! Turns out we were comparing arrays as unordered sets for some unknown reason. Updated to compare them correctly 👍

tests/spec_tests/query_tests/scalar_functions.yml
mattChiaravalloti: I'll spin it off into a follow-up ticket. A pretty significant number of tests would need to be updated and that would clutter this PR with tons of files and lines changed. A dedicated ticket/PR is probably for the best. The correctness will not change as a result of this. I'll file it now.

evergreen/suite-tasks.yml
mattChiaravalloti: Resolving as won't do for now.

nbagnard (approved):
LGTM! 🚀

test-utils/src/query/mod.rs
```diff
@@ -121,60 +121,36 @@ impl TestGenerator for QueryTestGenerator {
...
-            if e == a {
-                return seen_indices.insert(i);
-            }
-            false
+
+    expected
+        .iter()
+        .zip(actual.iter())
+        .all(|(expected_value, actual_value)| {
+            compare_bson_values(expected_value, actual_value, type_compare)
```
nicholascioli, line 136: Praise: Well done switching from procedural to functional. Absolute chef's kiss

test-utils/src/query/mod.rs
```diff
@@ -343,3 +304,379 @@ pub fn run_query(client: &Client, translation: Translation) -> Result<Vec<Docume
...
+        expected,
+        actual,
+    );
+
+    if ordered {
+        for (index, (e, a)) in expected.iter().zip(actual.iter()).enumerate() {
+            assert!(
+                // because NaN != NaN, we have to use custom comparison functions
+                compare_documents(e, a, type_compare),
+                "unexpected query result for {} at index {}, \nexpected: {:?}\nactual: {:?}",
```
nicholascioli: Nit: You can embed the values into this string directly:


```suggestion
                "unexpected query result for {desc} at index {index}, \nexpected: {e:?}\nactual: {a:?}",
```

test-utils/src/query/mod.rs
```diff
@@ -343,3 +304,379 @@ pub fn run_query(client: &Client, translation: Translation) -> Result<Vec<Docume
 
     Ok(result.into_iter().map(|d| d.unwrap()).collect())
 }
+
+/// assert_result_sets_equal compares two result sets for equality. It ensures
+/// that all expected results appear in the actual result set, and that no
+/// extra results appear in the actual result set.
+pub fn assert_result_sets_equal(
```
nicholascioli, line 311: Question: Is there a reason that this isn't defined in a test context? There isn't any non-test code that uses this, right?

nicholascioli (approved):
LGTM! Left a few nits, but nothing should be considered blocking.

test-utils/src/query/mod.rs
mattChiaravalloti: I think it's because this is an independent package that acts a utility library for a few of our test modules across multiple packages in this project. For example, this is used by the generated e2e tests in the `e2e-tests` package. I'm not sure you can publicly export function marked as `#[cfg(test)]` without also including additional features on them and then importing the crate with those features enabled. That might be something worth doing in the future, but for now this module is only used a dependency in the `e2e-test` package and the `schema-builder-library` package. The latter could benefit from marking it as a test-only dependency, I suppose

